PR #1, opened 2026-04-09
Test: validate worktree and PR workflow

Pumapumapumas:
Automated test. Safe to close.

PR #10, opened 2026-04-10
build-phase: GitHub Actions @claude PR comment integration

Pumapumapumas:
## Summary

- Adds `.github/workflows/claude-pr-comments.yml` — a GitHub Actions workflow that triggers when someone comments `@claude` on a PR
- Routes to revision (default) or revision-major workflows via `claude -p`
- Includes `@claude help` command that posts usage documentation
- Implements loop prevention, error handling, concurrency control, and JSONL log capture

## Deviation Summary: Planned vs Built

**Matched the plan:**
- `issue_comment` trigger with PR-only and bot-exclusion filters
- Three routes: default (revision), `revision-major:` prefix, and `help`
- Timeout handling (15 min revision, 30 min revision-major)
- Error comments on failure with link to run logs
- JSONL log capture (as GitHub Actions artifacts, 30-day retention)
- Acknowledgment thumbs-up reaction
- Concurrency control (one run per PR)

**Diverged from plan:**
- Invokes `claude -p` directly instead of calling `revision.sh`/`revision-major.sh` — the workflow scripts assume a local dev environment with worktrees, formatter, and jq. In GitHub Actions, we're already on the PR branch in a disposable runner, so direct invocation is simpler and more appropriate.

**Deferred:**
- Claude GitHub App installation (out-of-band setup task, documented in prerequisites)
- Fork PR access restriction (repo-level security setting, documented in security notes)

## Review Findings

**Addressed:**
- Fixed script injection vulnerability (moved step outputs to env vars instead of `${{ }}` in shell)
- Increased job timeout to 35 min (exceeds max step timeout of 30 min)
- Used `actions/checkout@v4` `ref` parameter for branch safety
- Added `GITHUB_RUN_ID` to log filenames for collision avoidance
- Added separate 9-stage prompt for revision-major route
- Added security documentation in file header
- Added git config step for Actions runner

**Deferred:**
- Fork contributor access control (repo-level concern)
- Bot check hardening beyond `user.type != 'Bot'` (sufficient for documented use case)
- npm version pinning (intentionally unpinned for latest Claude Code)

## Refactoring Suggestions

**Implemented:**
- Added `fromJSON` explanation comment
- Documented prompt divergence from canonical workflow scripts

**Deferred:**
- Split into two jobs (help vs execute) — adds complexity for minimal gain
- Shared prompt templates between scripts and Actions — future improvement
- Log filename / artifact name alignment — partially addressed with `GITHUB_RUN_ID`

## Test Results

[...]

PR #23, opened 2026-08-03
research: docs/standards/architecture/research

Pumapumapumas:
Re-assessment of the product-level research pool. **This is a re-run against an existing 8-paper pool, not a new one.**

## Sizing

**Large / architecture-layer — 11 topics, one above the rubric's 8–10 band, deliberately.** Three architectural layers landed since the 2026-07-27 assessment (parent/child decomposition, the extracted activities layer, the decide-only disposition engine) plus a language decision and a restructured roadmap, and none of the eight existing papers spoke to any of them. §2 requires the list to grow when the component grows. **3 of the 11 covered this cycle**, per §2's ~5-topic cap, sequenced most-decision-blocking first.

`system-overview.md` was **excluded from sizing as substantively stale** — it still describes "bash-over-Python" monoliths and knows nothing of parents/children, activities, or the disposition engine. That exclusion is written up as a finding in the synthesis.

**No papers retired.** Every existing subject still has a live roadmap destination; reasoning is in `topics.md`.

## Papers this cycle

| Topic | Confidence summary | Critic verdict |
|---|---|---|
| `convergence_stopping.md` | Definitive on what each study measured and on the negative finding that "3–5 passes" has no source describing our topology; directional on the 2026 preprints closest to our shape | **PASS** — 2 rounds. R1: 3 non-verbatim quotes (one phrase, "context bloat", was invented inside quote marks), 4 confidence downgrades, plus 2 more the analyst self-audited. R2 caught 1 residual non-verbatim span, now fixed |
| `workflow_reuse_boundary.md` | Definitive on what each system ships and deprecated and on the peer-reviewed empirics; derived on the central co-evolution claim; negative finding that no numeric threshold is documented | **PASS-WITH-FIXES** — re-sourced the ClusterTask→cluster-resolver claim to a source that actually states it ([S10] alone did not); 2 quote-fidelity fixes |
| `python_sdk_long_activities.md` | Definitive on SDK/server mechanics, version-anchored to `temporalio` 1.31.0 / server `main` @ 2026-08-03; derived on the recommended shape; unverified on orphaning and real 60-min heartbeat behaviour | **PASS-WITH-FIXES** — un-quoted an inference-from-absence dressed as an [S18] quote; 1 confidence re-mark; retagged an out-of-contract fifth confidence class |

Every correction was applied by the analyst and re-verified by the critic. Total: 3 correction rounds across 3 papers, none reaching the 3-round non-convergence limit.

## Synthesis action candidates (verbatim)

1. **Remove the "plateaus at 3–5 passes" justification from `revision.sh`.** The claim has no source describing our topology. This is a *provenance* correction, not a request to change the loop bound — the bound may still be right for cost reasons, but it must be justified by budget, not by a citation that does not exist — *change direction*
2. **Do not adopt a bare "stop when a pass finds nothing" rule.** Three independent findings kill it: non-monotonic yield, Böhme's adaptive-bias result, and the derived point that a hallucinating reviewer never emits an empty pass. Any convergence rule needs a hard count or budget backstop — *no change*
3. **Treat the 82%/9% overlap figures as insufficient input to the fork-vs-parameterize ruling.** The field's test is expected future co-evolution — "when one changed, did the other need the same change?" — answerable from commit history, not a diff. Also puts a third option on the table: tracked lineage with mechanical propagation — *change direction*
4. **Any parameter-count threshold we adopt must be labelled a local invention.** No numeric threshold is documented anywhere in the surveyed corpus — *adopt*
5. **Design the `claude_cli` activity as async-over-`asyncio.create_subprocess_exec`, not as a sync `def`.** The sync shape cannot heartbeat or be cancelled while blocked, and `graceful_shutdown_timeout` defaults to zero — *adopt*
6. **Carry transcripts as pointers into GitHub, not as payloads.** Do not reach for External Storage — *adopt*
7. **Open the question of whether a `claude -p` run decomposes into resumable per-turn legs.** The fork between single-activity and child-workflow shapes; not a Temporal question; no paper covers it — *new concept*
8. **Rewrite `temporal.md` at its next refresh rather than diffing it.** Pre-standard prose carrying a `Critic: PASS` its content does not support, past window, and its one stated gap is now closed — *change direction*
9. **Refresh the four past-window papers before the next planning run consumes them** — §5's planning fail-fast will otherwise stop that run — *no change*
10. **Revisit `system-overview.md`, or mark it stale in place** — *change direction*

Candidate #1 carries a full dependent trace (§4) naming five sites, one of which I explicitly did **not** verify and flag for the reviewer.

## Homeless findings (named in the synthesis, not parked)

- **A defined shape for production feedback.** Carried from the prior cycle, still homeless. The dated burn-test intake is a one-off, not a channel — nothing says where the next one goes.
- **Research Standard §3 has no confidence class for an authoritative speaker in an informal artifact.** A named Temporal maintainer on Temporal's own forum gets the same tag as an anonymous blog comment. Homeless *for us specifically*: the standard is vendored MIRROR, so the amendment goes upstream, and this repo has no surface holding "an upstream standards amendment we owe."

## Gaps / what research could not settle

- **Inter-process handoff contracts — redirected, not deferred.** The prior cycle named this a product-level gap; that was an altitude error. This folder's own `README.md` uses this exact question as its worked example of a *phase-level* one. It belongs in `phases/memory-management-framework/research/`, which does not exist because the phase doc is unwritten. Still the highest-value open research on the queue, and `convergence_stopping.md` deepens the dependency: without typed comparable outputs, convergence detection collapses to semantic distance over prose, which measures drift rather than discovery.
- **No first-party sample of a Temporal Python activity wrapping a subprocess exists** — the recommended `claude_cli` shape is a design hypothesis until tested. 17-item test plan in the paper.
- **Child-process orphaning on worker death is undocumented across every Temporal SDK.**
- **No study measures clone-divergence faults for CI/workflow definitions**, and **nothing studies duplicated prompt prose** — which is much of what our near-copies actually are.
- **The crux convergence experiment does not exist**: nothing measures marginal yield per *sequential* pass where each pass is a separate process with fresh context reading the *revised* artifact.

Four topics deferred with reasons in `topics.md`; two of them (per-turn decomposition, prompt-prose duplication) were surfaced by this cycle.


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Redirected **inter-process handoff contracts** from product-level to phase-level, overturning the previous cycle's placement of it as this pool's top gap. Alternatives: cover it here (the burn-test intake calls it the item that most needs real research); or leave it named at this altitude and defer again. Why: this folder's own `README.md` uses that exact question — *"should a parent hand off through a typed file or a parsed log"* — as its worked example of a **phase-level** question, and the standard's altitude test (does the finding *invalidate* a phase or *inform* one?) agrees. **The cost is real and I want it visible:** the highest-value open research on the queue is now uncovered *and* has no folder to live in, because `phases/memory-management-framework/` does not exist yet. A reviewer who thinks the README's example is wrong should overturn this rather than the placement.
- **[Low]** Set the topic count at **11, one above the Large tier's 8–10 band.** Alternatives: retire a paper to land on 10; or add only two topics. Why: three architectural layers landed since the last assessment and no existing paper speaks to any of them, and §2 both requires the list to grow with the component and calls its thresholds "a starting calibration." Retiring a paper to hit a number would have meant declaring a live subject dead. Stating the overage explicitly seemed better than hiding it by trimming.
- **[Low]** Ran **convergence stopping** this cycle even though the prior cycle deferred it as *"depends on the handoff contract landing first."* Alternatives: honour the stated dependency and defer again. Why: the dependency is on the *implementation* (typed payloads make convergence mechanizable), not on the *research* — the literature question is separable. And `revision.sh`'s loop bound is live in production today on a justification the operator had already flagged as a bad extrapolation. It turned out to be the cycle's most consequential finding, so I think this was right, but it did override an explicit prior-cycle judgement.
- **[Medium]** **Did not retire `temporal.md`**, against the Python-SDK analyst's explicit recommendation to "retire or rewrite" it. Alternatives: retire it. Why: §3 reserves retirement for a subject that *died*, and Temporal's direction was reaffirmed since the last assessment, not abandoned. The paper's problem is quality — pre-standard prose carrying a `Critic: PASS` its content does not support — which is a different defect with a different remedy. Surfaced as action candidate #8 ("rewrite at next refresh rather than diffing it") instead of laundering a quality complaint through the retirement mechanism.
- **[Medium]** Scoped the fork-vs-parameterize topic to the **premise** (does a parameterized shared library hold up?) rather than the concrete `build-phase` ruling. Alternatives: let the analyst rule on our actual scripts, which is what the burn-test intake most wants answered. Why: the concrete ruling is phase-level and the write boundary forbids a research run from performing planning. This looks like under-delivery and partly is — but the paper then established that the 82%/9% overlap figures are *not sufficient input* to that ruling in either direction, which is more useful to whoever makes it than a ruling from me would have been.
- **[Medium]** Spent a **third correction round** on the convergence paper for a defect the critic explicitly called non-blocking (a quote differing from source only in tense and one connective, no number affected). Alternatives: ship it; the substance was unaffected. Why: the paper's own `Critic:` line claimed body quotes were "re-fetched verbatim," which that span made untrue. A header that overclaims its own verification is a worse defect than the drift it conceals, because every downstream consumer reads the header and not the span.
- **[Medium]** Escalated the **staleness of `system-overview.md`** from "don't size against it" (what the dispatch instructed) into a written finding plus action candidate #10. Alternatives: silently not use it, as instructed. Why: the next cycle would walk into the same trap, and a stale architecture overview reads as authority. The dispatch invited the judgement; I took it as far as the write boundary allowed.

## Deferred Work

- **Four uncovered topics** (decide-only disposition; per-turn decomposition of a `claude -p` run; duplicated prompt prose; reflection-channel mining / bash→Python, both phase-level) — Why deferred: §2's binding ~5-topic-per-cycle cap; each carries its reason in-file. **Tracked at:** `docs/standards/architecture/research/topics.md` § *Gaps named, not covered this cycle*. **Verified by:** `git show origin/main:docs/standards/architecture/research/topics.md` → the file and its `## Gaps named, not covered this cycle` table exist on the default branch today and already carry the *previous* cycle's two deferrals ("Inter-process handoff contracts", "Convergence-based stopping conditions"), so the mechanism is proven rather than aspirational — this run read those rows and acted on both. **Honest caveat:** the four *new* rows are in this PR and only land on merge. If this PR is closed unmerged, these deferrals do not exist anywhere.
- **Refresh of four past-window papers** (`temporal.md`, `claude_code_integration_surface.md`, `anthropic_tos_and_enterprise.md`, `hook_sourcing_supplement.md`) — Why deferred: §5 assigns currency revalidation to `research-refresh.sh`, and the dispatch explicitly excluded it from this workflow. **Tracked at:** each paper's own `Revalidate:` header, which the refresh gate parses mechanically (`today − Last validated > interval` → due). **Verified by:** `git show origin/main:.../raw/<paper>.md | grep -E 'Last validated|Revalidate'` on all four → `temporal.md` 2026-07-04 / high — 4 weeks; `claude_code_integration_surface.md` 2026-07-25 / high — 4 weeks; `anthropic_tos_and_enterprise.md` 2026-07-24 / high — 4 weeks; `hook_sourcing_supplement.md` 2026-07-25 / high — 4 weeks. All four are on the default branch with parseable intervals and all four are past window as of 2026-08-03, so the gate fires independently of this PR. Also surfaced as action candidate #9 because §5's planning-stage fail-fast will otherwise stop the next planning run.

## Post-Run Reflection

**Friction**

- **`SendMessage` resumes an agent in the BACKGROUND, and nothing warns you.** The workflow prompt correctly mandates foreground `Agent` dispatch so a headless turn can never end with work outstanding — but Stage 4 says to "re-dispatch the research-analyst," and the natural way to do that (message the analyst that already holds the paper's context) has the opposite concurrency semantics. Two of three resumes went background and I had to invent foreground work to keep the run alive. This is a live run-killing hazard in the exact stage the prompt spends the most words protecting.
- **`SendMessage` failed for one of three analysts** with `No transcript found for agent ID`, while the other two resumed fine — same message, same turn. I reconstructed the correction as a fresh foreground analyst dispatch with the critic's findings inlined, which worked and was arguably better (it blocked the turn). But the failure mode is silent and asymmetric: you can't tell in advance which analysts are resumable.
- **Two artifacts in this folder disagreed on altitude for the same question**, with nothing indicating precedence: `topics.md` (prior run) listed inter-process handoff contracts as a product-level gap, while the folder's `README.md` uses that identical question as its example of a phase-level one. I ruled for the README, but a rule about which wins would have saved the judgement call.
- The dispatch's framing that the prior cycle's gaps were **"candidates, not instructions — assess them like anything else"** was the single most useful line in the task file. It is what made re-litigating one of them feel in-scope rather than insubordinate. Worth keeping in re-run dispatches verbatim.

- `docs/standards/architecture/system-overview.md` was moved under `standards/architecture/` on 2026-08-03 without its content being revisited, and now describes a system that no longer exists. Raised as action candidate #10; flagging separately because a *move* commit is exactly the kind that makes a doc look freshly touched.
- **No surface holds "an upstream vendored-standard amendment we owe."** `python_sdk_long_activities.md` found a genuine gap in the vendored Research Standard's confidence vocabulary (no class for an authoritative speaker in an informal artifact — a named Temporal maintainer on Temporal's own forum tags identically to an anonymous blog comment). Because the standard is vendored MIRROR, it can't be amended here, and there is no local place to record the debt. Named as a homeless finding in the synthesis.
- `phases/memory-management-framework/` does not exist, so the queue's highest-value research has no correct location to be written into. The phase doc being unwritten is the blocker, not the research.

- Stage 4 should state the `SendMessage`-resumes-in-background semantics explicitly, or simply mandate a **fresh foreground analyst dispatch** for corrections. It costs the analyst's original context, but the critic's findings are self-contained by design (the prompt already requires side-by-side paper/source quoting), and the fresh-dispatch path was both more reliable and headless-safe in this run.
- The correction-round budget ("MAX 3 rounds per paper") is currently silent on what a **non-blocking** finding costs. I spent a round on one and would do it again, but a run under budget pressure could reasonably read "non-blocking" as "ship it" — including when the non-blocking defect makes the paper's own `Critic:` header inaccurate, which is the case worth carving out.
- Two of three papers had their analyst self-audit find *additional* instances of the defect class the critic reported, unprompted. Stage 4 could ask for that explicitly ("audit every instance of the reported defect class, not only the cited ones") — it caught real issues here at near-zero cost.

PR #29, opened 2026-08-05
research-refresh: docs/standards/architecture/research — 1 paper

Pumapumapumas:
Revalidation cycle against `docs/standards/architecture/research`. One paper was due; the result is larger than a re-dating.

## Refreshed paper

**`raw/temporal.md`** — *Does Temporal supply what a durable workflow layer needs?* → **re-scoped** to *what committing to Temporal as a vendor costs us*

| | |
|---|---|
| **Four-category diff** | **CHANGED:** two server releases since 2026-07-04 (both 2026-07-08; v1.31.2 patches CVE-2026-5724 MEDIUM with a `system.disableStreamingAuthorizer` opt-out), CLI v1.8.2, Serverless Workers docs now exist at two paths. **NOW WRONG:** the paper's closing thesis — Temporal ships `contrib/langgraph`, so LangGraph is composition, not a rejected alternative; plus two unsourced claims removed (a named-companies adoption list, a "60%" figure). **MISSING:** the licence (MIT), all operational cost, all Cloud pricing, workflow-execution limits, Updates, Workflow Pause/Streams, the unlimited-retry default, and any case against its own thesis. **TOPIC:** the old question died — see below |
| **Revalidate** | `high — 4 weeks` → **`high — 6 weeks`** (top of §5's high band: the subject barely moved — two releases in 32 days, none in the last 28, licence unchanged, limits identical to a sibling's reading) plus an on-announcement override trigger list |
| **Critic** | **PASS-WITH-FIXES** — two independent passes, three repair rounds. Details below |
| **Size** | 39 lines → 799; 0 citations → 38 |

### Why the paper was 39 lines with no citations

It was a **pre-standard artifact** and had been consumed as the substrate paper for a live phase since 2026-07-04 while violating three separate §3 *"breaking it looks like"* clauses — no citations, no honest-boundary section, no content arc. **It was caught because it aged out, not because anything checked it.** Raised as a homeless finding.

## No retirement — one RE-SCOPE (human review requested)

**No paper is recommended for retirement.** But `raw/temporal.md`'s **topic** was re-scoped, which is a judgement call worth a human eye.

The old question failed on both halves: the **supply** half is decided (Temporal is chosen, the port is underway — continuing to ask it is decide-then-justify), and the **evidence** half was taken by siblings between 2026-07-27 and 2026-08-04. §6 retires a topic whose *subject* died; Temporal is the substrate of a live phase, so this is not that. The sweep found a real unowned question underneath — **nobody in the pool owned the cost and commitment of the vendor** (no paper stated the licence, priced Cloud, or carried the upgrade obligation), and retiring would have deleted the only home for it.

Note the paper's `Topic:`/`Feeds:` headers **were edited**, unlike `combination_prior_art.md` and `case_against.md`, whose headers were deliberately left alone when their destinations moved. Those are opposite cases — destination moved vs. question replaced — and the reasoning is recorded in `topics.md` so the two don't read as inconsistent.

## Critic record — the interesting part

**Round 1 (independent critic):** 4 blocking, all **MISCITED**, **zero fabricated** across 33 citations. A release count that was a truncated-fetch artifact *asserted under an "enumerated, not counted by a retrieval layer" heading*; two false negatives on Cloud pricing and the billable-Action definition that were **repo-path absences written up as documentation absences**; and a restructure claim built on a path the paper itself had constructed, which also carried a false flag against `dedicated_edge_routing.md`.

**Round 2 (second, independent critic — deliberately not the one that proposed the fixes):** verified all four repairs at source and re-derived all 11 counts by enumeration. It caught the pricing repair **converting one defect into a new one** — it correctly added a fourth Cloud plan and *invented a marketing-vs-documentation discrepancy to justify itself*, then used that invention to tell readers to distrust a source. It also caught a **stale dependent**: §7 still called the self-host-vs-Cloud decision research-blocked after §3.2 had reversed exactly that.

**Round 3:** both fixed, plus marker, citation and quotation cleanups. A full dependent sweep found no others.

**This is the second measured instance of a repair converting one defect into a worse one while appearing to resolve it, and a single critic round did not catch it.** That is a finding about the process, not this paper.

## Synthesis Diff

*What changed in `synthesis.md` relative to its 2026-08-04 version. This is the standup consumable.*

**Pool health.** The pool has **no past-window paper for the first time since 2026-07-04**. ⚠️ Two papers come due around **2026-08-17** (`subscription_economics.md`, `bernstein_capability_mining.md` — both `high — 2 weeks`); three more around 2026-08-22.

**Shifted conclusions.**

- **Temporal's LangGraph relationship inverted.** The pool previously carried LangGraph as prototype-grade and implicitly rejected. Temporal ships `contrib/langgraph` (plus `openai_agents`, `google_adk_agents`, `strands`): **composition, not substitution.**
- **Self-host-vs-Cloud went from "cannot be modelled" to "sizing question."** The refresh's own first pass claimed no first-party definition of a billable Action existed and called it the sole blocker; the critic closed it from `glossary.md` and `/cloud/actions` (**11 enumerated billable categories**). The residual unknown is a measurement, not a gap.
- **The Essentials plan is a floor, not an entry price** — "greater of $100/mo or 5% of Usage Spend." For a small single-operator fleet the floor binds, so $100 is the real number — but the plan table hides why.
- **No first-party Claude ↔ Temporal *runtime* integration exists**, established by enumerating `temporalio/contrib` (11 entries, no `anthropic/`, no `claude/`) while OpenAI, Google ADK, LangGraph and Strands all have one. **The hand-rolled integration is not a temporary state to wait out.**

**NEW action candidates (1–6).**

1. **Add a clause to `python_sdk_long_activities.md`'s heartbeat recommendation** — heartbeats are free at the SDK/transport layer and **billable on Temporal Cloud**. The recommendation survives; its stated justification ("the throttle makes per-line heartbeats free") does not transfer to billing. Ship the ceiling with it — `min(0.8 × heartbeat_timeout, 60 s)` bounds a 60-minute activity at **~60 heartbeat Actions** — or a planner will over-price Cloud. *(§4's trace-to-all-dependents rule: this is a correction in one paper landing on a sibling validated two days earlier.)*
2. **Schedule the self-host-vs-Cloud decision** — no longer research-blocked. The measurement: one representative workflow against the $1,000 trial credit, heartbeat contribution instrumented separately.
3. **Decide shard capacity deliberately before the first self-hosted workflow runs** — fixed at build time, *not adjustable later*. Belongs on the pre-worker decision list beside the queue-axis and Worker-Deployment-topology items; same class of one-way door.
4. **Override the default retry policy on every activity wrapping a paid API** — Temporal's default is **unlimited max attempts**.
5. **Amend the Serverless Workers reading in `dedicated_edge_routing.md`** — Lambda caps an activity at 15 minutes and cannot host the 10–60 minute `claude_cli` activity. **Only Cloud Run is shape-compatible.**
6. **Scope `Phase: Temporal Integration` on the basis that no first-party Claude runtime integration is coming.**

**REMOVED candidate.** Prior candidate 14 — *"Refresh `raw/temporal.md`, carried from two prior cycles"* — **done; this cycle is its execution.**

**UNCHANGED.** Candidates 7–19 (the differentiator corrections, the queue-axis standards conflict, the three cheap guards, the don't-build-a-dashboard finding, the Paperclip gate, the eight cost-S doctrine items, the missed-window fix, the stalled predicate, dedupe granularity, subscription-auth framing, two-tier failover). The 2026-08-04 trajectory answer is undisturbed.

**Homeless findings — one new, two strengthened.**

- **NEW: §3's paper contract is never checked on the way IN**, only when a paper ages out. A pre-standard paper inside its window is never checked at all.
- **STRENGTHENED (3rd corroboration): a count read through a summarizing fetch layer is unreliable.** New shape — the GitHub releases API returned **the same 7 objects for `?per_page=15` and `?per_page=100`** while pages 2–3 yielded 12 more. A page total is not a population.
- **STRENGTHENED (2nd measurement): a repaired span is a new claim**, and the round that closes findings gets the least scrutiny. Two independent critic passes is what caught it.
- Carried: the missing "upstream standards amendment we owe" surface (now **four** owed, since the Research Standard is vendored MIRROR); production-feedback channel; sizing-rubric bands.

**Gaps.** *Decide-only disposition* still not run — a refresh has no topic slot, so not a third displacement on the merits, but it is now the oldest unactioned item on the list. New non-topic gap recorded: the billable-Action **measurement**.

---

Per §7 this run wrote **only** inside `docs/standards/architecture/research/` (3 files). No roadmap, phase doc, sprint file, standard, or issue was touched — candidates are surfaced here for the reviewer and operator to route.


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** **Re-scoped `raw/temporal.md`'s topic instead of retiring it or leaving the question alone.** Alternatives: (a) retire — the old question is genuinely dead and its evidence half had been absorbed by three siblings; (b) leave the topic and just re-date the header. Why: §6 retires a topic whose *subject* died, and Temporal is the substrate of a live phase, so (a) misapplies the rule — but more practically, retiring would have deleted the only home for a real unowned question (nobody in the pool stated the licence, priced Cloud, or carried the upgrade obligation). (b) was rejected because a paper that justifies its existence by an open question acquires an interest in that question staying open. **This is the call most worth a human eye** — a reviewer who thinks the vendor-cost question belongs in a phase-level paper rather than at product altitude would retire this instead, and that is a defensible read.
- **[Low]** **Edited the paper's `Topic:`/`Feeds:` headers, when the 2026-08-04 cycle deliberately did NOT edit `combination_prior_art.md`'s and `case_against.md`'s headers after their destinations moved.** Alternative: re-point in `topics.md` only, as that cycle did. Why: those are opposite cases — destination moved while the question stood, vs. the question itself replaced — so a header that preserves an old commission would be recording a question nobody asked. Reasoning written into `topics.md` so the two don't read as drift. A reviewer could reasonably call this inconsistent anyway.
- **[Low]** **Accepted the analyst's resolution of the marker-ambiguity finding (distinguish `wording-uncertified` from `rendered, reduced`) rather than requiring the simpler rejection.** Alternative: reject it — §0 already resolved the ambiguity globally. Why: papers are consumed section-by-section (the same reason §3 requires a `Critic:` line on the paper itself), and a one-token rename fixes it at point of use. Cost: a high-volume mechanical rename, which then **missed two spans** because a marker was split across a line break. The finding was right; the fix carried its own risk.
- **[Low]** **Fixed a duplicated word ("a real, unowned, unowned question") myself rather than opening a round 4.** Alternative: another repair round. Why: a fourth dispatch to fix one word costs more attention than it protects, and the span is not a claim.
- **[Medium]** **Used a FRESH second critic rather than resuming the first.** Alternatives: resume critic 1 (it had already fetched all 33 sources, so cheaper and better-informed); accept round 1 and move on. Why: critic 1 *proposed* several of the repairs, so it could not be the independent check §3 wants on them. This decision is why the invented plan-count discrepancy was caught — critic 1 had supplied the fourth-plan fact that the bad repair was built around. Cost: ~108k tokens and a full re-fetch of all 38 sources.
- **[Medium]** **Treated "bring the paper to §3 contract" as in-scope for a refresh, not just re-validating facts.** Alternative: refresh the facts, re-date the header, and file the contract violations as findings. Why: the paper had no citations at all, so there was nothing to re-validate *against* — a fact-refresh of an unsourced paper produces another unsourced paper. Cost: this turned a one-paper refresh into a 39→799-line rewrite and three critic rounds. A reviewer who wanted a cheap refresh got an expensive one.
- **[Medium]** **Set `Revalidate` to 6 weeks (top of the high band) rather than holding at 4.** Alternative: keep 4 weeks given the topic now covers Cloud pricing and preview maturity labels, which are fast-moving. Why: §5 says a topic that did not move takes its band's max, and it demonstrably did not — two releases in 32 days, none in the last 28, licence unchanged, limits identical to a sibling's independent reading. Mitigated with an on-announcement override trigger list (pricing change, an Anthropic runtime integration appearing, licence change, breaking persistence change) so the longer interval is not a blind spot.
- **[Medium]** **Did NOT edit the two sibling papers whose errors this refresh found**, despite both being inside my write boundary. Alternatives: fix `python_sdk_long_activities.md`'s heartbeat justification and `dedicated_edge_routing.md`'s Serverless Workers reading directly; or say nothing. Why: neither paper is due, neither would get a critic pass in this dispatch, and an edited-but-unverified paper is worse than a correct one carrying a known amendment. Surfaced as action candidates 1 and 5 instead. See Deferred Work — this one does **not** have a tracker and I am not pretending otherwise.

## Deferred Work

One item, and it has **no tracker**. Stating that plainly rather than pointing at something that would launder it.

- **Amendments owed to two sibling papers** — `python_sdk_long_activities.md`'s heartbeat justification (heartbeats are free at the SDK layer, billable on Cloud) and `dedicated_edge_routing.md`'s Serverless Workers reading (Lambda's 15-minute cap rules it out for `claude_cli`). Why deferred: neither paper is due for revalidation, and editing a paper without a critic pass in the same dispatch produces unverified evidence — the exact failure this workflow's gate exists to prevent. **Tracked at: nowhere yet.** They are action candidates 1 and 5 in `synthesis.md` **on this branch**, which is not a valid pointer — that file's live copy on `main` does not contain them, and citing this PR would die at merge. Per §7 the routing path is: this PR is reviewed → the reviewer triages → a **planning** run applies the result. Verified by: `git show origin/main:docs/standards/architecture/research/synthesis.md` → its candidate table has **15 rows**, none of them these; `grep -c "billable\|15 minutes\|shape-compatible"` → **0 matches**; `python_sdk_long_activities.md` appears on **line 36 only**, in the Inputs table, never as a candidate source. (My first check grepped `heartbeat\|Lambda` and returned three hits — all unrelated: Paperclip's liveness-heartbeat design item, a conditional liveness heartbeat, and a wake-on-task Lambda mention. Recording that because a looser grep would have let me write "verified" against the wrong lines.) The candidates exist only on `worktree-research-refresh-20260805-111000`. If this PR is closed unmerged, both findings are lost — that is the honest risk and it is the reviewer's to accept or resolve.

Not deferrals, recorded so they aren't mistaken for one: `temporal.md`'s gaps 4 (patching/versioning cost unpriced) and 6 (OpenAI Agents GA-vs-Preview contradictory across three first-party surfaces) are **stated in the paper with their search methods**, which is §3's designated treatment for a gap. They are findings, not open work.

## Post-Run Reflection

**Friction**

- **`SendMessage` has no foreground mode, and the workflow's dispatch contract depends on one.** The prompt binds me to `run_in_background: false` because a text-only waiting turn kills a headless run — but continuing an existing agent (which is the right move for a repair round, since it holds the paper's context) is only possible via `SendMessage`, which *always* backgrounds. I had to hand-write polling loops in Bash to block the turn, and my **first two attempts terminated early on false signals** (a grep that matched the agent's own transcript, then an mtime that went stale during a long fetch). What finally worked was output-file **size** stability over 220s. This is a real gap between the workflow's stated contract and the available tools, and it cost three extra turns.
- The dispatcher's due-paper list was correct and the mechanical gate did its job — no friction there.

- **§3's paper contract is checked on the way OUT, never on the way IN.** `raw/temporal.md` sat in the pool for a month as the substrate paper for a live phase with no citations, no honest-boundary section and no content arc — three separate "breaking it looks like" clauses. It was caught only because it *aged out*. A pre-standard paper on a `low — 6 months` interval would go half a year unchecked. Worth a conformance check at pool-entry, or a one-time sweep of the remaining 20 papers for the same shape. Raised as a homeless finding in `synthesis.md` since the Research Standard is vendored MIRROR and can't be amended here.
- **Two papers come due around 2026-08-17** (`subscription_economics.md`, `bernstein_capability_mining.md`) and three more around 2026-08-22. If refresh cycles run one-at-a-time at this cost, that queue will not clear.

- **One critic round is not enough for a substantially-rewritten paper, and this run has the measurement.** Round 2 caught the round-1 *repair* inventing a false marketing-vs-documentation discrepancy and then using it to tell readers to distrust a source — a defect strictly worse than the one it closed, which had survived the analyst's own verification. This is the **second** measured instance of repair-conversion in this research programme. `research-refresh.sh`'s Stage 3 says "blocking findings are fixed and re-verified"; it does not say *by whom*, and resuming the same critic is the cheap reading. Suggest the workflow mandate that the re-verification pass be a **fresh** critic when the paper was substantially rewritten, and say so explicitly — I only did it because I reasoned it out mid-run.
- **The workflow prompt's foreground-dispatch contract should say what to do when an agent must be CONTINUED.** It correctly forbids background-dispatch-then-wait, but every repair round after Stage 2 needs exactly that shape. Either bless a documented blocking pattern (output-file size stability worked; mtime and content-grep both produced false positives) or tell the run to dispatch a fresh agent with the repair context instead.
- **Stage 4's write boundary is well-specified and it held**, including under pressure — the refresh found real errors in two sibling papers and a live standards conflict, and the "surface, don't route" rule gave a clean answer each time. No change needed; recording it as a thing that worked.

Issue #39, opened 2026-08-06
Research verification agents have no shell, so every verification-shaped check degrades to fetch-layer evidence — the one class this pool has measured as unreliable


Pumapumapumas:
Surfaced by the `research-verify` correction pass on PR #33 and verified against the agent definitions at `origin/main` `1f995aa`.

## What is wrong

`research-analyst` and `research-critic` are dispatched to do verification-shaped work — "confirm this candidate is absent from `origin/main`", "check this quote against the source repo" — but neither can run a shell:

```
config/agents/research-analyst.md:4:  tools: ["Read", "Grep", "Glob", "Write", "Edit", "WebSearch", "WebFetch"]
config/agents/research-critic.md:4:   tools: ["Read", "Grep", "Glob", "WebSearch", "WebFetch"]
config/agents/research-currency.md:4: tools: ["Read", "Grep", "Glob", "Write", "Edit", "WebSearch", "WebFetch"]
```

No `Bash`. So a prompt that says "run `git show origin/main:<path>`" or "`gh issue view 26`" cannot be executed by the agent it is addressed to.

## The consequence

The agent falls back to `WebFetch`. On PR #33 both analysts did exactly this — one substituted the GitHub JSON API and a local `main` checkout and honestly declared the substitution, and both had to report "I could not verify X" for checks that were trivially runnable with a shell.

That fallback lands the evidence in the **one class this research pool has spent four cycles measuring as unreliable**: the fetch layer, now with five documented failure modes (see #37), two of which survive re-fetching. Verification-shaped work is precisely the work that must not rest on it. The orchestrator then either re-runs the check itself (wasting the dispatch) or accepts an unverified claim.

## Why this is not a one-line config change

Granting `Bash` to `research-critic` would break a boundary that is load-bearing and independently evidenced. The critic is read-only **by design** — no `Write`, no `Edit` — and PR #33's correction pass measured the value directly: two analysts pushed back on premises the orchestrator handed them and both were right, because corrections had to route through the author rather than being applied by the checker. `Bash` restores write capability through the back door.

So the real question is a design ruling, not a flag:

- Does a read-only checker get a **read-only shell** (a constrained command allowlist — `git show`, `git log`, `grep`, `gh ... view`), and is such a grant expressible in the agent frontmatter today?
- Or do verification-shaped steps stay with the orchestrator, and the agent prompts stop naming shell commands they cannot run?

[...]

Issue #26, opened 2026-08-05
standup-tracker — claude-dot-files operating agenda (never closes)


Pumapumapumas:
**This issue never closes.** It is claude-dot-files' **operating agenda** — read FIRST at every standup, before the issue/PR sweep, as the frame that sweep lands inside.

Governed by [Documentation Standard § The Standup Tracker](https://github.com/helloskyy-io/Claude-Dot-Files/blob/main/docs/standards/documentation/documentation_standard.md) (vendored MIRROR from MDC-Master-Planning). Read it before editing.

**Scope: claude-dot-files only.** This workstation touches no MDC / Skyy-Command / SkyyNet repo — those are developed on the skyycommand dev VM and tracked by `mdc-master-planning#140`. Work owed *to* that platform is carried here only as a delivery obligation (T-10).

---

## The three flags

| Flag | Values | Meaning | Who sets it |
|---|---|---|---|
| **`status`** | `ready` · `not-ready` | **Permission to act.** An authorisation, not a fact about blockedness. | **OPERATOR ONLY** |
| **`state`** | `blocked` · `queued` · `in-progress` · `resolved` | Where it sits in completion. | whoever works it |
| **`ownership`** | `PM` · `operator` · `unassigned` | Who has claimed it. | claimed by the taker |
| **`resolved`** | `YYYY-MM-DD` | Completion date. **Only when `state: resolved`.** Makes pruning mechanical. | whoever resolves it |

`not-ready` + `in-progress` is a **detectable violation** — work in flight that was never authorised.

**Item shape:**

```
### T-NN · <short title>
<one line: what it actually is>
`status:` ready|not-ready · `state:` blocked|queued|in-progress|resolved · `ownership:` <who> · `blocked on:` <what, or none>
`resolved:` YYYY-MM-DD          # ONLY when state: resolved
```

### Binding rules

1. **Every unresolved item is ruled on at every standup** — acted on, re-stated with what changed, or explicitly carried WITH the reason it cannot move. "Reviewed the tracker" is not a disposition.
2. **Only what is NOT derivable elsewhere.** Open PRs, open issues and merged work are already swept — never restated here.
3. **Anything with a clean done-state and a proposed action is an ISSUE**, not a line here. Lines marked **⇒ issue-candidate** below are staged here only until they are filed; filing removes the line.
4. **Never a deferral target.** Parking an item here to clear a run's deliverable is a laundered deferral.
[...]

PR #51, opened 2026-08-07
research: fleet-reliability pool — 5 papers, all verified, 16 candidates

Pumapumapumas:
Research pool for the `Fleet Reliability` sprint section (`docs/development/sprint.md:172-184`), which is 📋 QUEUED, NEEDS PLANNING and has no phase doc. This runs **before** planning, per Research Standard §7.

**Tier: Medium — 5 topics, 5 papers.** Sizing reasoning and the deferred milestone are in `topics.md`.

## ⚠️ DRAFT — every paper carries `Critic: not-yet-verified — 2026-08-07`

No critic round was run: this workflow writes papers, and a **separate fresh-context run verifies them**. `synthesis.md` is labelled DRAFT throughout. Several analysts self-reported fetch-layer defects they caught (one discarded two fetches, one as an *"outright fabrication"* contradicted by our own logs) — that is exactly the class the gate exists to catch, and it has not run.

## Artifacts

| File | What |
|---|---|
| `topics.md` | Tier, the 5 topics with destinations, what upstream already covers, and 3 gaps deferred with reasons |
| `raw/durable_dispatch_identity.md` | Dispatch identity, state location, per-subsystem recovery contract, and the Temporal-port constraint |
| `raw/liveness_signal_measurement.md` | Measuring stalled / looping / stranded on a live `claude -p` process |
| `raw/credential_expiry_detection.md` | Detecting an expired subscription session pre-dispatch and mid-run |
| `raw/false_completion_detection.md` | Verifying a run's self-reported success against observable artifacts |
| `raw/blocked_work_notification.md` | Notification channel, delivery semantics, and the operator inbox |
| `synthesis.md` | The deliverable — 8 findings, **16 action candidates**, escalations, gaps |

## The four findings most worth a reviewer's time

1. **The fleet already produces the signals all three guards need, and discards them.** Three analysts found this independently: `claude auth status` exists and (derived) costs no turn while the fleet preflights with a turn-spending `claude -p "ping"`; the stream already carries an undocumented 30-second `tool_progress`/`heartbeat` (125 occurrences across 6 of 57 logs) that nothing parses; `run-claude.sh` computes the completion-contract verdict and throws it away.
2. **The existing credential preflight cannot detect an expired credential by construction.** `check_rate_limit()` runs `2>&1 >/dev/null` — stderr only — while auth failures print on **stdout**. The grep can never match, and control falls into the `return 0` "let the real run surface it" branch that nobody watches on an unattended machine. A defect in shipped code, not a missing feature.
3. **Three milestones need the same write at dispatch time**, which is the mechanical reason the sprint's "designed once" is right: the recovery contract *is* the dispatch record, stranded detection needs a claim record, and false-completion detection needs the pre-state to diff against. Related: identity is minted **inside** the activity in both the bash and Python paths, so under Temporal's default retry it becomes a fresh identity per attempt.
4. **The sequencing in the sprint is wrong in one place, on a correctness argument.** "Blocked" is artifact-*positive*; stalled/looping/stranded are artifact-*negative*. Ship the notifier before the liveness legs and the operator learns that silence means health — the same lesson the dashboard-nobody-opens taught, one step later.

## Escalations — surfaced, filed nowhere

Written into `synthesis.md` per the write boundary. Nothing was written outside `research/`; the product pool was read-only.

| # | Finding | Bears on | What I think it means |
|---|---|---|---|
| E1 | **`--bare` becoming the `-p` default forces a credential-model decision nobody has made.** Anthropic states bare mode *"will become the default for `-p` in a future release"* and that it *"never reads OAuth credentials or the system keychain"*; its own unattended product (GitHub Actions) carries no `/login` session. | Product pool — `subscription_economics.md`, `anthropic_tos_and_enterprise.md`, `edge_identity_trust.md`; the credential-at-the-edge thesis | Build the credential guard anyway (cheap and correct either way), but a product-pool cycle should cost the `setup-token` path before this fleet invests further in subscription-session tooling |
| E2 | **`sprint.md:181`'s "all three guards" is ambiguous between two triples in its own section** (line 180's cheap guards vs. line 182's liveness legs; a third exists upstream). | The fleet-reliability phase doc's scope | One line in the phase doc stating the reading; operator confirms. A research run does not edit planning artifacts |
| E3 | **Upstream's 1–2 day notifier estimate rests on a decayed anchor.** `gh-monitor.enabled: false` with a **2026-08-19 delete decision point**; its routes target `revision.sh`/`revision-minor.sh`, not among the nine scripts in `scripts/workflows/`; its config route keys are inert | Sprint sizing for milestone 4 | Rule on the decision point *before* the phase doc is written; if "delete", harvest the poller skeleton rather than rebuilding it |

E2 and E3 are operator rulings this component needs rather than challenges to what the project believes — listed separately in the synthesis so the escalation count is not inflated.

## Homeless findings

**None.** Candidates 1–15 land in the fleet-reliability phase doc (written by a planning run); candidate 16 is a sprint-item implication, which §7's table routes to the operator. Every candidate has a surface that already exists.

## Verification note

The dispatch's quoted `Fleet Reliability` section **exists on this branch** at `docs/development/sprint.md:172-184` — no unpushed-commit discrepancy. The dispatch's computed currency table reported 0 of 0 papers past window; the pool was genuinely empty (`raw/` held only `.gitkeep`).


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** **Sized this Medium at 5 topics — the top of the band — rather than 3–4.** Alternatives: 3 topics covering only the most decision-blocking milestones; 4 by folding the two "cheap guards" topics into one. Why: the sprint section carries five milestones and four of them name a mechanism that does not exist in the fleet with real alternatives available, so each earned a destination under §6. The workflow prompt is explicitly sceptical of reaching for a band's maximum, and that scepticism may be right here — the two cheap-guard topics (credential expiry, false completion) are the ones a reviewer could reasonably say should have been one paper. They stayed separate because their evidence bases share nothing: one is a Claude Code auth-surface question, the other is an agent-self-report-reliability question.

- **[Low]** **Split the three escalations into one belief-level item (E1) and two "operator rulings needed" items (E2, E3), rather than reporting three escalations.** Alternatives: report all three as escalations; suppress E2/E3 as merely procedural. Why: the prompt states that more than one or two escalations signals upward drift rather than a project in trouble. Three papers independently produced above-altitude findings, and I judged only E1 (the `--bare` credential-model forcing function) to bear on what the project *believes* — E2 and E3 are rulings this component needs before its own phase doc can be written. A reasonable reviewer could call that a distinction I invented to stay under a threshold. The reasoning is stated in the synthesis so it can be overruled.

- **[Low]** **Did not rule on `sprint.md:181`'s "all three guards" ambiguity, and escalated it instead.** Alternatives: adopt the analyst's reading (line 180's cheap guards) as settled in the synthesis; ask nobody and let the phase doc inherit it. Why: it decides a milestone's scope, and scope decisions on the operator's sprint file are not a research run's to make even when the evidence points one way. `durable_dispatch_identity.md` §4.4 records that the contract it designs serves all three candidate triples regardless, which lowers the cost of getting the ruling wrong.

- **[Medium]** **Deferred milestone 5 (per-credential quota headroom) rather than one of the guards.** Alternatives: defer the notifier (largest single build); defer false-completion (the fleet already has a partial guard). Why: quota headroom is the only milestone whose blocking unknown was already discharged upstream (`hermes_assessment.md` §5.1 — headroom is derivable from cap-errors), it is last in the sprint, and it constrains nothing else. Deferring the notifier instead would have been defensible on build size, but the notifier is the milestone whose *cost anchor* turned out to be decayed (E3), so leaving it unresearched would have left that undiscovered.

- **[Medium]** **Excluded the safety-hook wiring test — the third "cheap guard" — from the topic list entirely rather than making it a sixth topic or folding it in.** Alternatives: a sixth topic; a section inside the credential paper. Why: the mechanism is already named in `sprint.md:168`'s Managed Configuration blocker note, so a paper would have restated a known fact — it needs a test written, not evidence gathered. This call was independently vindicated and complicated: `false_completion_detection.md` §5.2 arrived at the conclusion that the false-completion guard needs its own wiring test of *identical shape with an identical `--setting-sources`/`--bare` dependency*, which is an argument for planning the two together (candidate 16), not for researching the hook test separately.

- **[Medium]** **Gave every analyst an explicit "already settled upstream — cite, do not re-derive" list naming specific papers and sections, which narrowed their scope substantially.** Alternatives: point them at the upstream synthesis and let them judge; give no upstream context. Why: the workflow's altitude constraint makes re-deriving a settled question actively harmful (a second answer that can drift from the first). The cost is real and worth naming: the credential analyst was told not to re-run the industry survey on unattended re-auth, so this pool cannot independently corroborate that upstream finding — it inherits it.

- **[Medium]** **Surfaced candidate 1 — re-sequence the milestones 2 → (1,3) → 4 — as a `change direction` candidate against the operator's sprint ordering.** Alternatives: report it as a neutral observation; leave it inside the notifier paper. Why: the argument is a correctness one, not a preference (`blocked_work_notification.md` §1.3 — shipping an artifact-positive detector alone teaches the operator that silence means health), and burying it in one paper's §1.3 would mean the phase-doc author builds in sprint order without seeing it. It is surfaced, not applied; `sprint.md` was not touched.

## Deferred Work

- **Per-credential quota headroom research (milestone 5)** — Why deferred: Research Standard §2 caps a cycle at ~5 topics; this is the least decision-blocking of six candidates and its blocking unknown was already discharged upstream. Tracked at: `docs/development/sprint.md` line 184, the open milestone `- [ ] **Per-credential quota headroom** — derived from observed cap-errors, with no provider telemetry`. Verified by: `git show origin/main:docs/development/sprint.md | grep -n "quota headroom"` → returned `184:- [ ] **Per-credential quota headroom** — derived from observed cap-errors, with no provider telemetry` on the default branch, unchecked. (`topics.md` in this PR carries the deferral reasoning and the cycle-2 pointer, but it does not exist on `main` yet — `git ls-tree origin/main docs/development/fleet-reliability/ -r --name-only` returns only `research/README.md` and `research/raw/.gitkeep` — so the sprint line is cited as the tracker, not `topics.md`.)

- **Safety-hook wiring test (third of the three cheap guards)** — Why deferred: not deferred *work*, deliberately not a research topic — it needs a test written, not evidence gathered, and the mechanism is already named in `sprint.md:168`. Tracked at: `docs/development/sprint.md` line 180, the open milestone `- [ ] **Three cheap guards** — credential expiry, false completion, and a safety-hook wiring test`. Verified by: `git show origin/main:docs/development/sprint.md | grep -n "safety-hook"` → returned `180:- [ ] **Three cheap guards** — credential expiry, false completion, and a safety-hook wiring test` on the default branch, unchecked.

- **Everything else this run produced is surfaced, not deferred.** The 16 action candidates and 3 escalations live in `synthesis.md` and this PR body for the reviewer and operator to triage — per §7 that is the finished state of a research run, not a deferral, and no pointer is claimed for them.

## Post-Run Reflection

**Friction**

- **`Write` with a relative path silently wrote to the wrong location.** `synthesis.md` landed at `research/raw/docs/development/fleet-reliability/research/synthesis.md` because the relative path resolved against a CWD left over from an earlier `Bash` call that had `cd`-ed into `raw/`. The workflow prompt contains both *"Bash CWD persists between calls — treat it as unpredictable"* and *"Prefer relative paths inside the worktree for Read/Grep/Glob/Edit/Write."* Those two rules are in direct tension for `Write`, and the failure is silent — `git status` was the only thing that caught it. Had I committed without checking, the PR would have shipped a synthesis at a path no consumer reads, while `README.md` documents `synthesis.md` as *"the only file in here the phase doc should need to cite."*
- **A subagent's return was flagged by the harness for an instruction-shaped pattern (`bypass-permissions`).** It was a false positive on legitimate research content: `durable_dispatch_identity.md` correctly reports that Claude Code does not restore `bypassPermissions`, `--mcp-config`, `--settings`, `--plugin-dir` or `--add-dir` on a resumed session, which is the paper's central argument against auto-resume. Relaying it per the harness instruction; no action needed, but a research pool about a permissions-bypassing fleet will keep tripping this.
- **The dispatch's paper-currency table was empty (0 of 0) because the pool was new.** Correct behaviour, but the surrounding prompt text is written for the re-run case ("use these verdicts verbatim", "a prior synthesis is a consumable") and reads oddly on a cycle-1 run. Minor.

- **`scripts/workflows/activities/run-claude.sh`'s `check_rate_limit()` has a real defect that this research found incidentally** — `2>&1 >/dev/null` captures stderr while Claude Code prints auth failures on stdout, so the rate-limit grep can never see them and an expired credential falls into the `return 0` pass-through branch. It is candidate 2 in the synthesis, it is minutes to fix, and it is independent of whether the Fleet Reliability sprint is planned. Worth fixing outside this PR rather than waiting for the phase doc.
- **`gh-monitor`'s 2026-08-19 delete decision point is twelve days out** and its config route-enable keys (`enable-build`, `enable-build-minor`) are not the keys the script reads (`enable-revision`, `enable-revision-minor`), so they are inert. Ruling on it before the fleet-reliability phase doc is written would prevent the notifier being sized against an anchor that may be deleted.
- **The scaffolded `research/README.md` is good and worth keeping as the convention** — it made the artifact contract legible before the standard was read.

- **The research workflow prompt should require absolute paths for `Write` specifically**, or drop `Write` from the "prefer relative paths" list. The current pairing produced a silent misplacement of the run's single stable consumption artifact. A one-line change: *"use absolute worktree-rooted paths for Write — CWD is unpredictable and a misplaced write is silent."*
- **Consider having the workflow verify the three artifact paths exist at their contract locations before the final commit** — `topics.md`, `raw/*.md`, and `synthesis.md` at exactly the dispatched research dir. The prompt already requires a `git status` check for untracked files; extending it to "confirm `synthesis.md` is at `<research-dir>/synthesis.md` and nowhere else" would have caught this mechanically rather than by luck.
- **The Stage 2 sizing guidance leans hard against reaching a band's maximum, but gives no test for when the maximum is genuinely correct.** §2's own test (one destination = one component) says 5 was right here, yet the prompt's tone says a 5-topic list is probably a misread. A sentence reconciling them — e.g. *"the band max is correct when every topic's destination is the same single doc and each names a distinct mechanism"* — would remove the ambiguity. This run resolved it by justifying against the destination and stating the reasoning in `topics.md`, which is what the prompt asks for, but the tension cost real deliberation.

Pumapumapumas:
# research-verify — fleet-reliability pool

All five papers went through the critic gate and are corrected. `synthesis.md` was rewritten against the corrected pool and independently verified. The DRAFT/unverified banner is gone.

| Paper | Rounds | Final verdict |
|---|---|---|
| `durable_dispatch_identity.md` | 2 | **PASS** (critic round 2) |
| `liveness_signal_measurement.md` | 4 | **PASS** (critic round 4) |
| `credential_expiry_detection.md` | 4 | **PASS** (critic round 4) |
| `false_completion_detection.md` | 4 | **PASS** (critic round 4) |
| `blocked_work_notification.md` | 5 | **PASS** (critic round 5) |
| `synthesis.md` | 1 | **PASS** |

Each paper's own header carries `Critic: PASS-WITH-FIXES (<what was fixed>) — 2026-08-07`, per §3 — every paper needed repair, and the final critic round on each returned clean. No paper was dropped for non-convergence; no correction budget was exhausted (ceiling 6, max used 5).

**Verification moved real content — it was not a rubber stamp:**

- **One paper's headline conclusion was inverted.** `liveness_signal_measurement.md` claimed the `claude -p` heartbeat was undocumented and escalated that as a portfolio risk. It is first-party documented — `SDKToolProgressMessage.heartbeat?: boolean`, 30-second cadence, gated at Agent SDK v0.3.214. The original NOT-PRESENT finding came from a fetch layer that silently truncated a 4,830-line document. The same doc also states the heartbeat does **not** cover the Agent tool *or tool calls inside a subagent* — a coverage hole the logs alone could not reveal, and now candidate 10's primary driver.
- **A guard gained a mandatory safety requirement it did not have.** Candidate 3 proposed replacing the turn-spending preflight with `claude auth status`. Claude Code v2.1.139 documents a deadlock where expired credentials blocked that exact subcommand "with no way to recover" — i.e. it can hang on precisely the state it is being asked to detect. Candidate 3 now mandates a `timeout` treated as park-and-notify, never pass. As drafted it would have put an unbounded call in the fleet's preflight path.
- **A gap was deleted outright.** Candea & Fox's *Crash-Only Software* was recorded as unretrievable across five hosts; it retrieves fine, and now backs three sections of `durable_dispatch_identity.md`.
- **Correction ran in both directions.** Analysts rejected critic findings after re-checking primary evidence, and the rejections held — three times (the 2-of-4 `subagent_retry` count; the claimed CHANGELOG gate on `heartbeat`, which lives in the reference page instead; and AWS SQS's source form, which turns out to be `Accept`-header content negotiation).

## Decision Log

- **[Low]** Ruled the Candea & Fox vs. Meyer 1992 confidence-tier difference is **not** a pool-level calibration inconsistency, and did not reopen `durable_dispatch_identity.md` to upgrade its spans. Alternatives: reopen the PASSed paper and promote the tier for consistency; or flag it as an open inconsistency. Why: the evidentiary events genuinely differ — Meyer's spans came from deterministic `pdftotext` extraction matched character-for-character against two independently hosted mirrors, Candea & Fox's from a visual page-renderer read that cannot be certified character-exact. Both papers applied the same rule correctly to different situations. The synthesis critic adjudicated this independently and reached the same conclusion. Reopening would have cost two rounds to move no conclusion, and the current tier errs conservative.
- **[Low]** Granted `blocked_work_notification.md` a fifth round on §1.2 alone, at the analyst's own request, after its item list had emptied. Alternatives: close at round 4. Why: §1.2 had produced a defect in every round, and the analyst's argument — that an empty item list is not evidence of convergence — was correct. The round found a real one: the analyst's own round-3 prose asserted a corpus-wide absence from a single directory listing, disproved by the `:doc:` link target inside the sentence it was quoting.
- **[Medium]** Accepted three analyst disputes against critic findings, after independently re-checking each myself rather than deferring to either party. Alternatives: side with the critic by default (the gate's usual presumption). Why: on all three the analyst was right, and a gate that cannot be corrected by evidence is not a gate.
- **[Medium]** Directed analysts to record critic-obtained provenance explicitly ("the exact characters were obtained by a raw non-summarizing `curl` during the verification pass; this paper's own fetch layer cannot reach line 4,388") rather than writing those spans as their own fetches. Alternatives: have them assert the spans plainly, which is laundering; or drop the claims, which discards settled facts. Why: the analyst has no Bash and physically cannot reach the source. Naming who retrieved what is traceable under §3 and is the truth.
- **[Medium]** Ran correction rounds by resuming the original analysts via `SendMessage` + blocking `TaskOutput` rather than spawning fresh ones. Alternatives: a fresh analyst per round. Why: resume preserves the paper's authoring context, which matters most on rounds 3–5 where the fixes are narrow and the risk is collateral churn. Cost: `SendMessage` backgrounds the agent, so each round needed a blocking bridge.
- **[Medium]** Overrode `liveness_signal_measurement.md`'s CONTESTED/directional hedge on the heartbeat and required it un-hedged to definitive. Alternatives: leave it contested, since the authoring analyst could not reproduce the fetch. Why: three independent non-summarizing retrievals settled it, including my own. Leaving it contested would have left a false gap in the synthesis and understated the design's evidentiary footing — the calibration error is less dangerous than the fabrication it guards against, but it is still wrong.

## Post-Run Reflection

**Friction**

- **The analyst/critic tooling asymmetry drove most of this cycle's correction rounds.** `research-analyst` has `WebFetch` but no `Bash`; `research-critic` has both. The summarizing fetch layer silently truncates large first-party documents (it returned `[Content truncated due to length...]` on a 263 KB SDK reference), mis-associates changelog entries with the wrong release heading, and cannot extract embedded OCR text from Capture-processed PDFs. Every one of those failures was written into a paper as a fact about the *source*. This produced **ten defect instances across all five papers** — enumerated in `synthesis.md`'s "How much to trust a negative finding in this pool" section, which is now the pool's calibration note. The critic could settle each one in a single `curl`.
- **No intermediate commits between correction rounds.** Two critics independently flagged that `git diff <base>` could not isolate a single round's changes, because rounds 1–5 all sat uncommitted on top of the original pool commit. Both worked around it; both said so.
- **Four of nine critic dispatches wrote files to `/tmp` despite explicit read-only instructions in their prompts** (`curl -o`, `>` redirect). All four disclosed it unprompted, which is the system working — but `research-critic` has `Bash`, so its read-only mandate is prompt-enforced only. Stray files left behind: `/tmp/does_not_exist_dummy`, `/tmp/changelog_main.txt`, `/tmp/tasks_rst_check.txt`. (`/tmp/cc-changelog.md` is mine, from my own verification.) I did not delete any of them.

- All five papers were validated on the same day with 3–6 week windows, so the whole pool comes due within roughly three weeks of itself. That is one refresh cycle rather than a staggered drip — consistent with §2's ~5-topics-per-cycle cap, so probably fine, but worth knowing before the first refresh fires. Noting it plainly rather than filing it; it has no natural home and an invented one would be worse.
- The synthesis records that `durable_dispatch_identity.md`'s header does not enumerate its round count, unlike the other four. Harmless, but a consistent `Critic:` convention across the pool would make round counts readable at a glance.

**Tooling-level suggestions (claude-dot-files)** — flagging for the architecture session, not editing:

1. **Give `research-analyst` a raw-retrieval path, or make the research workflow's verification step mandatory before a paper's first negative finding is written.** This is the highest-leverage item in this reflection. A negative finding is the one claim class the analyst's tooling is structurally unable to establish, and the Research Standard §3 already requires negative findings to state their search method — but a method executed through a summarizing layer produces a confidently-wrong result that reads exactly like a correct one. Ten instances in one five-paper cycle is not a tail case.
2. **The standard's "a negative finding states its search method" rule may need a companion clause:** *a search method delegated to a summarizing retrieval layer does not establish an absence.* The papers here all stated their methods honestly; the methods were simply not capable of supporting the conclusions. This is a §3 amendment candidate for the human-in-the-loop standards path, surfaced not filed.
3. **Consider having the research workflow commit after each correction round.** It costs one commit per round and gives every subsequent critic a clean diff anchor.
4. **`research-critic`'s read-only property is prompt-enforced, not tool-enforced.** Four violations in nine dispatches suggests the prompt is not sufficient. Removing `Bash` would break the raw-retrieval capability that made this cycle work, so the fix is probably a sandboxed or write-blocked shell rather than no shell.

Pumapumapumas:
# PR Review — disposition pass 1

**Verdict: HOLD — needs-assistance.** Two operator actions, neither of them about the research. The pool itself is clean: all six research blocking checks pass.

## The six blocking checks

| # | Check | Result |
|---|---|---|
| 1 | Citations resolve, claims match source | **PASS** — critic gate ran 2–5 rounds per paper (`957d473`); fresh-context sample of 14 cited URLs across the pool returned HTTP 200 on all 14 |
| 2 | Internal links resolve | **PASS** — the one relative markdown link (`README.md` → `../../../standards/research/research_standard.md`) resolves; all 38 backticked repo paths across the pool exist. Four unresolved-looking paths (`docs/design/design.md`, `docs/confirms.md`, `docs/gateway/restart-recovery.md`, `docs/tools/loop-detection.md`) are Kafka / RabbitMQ / OpenClaw paths, each correctly attributed with its own URL |
| 3 | Header block present and machine-parseable | **PASS** — all five carry `Topic` / `Feeds` / `Last validated` / `Revalidate` / `Confidence` / `Critic` |
| 4 | Honest-boundary section present | **PASS** — all five (`blocked` §5, `liveness` §6, `credential` §6, `false_completion` §5, `durable` §7 + §8) |
| 5 | No factually false statement about our platform | **PASS** — ten load-bearing platform claims verified against the tree: `check_rate_limit()`'s `2>&1 >/dev/null` at `run-claude.sh:82` and its `return 0` fall-through at :90; `.gitignore` line 2 = `.claude/`; `config.yaml` `gh-monitor.enabled: false` + the 2026-08-19 decision point; `gh-monitor.sh` reading `enable-revision*` while `config.yaml` sets `enable-build*` (inert, confirmed); exactly nine `scripts/workflows/*.sh`, with `revision.sh`/`revision-minor.sh` absent; `sprint.md` lines 172/180/181/182/183/184 exact; `sprint.md:168`'s `block-dangerous.sh` note; `notify-done.sh` firing `notify-send` on `Stop`; `assistant_activities.py:252-253` minting the stamp inside the activity; `run-claude.sh`'s `COMPLETION_PATTERN` verdict computed and never persisted |
| 6 | Source-count floor (10–20 for medium+) | **PASS** — 45 / 39 / 34 / 26 / 25 |

## Disposition table

|---|---|---|---|
| The PR's headline says this pool is unverified when it is verified (`stale-draft-label`) | doc-drift | **HOLD** — needs-assistance | Title reads *"(DRAFT, unverified)"* and the body says *"every paper carries `Critic: not-yet-verified`"* and *"No critic round was run."* Two independent checks say otherwise: `grep -rn "not-yet-verified"` across the pool returns exactly one hit — `synthesis.md:7` stating that **none** carries it — and every paper's header carries `Critic: PASS-WITH-FIXES`. `grep -rni draft` finds no status label left in any artifact. The `research-verify` comment in this thread already says *"The DRAFT/unverified banner is gone"*; the PR's own surface was never updated |
| The critic's mandatory clone-and-grep is forbidden by the same file's absolute write ban (`critic-write-ban-blocks-clone`) | correctness | **HOLD** — needs-assistance | On `origin/main`, `config/agents/research-critic.md` line 12: *"You must not write, anywhere, by any means… **Not to the repo, not to `/tmp`, not to a scratch path**"*, framed as *"absolute and not a preference."* Lines 59–65 of the same file: *"A git-hosted source is CLONED and grepped, never fetched. This is the strongest verification available and it is now **mandatory** for that source class"* — via `git clone --depth 1 --filter=blob:none <repo> /tmp/verify-<name>`. Both shipped in `10dc24e`. This is the run's reflection item 4 (*"read-only is prompt-enforced… four violations in nine dispatches"*), re-checked against the current tree, where the remedy now contradicts itself |
| A summarizing-layer absence is admissible evidence under Research Standard §3 (`research-std-absence-clause`) | standards-implication | **DEFERRED** — filed | The standard closed this hole for quotations (§3 *"Verbatim means the exact characters were returned"*) and left it open for absences, which need only *state* a method, not a capable one. Ten instances in this one cycle, one of which inverted `liveness_signal_measurement.md`'s headline. Vendored MIRROR — homeless by construction here. Filed upstream: **[MDC-Master-Planning#143](https://github.com/helloskyy-io/MDC-Master-Planning/issues/143)** (verified open). Upstream re-checked at `main` = `644971e`, blob `68fd083`: no such clause exists, and neither open upstream issue (#140, #142) covers it |
| `research-analyst` cannot establish an absence its papers assert (`analyst-fetch-layer-gap`) | friction | **FIXED** | The run's own *"highest-leverage item in this reflection."* Verified already shipped on `origin/main` at **`10dc24e`** (2026-08-07 15:43 UTC, `Closes #37 and #39`), which landed 23 minutes *before* the reflection was written from a worktree branched earlier: `git show origin/main:config/agents/research-analyst.md` line 4 now reads `tools: [… "Bash" …]`, with a read-only discipline and a mandatory clone-and-grep path. Issue #37 is CLOSED. Prescribing this would have re-asked a question answered the same afternoon |
| Milestone 5, per-credential quota headroom, not researched (`deferral-quota-headroom`) | deferral | **DEFERRED** | Pointer VERIFIED on the default branch, not on this PR: `git show origin/main:docs/development/sprint.md` → line 184, `- [ ] **Per-credential quota headroom** …`, open and unchecked. `topics.md` records the cycle-2 plan against this same directory |
| Safety-hook wiring test not researched (`deferral-safety-hook-test`) | deferral | **DEFERRED** | Pointer VERIFIED on the default branch: `git show origin/main:docs/development/sprint.md` → line 180, `- [ ] **Three cheap guards** — credential expiry, false completion, and a safety-hook wiring test`, open and unchecked. Candidate 16 additionally routes it to the operator as a sprint-item implication |

**Laundered deferrals: 0 of 2.** Both deferrals point at open sprint milestones on `origin/main` and both were verified by opening them. The run's third Deferred Work bullet correctly claims *no* pointer for the candidates and escalations, which is the honest statement rather than a deferral.

## Cargo — carried through untouched

Per the review boundary, a research synthesis surfaces candidates **by design**; they are the deliverable, not defects. Enumerated so they are visible, **not ruled on here**:

- **16 action candidates** (`synthesis.md` § *Action candidates*) — 1–15 destined for the fleet-reliability phase doc, 16 a sprint-item implication for the operator.
- **1 belief-level escalation** (E1, `--bare` becoming the `-p` default forcing a credential-model decision) and **2 operator rulings** (E2 the `"all three guards"` ambiguity; E3 the decayed notifier cost anchor). E3's underlying timer already has a home — `#26` **T-13 · `gh-monitor` deletion decision — ~2026-08-19**, verified present.
- **2 clarifications for other pools** (the SQLite multi-machine caveat; `--include-partial-messages` as an unexercised liveness lever).

Ruling on these is the planning step this pool feeds, in a pass with more context. Holding the pool for them would make standards a predecessor of research, which is backwards — and a 3-week-window paper burns its own revalidation budget while it waits.

## WHAT HAPPENS NEXT

Two operator actions. Neither touches the branch; neither is worth a dispatch.

**1 — `stale-draft-label` · correct the PR's title and body** *(~30 seconds)*

> `reframe:` Is the problem that the body is out of date, or that verification status is asserted in two places and only one was updated when it changed?
> `bp:` A merged PR's title and body are the durable record of what merged. A status label that contradicts the artifact is a false statement at the most-read surface — and after merge it is what git history answers *"was this pool ever verified?"* with.
> `recommendation:` `gh pr edit 51` — drop `(DRAFT, unverified)` from the title, and replace the `⚠️ DRAFT` section of the body with the verdict table already posted in the `research-verify` comment. Nothing else in the body needs to change; its four headline findings and the escalation table all still verify.

**2 — `critic-write-ban-blocks-clone` · resolve the contradiction in `research-critic.md` on `main`** *(one or two sentences)*

> `reframe:` Is this a wording nit, or did the commit that made clone-and-grep mandatory simultaneously forbid the only place it can write?
> `bp:` A prohibition protects an **invariant**, not a mechanism. The invariant here is stated in the file itself — *"an actor that can fix a defect and then declare it verified is verifying its own work"* — and a scratch clone read only by `grep -F` does not touch the artifact under review. Name the invariant and carve out the scratch path; leaving an absolute rule that the mandatory procedure violates means the next critic either skips the strongest check available or violates an "absolute" rule on its first use. Last cycle already measured 4 of 9 dispatches writing to `/tmp` and self-disclosing.
> `recommendation:` Amend `config/agents/research-critic.md`: keep the ban absolute for the repo, the worktree, and any path a consumer reads, and carve out an explicitly-named scratch clone (`/tmp/verify-<name>`) usable **only** as a read target, never as an output — no `>`, no `tee`, no editing inside it. Operator's call whether to edit directly or fire a scoped dispatch; it is a claude-dot-files agent definition, so it is not mine to write.

Once those two land, this is a MERGE. Item 2 is independent of the pool's correctness — stated so the sequencing is the operator's to choose, not because it can be skipped.

```yaml
pr_review:
  pr: 51
  pass: 1
  attempt: 2
  verdict: HOLD
  converged: false
  findings:
    - id: stale-draft-label
      title: the PR's title and body assert this pool is unverified, and it is verified
      category: doc-drift
      consequence: the title merges into git history recording "(DRAFT, unverified)" for a pool where all five papers carry Critic PASS-WITH-FIXES; a reader or planning run checking whether the pool was ever gated gets the opposite of the truth from the PR's most-read surface, and may discount the evidence or re-run a verification that already happened
      disposition: hold
      remedy: operator-action
      hold_kind: needs-assistance
    - id: critic-write-ban-blocks-clone
      title: the critic's mandatory clone-and-grep verification is forbidden by an absolute write ban in the same file
      category: correctness
      consequence: the next research-critic dispatch faces an absolute prohibition and a mandatory procedure that collide, so either the strongest verification class silently reverts to fetching - the only check that catches near-duplicate blending, the fabrication class that once survived two months and propagated through five documents - or the read-only property the verdict's meaning rests on is violated on its first use
      disposition: hold
      remedy: operator-action
      hold_kind: needs-assistance
      reviewed_sha: 10dc24e96c69d9190da965f28e46e2a957defa59
    - id: research-std-absence-clause
      title: a negative finding produced by a retrieval layer that cannot establish absences is admissible evidence under Research Standard 3
      consequence: an absence is the one claim class a consumer cannot sanity-check, and section 3 requires only that a search method be stated, not that it be capable - so a summarizing-layer absence satisfies the rule as written and a critic has no rule to fail it on; ten instances landed in this single cycle and one inverted a paper's headline conclusion and drove a portfolio-level escalation
      disposition: deferred
      pointer: https://github.com/helloskyy-io/MDC-Master-Planning/issues/143
      pointer_verified: true
      reviewed_sha: 644971ef3a4deda4332350aed042ca097c1a97af
    - id: analyst-fetch-layer-gap
      title: research-analyst had no raw-retrieval path, so every absence it asserted rested on a layer that cannot establish one
      category: friction
      consequence: would have been ten false-absence instances per cycle absorbed by multi-round critic correction - except it is already fixed, and prescribing it would have re-asked a question answered the same afternoon
      disposition: fixed
      remedy: none
      reviewed_sha: 10dc24e96c69d9190da965f28e46e2a957defa59
    - id: deferral-quota-headroom
      title: milestone 5 has no research coverage and cycle 2 must pick it up
      category: deferral
      consequence: without a tracked home the last milestone gets planned with no component-level provider-error taxonomy, or re-derived from scratch next cycle
      disposition: deferred
      pointer: "docs/development/sprint.md line 184 on origin/main - open, unchecked milestone: Per-credential quota headroom"
      pointer_verified: true
    - id: deferral-safety-hook-test
      title: the third cheap guard has no research coverage and needs a test written rather than evidence gathered
      category: deferral
      consequence: without a tracked home the safety-hook wiring test is silently dropped from the three cheap guards the sprint milestone names
      disposition: deferred
      pointer: "docs/development/sprint.md line 180 on origin/main - open, unchecked milestone: Three cheap guards"
      pointer_verified: true
  next_steps:
    - item: stale-draft-label
      kind: needs-assistance
      note: gh pr edit 51 - drop "(DRAFT, unverified)" from the title, replace the DRAFT section of the body with the research-verify verdict table already in this thread
      why_human: operator-action
      reframe: is the problem that the body is stale, or that verification status is asserted in two places and only one was updated when it changed
      bp: a merged PR's title and body are the durable record of what merged, and a status label contradicting the artifact is a false statement at the most-read surface
      recommendation: |
        Edit the PR title and body directly. Title: drop the "(DRAFT, unverified)" suffix.
        Body: replace the "⚠️ DRAFT — every paper carries Critic: not-yet-verified" section with
        the five-row verdict table from the research-verify comment. The rest of the body still
        verifies as written and needs no change.
    - item: critic-write-ban-blocks-clone
      kind: needs-assistance
      note: research-critic.md forbids writing to /tmp absolutely while mandating git clone into /tmp/verify-<name>; carve out the scratch path or the strongest verification check is unusable by rule
      why_human: operator-action
      reframe: is this a wording nit, or did the commit that made clone-and-grep mandatory simultaneously forbid the only place it can write
      bp: a prohibition protects an invariant, not a mechanism - the invariant is that the critic never modifies the artifact under review, and a scratch clone read only by grep -F does not touch it
      recommendation: |
        Amend config/agents/research-critic.md on main. Keep the write ban absolute for the repo,
        the worktree, and any path a consumer reads. Carve out one explicitly-named scratch clone
        (/tmp/verify-<name>) usable ONLY as a read target for grep -F - never as an output, no
        redirects, no editing inside it. One or two sentences. It is a claude-dot-files agent
        definition, so it is the operator's to write or to dispatch; a scoped build-minor run
        against main would also fit, but the fix is smaller than the dispatch.
    - item: research-std-absence-clause
      kind: file-issue
      note: vendored MIRROR standard - the amendment cannot be written in the consuming repo, so it is homeless by construction and was filed upstream where the standard lives
      issue_url: https://github.com/helloskyy-io/MDC-Master-Planning/issues/143
      issue_repo: helloskyy-io/MDC-Master-Planning
      qualified: unrelated + substantial + not-already-covered
  laundered_deferrals:
    caught: 0
    of_total: 2
  homeless_items: 1
  redispatched: false
```

## Post-Run Reflection

**Notes that did not clear the consequence gate** — recorded here rather than as findings, because nothing breaks if they are never addressed:

- `synthesis.md:7` states round counts (3/3/3/4/unstated) that are each one lower than the `research-verify` comment's table (4/4/4/5/2). The papers' own `Critic:` headers agree with the synthesis, so the two are almost certainly counting fix-rounds versus total critic rounds including the final clean pass. It understates verification depth, which is the conservative direction.
- `synthesis.md:45` compresses the completion-contract finding to *"computes … and discards it."* The verdict is computed and then **not persisted** — `run-claude.sh` does fail the run loudly on a miss. `durable_dispatch_identity.md` §4.2 states it precisely (`"nowhere yet"` in the state-location column) and §4.3 R3 is exact, so the pointer a phase-doc author follows lands on the accurate statement.
- All five papers were validated on 2026-08-07 with 3–6 week windows, so the whole pool comes due within ~3 weeks of itself. The `research-verify` run noted this plainly and declined to invent a home for it; that was the right call and this is the same note, not a finding.

**Friction / tooling-level suggestions for `review-pr` itself:**

- **The research-review axis and the current-tree check are the two rules that did all the work this pass, and they pull in opposite directions.** The axis says expect zero findings and resist enumerating; the current-tree check says go verify the prescription hasn't already landed. Both fired here: the axis correctly kept 16 candidates and 3 escalations out of the runway, and the current-tree check killed what would otherwise have been this review's headline finding — `research-analyst` has no shell — which had shipped on `main` 23 minutes before the run that surfaced it finished writing. Worth stating in the prompt that a reflection item from a worktree-based run is a claim about the branch point, not about `main`, and always needs the current-tree check before it becomes a finding.
- **The disposition enum has no state for "live defect, elsewhere on the default branch, too small to file, no existing home."** `noted` is barred (there IS a live defect), `deferred` needs a home, the filing authority's substantiality bar correctly rejects it, so `hold` is the only honest option — and that holds a research PR for something independent of it. `critic-write-ban-blocks-clone` is exactly that shape. It resolved fine here because both HOLD items happened to be one-minute operator actions, but the enum forced a heavier verdict than the item warrants. Flagging as a possible schema gap rather than working around it.

PR #54, opened 2026-08-07
plan-revision: plan the Memory Management Framework — roadmap + 5 phase docs

Pumapumapumas:
Plans the **Memory Management Framework** from its merged, critic-verified research. Six new docs; every checkbox unchecked. No code, no standards, no `sprint.md` edits.

## Planning changes made

- **`memory-management-framework/roadmap.md`** — status, five phase paragraphs with 3–5 completion criteria each, dependencies in both directions, a **Key Decisions** block (14 rulings, each with evidence, rejected alternative, and cost), a **Standards-amendment candidates** section (5 candidates, surfaced not written), and a table of where the research contradicted the sprint milestones.
- **`phase1_measure_the_channel.md`** — six experiments (E1, E2, E3, E5, E6, E7); three can cancel downstream work. Includes a `§Runtime Verification` block.
- **`phase2_kind1_framework.md`** — Kind 1 written down as an interface, including **retrievability**, which closes a CPI deferral whose watch-criteria name this doc.
- **`phase3_typed_exit_record.md`** — arrangement A, split abstention, the fail-safe contract, the **channel's** properties, and the single-pair proof.
- **`phase4_fleet_migration.md`** — fleet-wide rollout across **both** fleets, capability-parity audit.
- **`phase5_convergence_stopping.md`** — a class-(iii) asserted convergence flag replaced by a class-(ii) computed delta.
- **`docs/file_structure.txt`** — component entry added; the annotation asserting a single doc shape amended.

## Why five phases

The sprint entry lists four milestones; the plan has five, and the extra one is measurement. Two of the research's experiments can *cancel* work — if `is_error` never disagrees with the propagated exit status, milestone 1 is a no-op; if the prose grep has never missed, the transport upgrade buys nothing measurable. Folded into a design phase, those are gates the same run walks past on its way to the design it already intended to write. The other splits: documenting Kind 1 is an authoring task with its own done-state; design-and-prove-one-pair is depth while fleet migration is breadth, and bundled they'd be indistinguishable in verification; convergence is a separate capability whose failure mode (firing early, silently) has no natural alarm.

## Where the research contradicted the sprint milestones

- **The parent already gates on more than the milestone claimed.** `build.sh:60` is `set -euo pipefail` under `tee`, and `run-claude.sh:167,201-204` read `subtype` and `.result`. Real gap: `is_error`, `permission_denials[]`, `num_turns`-vs-cap — plus the sharper one the milestone omits, that no first-party exit-code table exists, so class-(i) routing rests on an undocumented mapping.
- **The abstention member must split in two** — a computed *could-not-check* and an asserted *needs-a-ruling*. Not in the sprint item; it changes the vocabulary.
- **Routing on non-model values is mature and boring outside the agent corpus.** Three first-party vocabularies converge; Kubernetes `podFailurePolicy` is a ready-made fail-safe contract. Borrow the shape.
- **The LLM-producer premise is half right.** CI has measured well-formed-wrong results too; what differs is defect-with-a-fix vs stationary-rate-with-a-distribution, which changes the fail-safe contract, not the taxonomy.
- **Convergence already partly ships** — a milestone-level correction the sprint entry does not contain, and one the *research* also missed (see below).

## Deviations from plan

- **The dispatch's premise that Documentation Standard §0/§1/§2 is binding here is wrong**, and standards-architect + quality-control both caught it. `docs/standards/documentation/README.md:34` lists *Development Planning Files* — the section containing §0, §1, §2, Live-Runtime Verification and Capability-Parity — under "What does NOT bind here." The local authority is `sprint.md:28`: *"A component that outgrows one phase gets its own `roadmap.md` plus numbered `phaseN_<name>.md` files."* **The shape is unchanged** — it is independently correct under the local convention — but the cited authority is now `sprint.md`, and the `§Runtime Verification` / `§Capability Parity` blocks are labelled adopted-practice rather than binding.
- **The dispatch's claim that the migration changes none of the prompt content is false.** `review-pr.sh:427` is a stage titled "PRINT THE VERDICT"; `build-draft.sh:374` instructs printing the PR URL as the final line. The verdict is model-authored, so the record is prompt-borne. Phase 4 is sized accordingly.

## Architect review — 2 Critical, 7 Warning; all addressed

- Phase 4's "0 bytes of prompt content" false → **fixed**: prompt-surface row added, plus a Phase 3 ruling on record authorship and a Phase 4 prompt↔field correspondence check.
- Migration enumeration bash-only; Python declares `COMPLETION_PATTERN = r"…"` with spaces → **fixed**: 21 declarations across two fleets, language-agnostic greps prescribed.
- Warnings fixed: second verdict call site (`build-minor.sh:281`, and `build.sh:277` not `:281`); transport decision now weighs Temporal activity-boundary/replay; record path must be anchored outside the worktree; cross-language schema ownership is now an explicit Phase 3 ruling; roadmap T-labels → E-labels; Phase 1 now *cites* fleet-reliability's liveness paper instead of re-deriving it; the CPI "runway" deferral is placed as a Phase 2 retrievability requirement.
- Recommendations taken: bash-fleet disposition raised as a blocking ruling before Phase 4 is sequenced; split-by-consumer named as a deliberate choice; costs added to Key Decisions; the measured 0.9% (4/443) turn-cap rate cited.

## Security review — 3 High, 4 Medium, 1 Low; all addressed

- **Channel staleness** — `build.sh` calls `review-pr` twice per run; a reused path lets a stale pass-1 record survive a pass-2 child that dies, making the "absent → human" arm unreachable exactly when needed → **fixed**: fresh path per invocation, run-identity binding, and a stale-record test that a first-invocation-only test would pass despite.
- **PR-URL regex deleted with the extraction** — it pins the host and guarantees a digits-only PR number flowing into `gh pr view/checks` → **fixed**: retained as a validator on the field; a mismatch routes to the human arm.
- **`permission_denials[]` closable as a routing no-op** — it is the only trace the sole in-run safety control leaves under `--dangerously-skip-permissions` → **fixed**: unbundled into safety observability with no no-op path, never auto-redispatch, and E1 now forces a denial so it is observed at least once.
- Medium fixed: per-field publish classification before the record reaches a PR comment (denial entries carry `tool_input.command`); record anchored outside the worktree (children run `git add -A`); an unknown-`schema_version` arm added to the fail-safe contract; transport ruled on trust boundary, not just availability; consumer-side completeness check made the requirement with atomic write as the additional measure.
- Low → surfaced as amendment candidate #3: `workflow-scripts.md § Test Fixture Placement` describes a `block-dangerous.sh` protection that does not exist (the hook exits 0 for non-`Bash` tools).

## Planner review — 3 Critical, 6 Warning, 2 Info; all addressed

- **Phase 5 planned as greenfield** — `review-pr.sh:355` already emits `converged`, `:221`/`:357` already mandate stable ids → **fixed**: reframed as a class-(iii)→class-(ii) replacement, incumbent added to Dependencies, `§Capability Parity` requirement added, key-naming ruling added.
- Same 0-bytes finding → fixed above.
- **`permission_denials[]`/`num_turns` closable on an E1 that measured neither** → **fixed**: E1's tuple now covers all of them, and each closes on its own evidence.
- Warnings fixed: the measurement that could prove Phase 5 decorative moved into Phase 1 as **E7**; Phase 5's hard dependency corrected to Phase 3 with Phase 4 soft (may run in parallel); E3's data source must be stated and only two Phase 3 requirements block on it; E6's Autonomous Operation clause deleted; E5 given an explicit adjudication procedure; Phase 4's single-child rendering slice moved into Phase 3.
- Info fixed: roadmap argument #2 rewritten (it cited an archive step that lives elsewhere); the problem-statement element-3 escalation added as amendment candidate #4.

## Quality-control review — 1 new Critical, 2 Warning, 1 Info; all addressed

- **A second, more consequential instance of the same miscitation** — the "Why five phases" rationale opened by citing the non-binding §2, with no local rescue (`sprint.md` has no phase-splitting language) → **fixed**: the rationale now states the test in the plan's own words and owns the reasoning.
- Archive-as-by-product misattributed to Phase 3 in two docs → **fixed** in both (and resolved structurally by moving the step there).
- Phase 5's silence on the shipped `converged` field → fixed above.
- `build.sh:281` off-by-four → fixed.
- **Meta-pattern accepted, not deflected:** QC found five instances of one habit — this plan demanded rigor of future work (mandatory verification blocks, "re-run the greps," "state the denominator") while not applying that apparatus to its own counts and citations. That is a fair reading of the first draft. Phase 1 and Phase 4 now carry the failure as an inline caution, and the plan's numbers were re-derived by grep rather than restated.

## Cross-reference consistency: pass

All six docs' relative links resolve; all quoted spans from `workflow-scripts.md` and the research pool verified character-accurate. Filenames conform to `^phase[0-9]+[a-z]?_[a-z0-9_-]+\.md$`. No `docs/standards/` file and no `sprint.md` line was modified (`git diff --stat` confirms). Zero checked checkboxes.

## Surfaced for the operator — not written

1. **`sprint.md`'s Memory Management Framework entry** still reads *"Phase doc: not yet written — writing it is the planning step,"* and now needs a link to `memory-management-framework/roadmap.md`. Human-in-the-loop; not edited.
2. **A blocking cross-component ruling:** whether the bash fleet is migrated or retired. Temporal Integration carries it as an open question while being gated on Phase 4. Phase 4 should not be sequenced until it is answered.
3. **A research-side correction candidate.** The synthesis (candidate 7, and §10) states the measurement argument *"has no working incumbent to beat."* `review-pr.sh:323,355` shows a `converged` flag already ships. The claim is narrowly defensible — nothing routes on that flag — but as written it reads stronger than the tree supports. The research pool is verified evidence and read-only to this run; flagging it rather than editing it.
4. **Five standards-amendment candidates** are listed in the roadmap, including one (`workflow-scripts.md § Test Fixture Placement`) that describes a safety control the shipped hook does not implement.


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Kept the roadmap+phase-docs shape after discovering the dispatch's authority for it was wrong. Alternatives: (a) rebuild as a single `<name>/<name>.md`, (b) stop and ask. Why: `sprint.md:28` independently mandates this shape for a component that outgrows one phase, and this one does on any reading — so the outcome was right and only the citation was wrong. Rebuilding would have discarded correct work over a footnote. **This is the decision most worth a second opinion**, because it means I proceeded on a premise the dispatch got wrong rather than halting.
- **[Low]** Moved the Kind-1 rendering/archive step out of Phase 4 and into Phase 3. Alternatives: leave it in Phase 4 (as drafted); split it into its own phase. Why: planner argued its failure mode (an operator loses prose) is unrelated to Phase 4's (an unenumerated call site), and it concerns only the child Phase 3 already opens. It grows Phase 3, which I had deliberately kept narrow — the trade is real and I took it because it also resolved the roadmap's Phase-2-ordering rationale, which had been citing this step in the wrong phase.
- **[Medium]** Chose five phases over four. Alternatives: four (mirroring the sprint milestones); three (folding measurement into design). Why: two experiments can *cancel* downstream work, and a cancel-check inside the phase it would cancel is a gate that never fires. QC then found a sixth instance of the same reasoning I had not applied — the measurement that could prove Phase 5 decorative sat inside Phase 5 — which is evidence the argument is right and I under-applied it. Fixed by adding E7 to Phase 1.
- **[Medium]** Ruled the stall/liveness axis out of this component and into Fleet Reliability. Alternatives: build it here (the research named this component as a candidate destination); leave the scope call open. Why: `sprint.md` already places the three-legged liveness predicate in Fleet Reliability, and that component's pool has already *measured* the progress signals. Leaving a raised scope question unruled is how work gets built twice. It has a sprint implication, so it is surfaced rather than written.
- **[Medium]** Did not edit the research synthesis despite finding one of its claims contradicted by the tree ("the measurement argument has no working incumbent to beat" — `review-pr.sh:355` ships a `converged` flag). Alternatives: correct it; treat the plan's narrower framing as sufficient. Why: the pool is critic-gated verified evidence and read-only to this run. The plan states the accurate version and the discrepancy is surfaced; correcting a verified paper from a planning run would bypass the process that made it trustworthy.
- **[Medium]** Accepted QC's meta-finding rather than arguing the individual instances. Alternatives: dispute the count (a couple were arguably inherited from the dispatch rather than authored). Why: two of the five *were* mine, the pattern is real, and the useful response is the inline caution now in Phase 1 and Phase 4 — not an apportionment argument.

## Deferred Work

Nothing was deferred to avoid doing it; these are items this run is not permitted to write.

- **Update `sprint.md`'s Memory Management Framework entry** (it still says the phase doc is not yet written) — Why not done: sprint file is human-in-the-loop only. Tracked at: **no tracker — surfaced in the PR body with no pointer.** Verified by: `git show origin/main:docs/development/sprint.md | sed -n '145p'` → `**Phase doc:** not yet written — writing it is the planning step.` — confirmed stale on the default branch, and confirmed nowhere else records it.
- **The bash-fleet disposition ruling** (retire, or keep as an edge fallback) — Why not done: joint with Temporal Integration; not this component's call alone. Tracked at: `docs/development/temporal-integration/temporal-integration.md:63`, where it is already an open question. Verified by: `git show origin/main:docs/development/temporal-integration/temporal-integration.md | grep -n "bash fleet after Stage A"` → line 63, `**What happens to the bash fleet after Stage A** — retired, or kept as an edge fallback needing no runtime.` — open on the default branch and covering exactly this item. Phase 4's Dependencies now names it as blocking that phase's sequencing.
- **Correcting the synthesis's "no working incumbent" claim** — Why not done: the research pool is verified evidence, read-only to this run. Tracked at: **no tracker.** Surfaced in the PR body as a research-side correction candidate; I did not open an issue because an autonomous dispatch may not file deferred work as an issue.

**Placed, not deferred** (recording where, since these are resolutions rather than carries): the CPI deferral *"a correction pass cannot machine-read the prior pass's runway"* — verified by `git show origin/main:docs/development/cpi-decisions.md | grep -n "cannot machine-read the prior pass's runway"` → line 974, present on the default branch with watch-criteria *"ship as part of the Memory Management phase doc"* — is now a retrievability requirement and a dedicated checklist section in `phase2_kind1_framework.md`. Its trigger had fired; this is the doc it named.

## Post-Run Reflection

**Friction**

- **The dispatch's binding premise was wrong, and checking it was not part of the instruction.** It stated §0/§1/§2 as binding and told me to treat §0 as settling the shape. `docs/standards/documentation/README.md:34` excludes that entire section from binding here. I followed the instruction and cited §0; two reviewers caught it independently. A dispatch that names a vendored standard as binding should also name its applicability note, because a vendored MIRROR in this repo has one by convention.
- **The dispatch asserted "it changes none of the 229KB of prompt content" as a fact to plan around.** It is false — `review-pr.sh:427` is literally a stage titled "PRINT THE VERDICT." I propagated it into a phase's sizing. Assertions in a dispatch's "live constraint" section read as verified context; this one wasn't, and I should have grepped it before building a phase on it.
- **The research pool numbers its experiments T1–T6 across two papers with two different meanings for `T3`.** I renamed to E-labels, then left the roadmap citing T-labels — an internally inconsistent artifact until architect caught it. Renaming a label set requires a sweep, not a rename.
- **A bash-shaped grep silently measures half this fleet.** `COMPLETION_PATTERN=` finds 11; `COMPLETION_PATTERN\s*=` finds 21, because the Python tree uses spaces. This is now a caution inside the phase docs, but it is a general hazard for any run enumerating this repo: there are two fleets and most greps only see one.

- `docs/file_structure.txt` is stale in ways this build did not touch and did not fix: it lists `loose_ends.md` (deleted in `aea01e5`), lists `github-actions-integration/` as a top-level component folder when it is a file under `autonomous-execution/`, and omits four component folders (`fleet-reliability/`, `managed-configuration/`, `memory-management-framework/` before this PR, `workflow-decomposition/`). Surgical-changes discipline says mention rather than fix; mentioning.
- `docs/standards/architecture/research/topics.md:86` points at `docs/development/phases/memory-management-framework/research/` — a `phases/` level the repo does not have. Now a live broken reference, since the real folder exists and is populated.
- Every other component uses `<name>/<name>.md`, so this is the first roadmap+phases component here and there was no local exemplar to match. If the shape is expected to recur, one of these six docs is now the exemplar and it would be worth saying so somewhere findable.

- **Stage 4a's four reviewers found things a self-check would not have**, and three of the five most valuable findings came from reviewers verifying the *plan's own factual claims* against the tree rather than reviewing its reasoning. The stage instruction frames the lenses as design questions; adding an explicit "verify every count, line citation and 'no X changes' claim the artifact makes" to all four lens briefs would catch this class earlier and more cheaply than a Critical in review.
- **The workflow prompt has no instruction to verify a dispatch's own asserted facts.** It has strong verification discipline for deferral pointers ("verification is by fetch, never by plausibility") and none for the premises in the task description. Both failure modes I hit came from that gap. The deferred-work rule's exact framing would transfer directly.
- The `plan-revision.sh` worktree left me on a detached HEAD, so `git push -u origin HEAD` failed and I had to name a branch myself. Minor, but every run that reaches Stage 6 hits it.

Issue #55, opened 2026-08-08
vendor-standards --check is the drift gate for four MIRROR standard sets and runs on the merge path nowhere — blocked on where CI gets the upstream


Pumapumapumas:
`scripts/helpers/vendor-standards.sh --check` is the drift detector for the four vendored MIRROR standard sets that `CLAUDE.md` says MUST NOT be edited locally (`docs/standards/temporal/`, `documentation/`, `research/`, `testing/`). It is invoked by convention only. **A local edit to a vendored standard merges green today.**

It belongs on the merge path with the other controls added in the tests.yml expansion, and it was deliberately left out of that PR. This issue carries the blocking question.

## Why it could not be gated there

The script resolves its upstream by looking for a clone on disk:

```
for candidate in "$HOME/Repos/mdc-master-planning" /opt/skyy-net/mdc-master-planning; do
    if [[ -d "$candidate/standards/development/temporal" ]]; then SRC="$candidate"; break; fi
done
if [[ -z "$SRC" ]]; then ... exit 1; fi
```

On a GitHub runner neither path exists, so `--check` exits 1 on "not found". Measured by simulating the runner's environment:

```
$ HOME=/tmp/nonexistent-home-xyz scripts/helpers/vendor-standards.sh --check
Error: missing upstream file: .../research/research_standard.md
EXIT=1
```

`main` is branch-protected on the `suite` job, so a step that exits 1 on arrival blocks every merge in the repo.

## Why the obvious workarounds were rejected

- **Skip the step when the clone is absent** (`if [ -d ... ]`) — this is the failure mode, not the fix. It emits a green check that verified nothing, and a green check is read by reviewers as evidence. Worse than the gap it hides.
- **`|| true`** — same class, more obviously.
- **Add a fetch/checkout of the upstream to CI** — this is the real answer, but it requires a ruling nobody has made: which repo and ref CI compares against, and how a private repo is authenticated from this workflow. That ruling is the content of this issue.

## What has to be decided

1. **Where does the upstream come from on a runner?** `actions/checkout` of `helloskyy-io/MDC-Master-Planning` with a PAT/app token is the conventional answer; it adds a secret and a second private-repo dependency to a gate that currently has neither.
[...]

Issue #53, opened 2026-08-07
The CI gate runs the test suite but nothing lints the bash fleet — shellcheck, bash -n, JSON/YAML validity all unchecked on the merge path


Pumapumapumas:
Drained from `docs/development/loose_ends.md` (2026-05-28 architect review), which is being deleted per Documentation Standard § Deferred Work.

## Why this one has a done-state today

The original entry was *"CI on this repo. No `.github/workflows/`."* **Half of it shipped today** — `.github/workflows/tests.yml` runs the master runner plus a `ruff --select F821` sweep, and `main` is branch-protected requiring it. That half is closed (#30).

**The other half is still unowned:** nothing runs `shellcheck`, `bash -n`, or JSON/YAML validation.

## What is exposed

There are nine bash workflow scripts and several hook scripts. A syntax error in one surfaces when it is dispatched — during a run, in a worktree, after a worktree has been cut. `config.yaml` and `settings.json` are both parsed at runtime by things that fail late and unhelpfully.

This is the same class the executability sweep already covers on the Python side: **a name used but never defined passes every unit test and crashes at runtime.** `ruff --select F821` closed it for Python. Bash and the config files have no equivalent.

## Why an expansion rather than a sibling

Per the placement rule, this is **an expansion of the existing gate**, not a new one. It is steps added to `tests.yml`, running unfiltered like everything else there — the Tier Enforcement clause binds the tier gate, and these are auxiliary checks in the same job.

## Scope

- `shellcheck` over `scripts/**/*.sh` and `config/hooks/*.sh`
- `bash -n` as a syntax floor for anything shellcheck cannot parse
- `config.yaml` and `settings.json` parsed and validated
- decide per-check whether a finding blocks or warns — shellcheck on a mature bash fleet will surface real style noise on the first run, and a gate that everyone learns to ignore is worse than none

Pumapumapumas:
Closed by PR #56 (`00a368e`) — but with a **scope different from what this issue proposed**, recorded here rather than in a PR body that dies at merge.

## What shipped

Five steps in the `suite` job, all demonstrated green before merge:

- **`lint-prompts.sh` gated** — it was written after two fleet-blocking exit-127 outages in one day and had been invoked by no gate since. Highest-value item here, and **this issue never mentioned it.**
- **`bash -n`** across the tree, described honestly in its step comment as a syntax floor **with a known blind spot** (`3975ed6`: a file can be valid bash and mean something catastrophically different).
[...]

Issue #35, opened 2026-08-06
Half the vendored Testing Standard does not apply here, and every reviewer re-judges which half


Pumapumapumas:
## The consequence

Roughly half of `docs/standards/testing/testing_standard.md` is Django / Kubernetes / ArgoCD / Vault / skyy-command material with no bearing on this repo, and **nothing says which half binds here**. Every reviewer who touches the testing tier re-derives that judgement clause by clause, independently, and they will not all reach the same answer.

This is not hypothetical. On PR #31 alone:

- Three separate review dispatches each had to judge applicability from scratch.
- One live disagreement reached the PR body: whether `testing/suites/bash.sh` was a required deliverable. The build run resolved it by quoting §Tier 2 ("Only create suite runners for frameworks actually in use") against a task brief that had named the file — a correct call, but one that cost an argument in the PR body because no applicability layer existed to settle it in advance.

A binding standard whose scope is re-litigated per reviewer produces inconsistent enforcement, and inconsistent enforcement of a *binding* document is worse than a narrower document consistently applied.

## Evidence

Verified against `origin/main` @ `1f995aa79ff88c823fdc629a4be4ca47cee3a1a3`:

```
$ ls docs/standards/testing/
testing_standard.md

$ ls docs/standards/temporal/
claude-dot-files-addendum.md
README.md
stateful_patterns.md
temporal_standard.md
worker_deployment_standard.md
```

`docs/standards/temporal/` already has exactly the pattern that is missing here — a `README.md` stating what binds today versus what applies only once workers exist, plus `claude-dot-files-addendum.md` for local additions. Root `CLAUDE.md` points readers at that README by name. The testing standard gets no such treatment.

Clauses in `testing_standard.md` that plainly do not apply to this repo, as examples of the surface needing a ruling: the ArgoCD app-of-apps e2e ladder (§ lines 253–279), Vault / 1Password credential handling in tests (§ line 285 onward), the Django app-structure references (§ line 134), and the multi-repo framework table (§ lines 495–503).

## Why this could not be written by the run that found it

[...]

Issue #30, opened 2026-08-06
Testing Standard § Tier Enforcement: no automated gate runs the unit tier on the merge path


Pumapumapumas:
## The gap

`docs/standards/testing/testing_standard.md` § *Tier Enforcement — a tier that no gate runs is unenforced (binding)* requires:

> **Every tier the master runner can discover MUST have a corresponding automated gate.** ... "The engineer runs it before pushing" is a convention, not a gate — it is unobservable, unevenly applied, and silently lapses. The property that matters is that a change **cannot merge** without the tier having run.

As of the three-tier restructure, this repo has:

- Tier 1 `testing/run-all.sh`
- Tier 2 `testing/suites/python.sh`
- Tier 3 `scripts/workflows/temporal/tests/unit/` — 140 passing tests

and **no automated gate that runs any of them**. There is no `.github/workflows/` directory in the repo, and no hook in `config/hooks/` invokes the runners. The only references to `run-all.sh` are prose instructions inside workflow-script prompts (`scripts/workflows/build-phase.sh`, `scripts/workflows/children/build-refine.sh`), which tell a dispatched engineer to run it — a convention, not a gate.

## Consequence

A regression in the temporal test tree can merge to `main` with nothing blocking it. The tier is correctly placed, fully discoverable, and composed of tests that demonstrably fail when the property they assert is violated (each was mutation-checked) — and it still catches nothing on the merge path, because nothing invokes it there. This is the failure mode the standard singles out as looking healthiest of all: the files exist, the count grows, `run-all.sh` finds them, and a human running them locally sees green.

## Remedy

Add a GitHub Actions workflow that runs the unit tier on every PR touching `scripts/workflows/temporal/**` or `testing/**`:

```yaml
name: tests
on:
  pull_request:
    paths: ['scripts/workflows/temporal/**', 'testing/**', 'conftest.py', 'pytest.ini']
jobs:
  unit:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with: { python-version: '3.13' }
      - run: pip install pytest
      - run: ./testing/run-all.sh unit
```

`run-all.sh` already exits non-zero on failure and already fails loudly when zero suites ran, so it is safe to use directly as the gate command — a green exit means tests actually executed, not that discovery came up empty.

## Why this was not done in the restructure PR

[...]

PR #56, opened 2026-08-08
build-draft: put the repo's existing controls on the merge path — prompt lint, settings.json, bash -n, shellcheck

Pumapumapumas:
Expands the `suite` job in `.github/workflows/tests.yml` with four checks. Addresses #53, **with a scope that differs from what that issue proposes** — see "Divergence from #53" below.

## Update — review-pr ruling applied (this pass)

`review-pr`'s pass-1 disposition (see PR comments) held three items. Rebased
onto `origin/main` first — it had gained the re-vendored Testing Standard's
**"A control that no gate runs is not a control (binding)"** clause
(`9172cba`), which two of these three items directly answer to.

1. **`testing-readme-describes-the-pre-pr-gate` (HOLD — redispatch).**
   `docs/standards/testing/README.md`'s Tier Enforcement row and § The gate
   both predated the new clause. Row restated against the six-check gate; §
   The gate now names the consequence the clause requires for an advisory
   control: **a local edit to a vendored MIRROR standard merges green
   today.**

2. **`settings-validator-cannot-be-verified-or-run` (HOLD — needs-assistance,
   RULED).** The operator's ruling overrides the original DoD's placement
   instruction, deliberately: moved the validator from an inline CI heredoc
   to `scripts/helpers/check-settings.sh` (a step in the `suite` job,
   matching `lint-prompts.sh`'s shape — the DoD is satisfied literally) with
   the logic in an importable `scripts/helpers/check_settings.py`. The
   eleven NC-A..K negative controls that proved the parse logic correct —
   previously prose in this PR body only — are now a committed, re-runnable
   pytest at `scripts/helpers/tests/unit/test_check_settings.py`, gated by
   `run-all.sh`. This was the one control in the repo whose mutation
   evidence could not be committed or re-run; both gaps are closed. The
   two paragraphs under "Surfaced" below describing this as undone are
   stale — left in place for the historical record, corrected here.

3. **`gate-job-supply-chain-unpinned` (HOLD — needs-assistance, RULED).**
   SHA-pinned `actions/checkout`, `actions/setup-python` and
   `actions/upload-artifact`, each with the resolved tag in a trailing
   comment. "Pre-existing" and upgrade-friction were explicitly rejected as
   dispositions — friction is priced by the operator, not waived. This
   branch's own CI was verified green with the pins in place before this PR
   was pushed (see the run linked from this PR's checks).

`vendor-standards-control-ungated` stays at **#55**, untouched, per the
disposition's ruling. The two escalations (`documentation/README.md:36`,
`workflow-scripts.md`'s `lib/` references) are live on `main` and are the
operator's to handle separately — not touched here.

## The reframe

#53 asks for shellcheck, `bash -n` and config validation, reasoning that the bash fleet has no equivalent to the `ruff --select F821` sweep that covers Python. Running that question down changed it.

**The gate's contract was never stated, and that is the actual defect.** `tests.yml` describes itself as the tier gate the Testing Standard requires, and its own header makes the argument that decides this:

> "The engineer runs it before pushing" is a convention, not a gate — it is unobservable, unevenly applied, and silently lapses.

That argument was applied to the test runner and to nothing else. **Two controls already exist in this repo, were each built in response to a real failure, and are invoked by convention only.** The question is not "which linters should we adopt" — it is **"which controls do we already have that the merge path cannot see."** The rest is derived rather than argued.

The clearest evidence is that `docs/standards/workflow-scripts.md:700` already says this, in these words:

> **Prompt edits MUST pass `lint-prompts.sh` AND `bash -n`**

A MUST enforced by memory. Referenced in three documents, gated nowhere.

## What was added

| # | Step | Class it catches | Status |
|---|---|---|---|
| 1 | Prompt-construction lint | "prompt strings are code" — the exit-127 outage class | **shipped** |
| 2 | Vendored-standards drift | local edits to MIRROR standards | **blocked — #55** |
| 3 | Validate `settings.json` | broken permission/hook config | **shipped** |
| 4 | Bash syntax floor | unparseable bash | **shipped** |
| 5 | shellcheck | outright-wrong shell constructs | **shipped at `--severity=error`** |
| 6 | `config.yaml` validation | — | **rejected, see below** |

### 1 · Prompt-construction lint — the highest-value item, and one #53 never mentions

`scripts/helpers/lint-prompts.sh` was written after **two fleet outages in one day** (`3975ed6`). Its own header records them: an unescaped backtick made `run_in_background: false` execute as a command (exit 127, `research.sh` died before Stage 1, caught by another team on a live dispatch), and unescaped double quotes closed a `PROMPT=` mid-string so the remaining prose parsed as commands (exit 127 again).

**`bash -n` passed both.** The first is valid syntax; in the second the stray quotes balanced, so the file was valid bash that simply meant something else.

It catches them by *constructing* each prompt block under `env -i` with only `cat` on `PATH` and failing if bash does anything but assign a string. That is deliberate: a pattern list built for the first vector certified the file that carried the second.

**Honest limit, recorded in the step comment:** its Pass 3 (every `MODEL_KEY` resolves in `config.yaml`) is behind `command -v yq` and self-skips when `yq` is absent — while still printing the clean line, which claims more than it checked. That property is covered independently by `test_model_keys_resolve.py` under the master runner. Passes 1 and 2, the outage class, need nothing beyond bash.

### 2 · Vendored-standards drift — NOT gated, and this is a deviation

**This was the second-highest item in scope and it could not be delivered.** `vendor-standards.sh --check` resolves its upstream by looking for a clone on disk and exits 1 when it finds none:

```
$ HOME=/tmp/nonexistent-home-xyz scripts/helpers/vendor-standards.sh --check
Error: missing upstream file: .../research/research_standard.md
EXIT=1
```

On a GitHub runner neither candidate path exists, so that is **every** run. `main` is branch-protected on the `suite` job, so wiring it as specified would have gone red on arrival and blocked every merge in the repo.

Both obvious workarounds were rejected rather than taken:

- **Skipping the step when the clone is absent** emits a green check that verified nothing. That is the exact failure this PR's own reasoning condemns — a green check is read by reviewers as evidence — and it is worse than the gap it hides.
- **Adding a fetch to CI** is the real answer, but requires deciding which repo/ref CI compares against and how a private repo authenticates. That ruling was explicitly out of scope here.

Filed as **#55** with the three decisions it needs, and recorded as a comment block in `tests.yml` itself so the omission reads as a decision rather than an oversight. The related hole noted in scope — that a green `--check` is only as good as the freshness of the operator's clone, and nobody fetches first — is captured there too.

### 3 · Validate `settings.json`

Verified: nothing anywhere parses it as a check. `install.sh` only symlinks it; the only other repo hit is a comment. It configures the permission model and the hooks, so a malformed one can mean **the hooks never load at all** — silently removing the only control operating during an autonomous run.

Parse validity is the floor. The step goes further, as scope asked: `install.sh` symlinks `~/.claude/hooks` → `config/hooks`, so every `$HOME/.claude/hooks/x.sh` command must resolve to an **executable** `config/hooks/x.sh`. It also fails when *zero* references were found — same principle as the master runner's "nothing ran at all" exit, since a check that silently verified nothing after a restructure looks healthiest of all.

### 4 · Bash syntax floor

All 37 `.sh` files pass today; a zero-noise ratchet. **Described in the comment as a syntax floor with a named blind spot**, citing `3975ed6`: a file can be valid bash and still mean something catastrophically different. Overselling this step would be worse than no comment, because the next reader takes it for coverage of the prompt class — which item 1, not this, provides.

### 5 · shellcheck — measured first, then adopted

shellcheck was not installed on the dev machine, so **nobody had ever seen the finding count.** Obtained via `pip install "shellcheck-py>=0.10,<0.11"` (shellcheck 0.10.0) and run over all 37 files:

| severity | bare | with `-s bash` |
|---|---|---|
| **error** | 2 | **0** |
| warning | 40 | 38 |
| info | 84 | 82 |
| style | 86 | 84 |

**Adopted: `--severity=error -s bash`. Zero findings, no baseline file, no excluded codes, no `|| true`.** #53's own warning is the binding constraint — *"a gate that everyone learns to ignore is worse than none"* — and `|| true` is that failure in its purest form. No cleanup campaign was opened.

Warning level is 38, of which **32 are `SC2034` unused-variable**. Ratcheting to warning later is a matter of fixing those 32, not suppressing them.

**Why `-s bash`, stated plainly because it is the only thing between this step and green.** Both bare-mode errors are `SC2148` ("shell is unknown") on `activities/run-claude.sh` and `common/shared-prompts.sh` — sourced libraries that intentionally carry no shebang. The conventional fix is a `# shellcheck shell=bash` directive in each, i.e. editing a script to satisfy a linter, which scope forbids. Declaring the dialect CI-side touches no file and changes nothing for the other 35, which all already declare bash via `#!/usr/bin/env bash`. **SC2148 on those two is reported here as a finding rather than silenced by pretending it was never raised** — fixing it is a separate ruling.

**Nothing was baselined or excluded.** The `-s bash` flag is a dialect declaration, not a suppression, and it is the only flag beyond the severity threshold.

**One measurement worth keeping:** shellcheck flags `SC1078`/`SC2140` on `research.sh` and `research-refresh.sh` — the exact two files from the `3975ed6` outage — and both are **false positives** on the already-escaped forms. shellcheck is not a substitute for item 1; it is a live demonstration of why the execution check exists.

### 6 · `config.yaml` validation — REJECTED

`scripts/workflows/temporal/tests/unit/test_model_keys_resolve.py:30` sets `CONFIG = REPO_ROOT / "config.yaml"` and loads it at test time. `run-all.sh` is already gated by this workflow. **Malformed YAML therefore already fails the suite on the merge path**, and a dedicated step would add a second red signal for a condition already covered. Recorded here so a later cycle does not re-propose it.

## Divergence from #53

| #53 says | This PR | Why |
|---|---|---|
| shellcheck over `scripts/**` and `config/hooks/*.sh` | shellcheck over **every** `.sh` in the tree | Same reasoning as the no-`paths:` rule — a scoped sweep skips `install.sh`, `testing/`, and the V2 shims for no gain |
| `bash -n` as a floor | shipped as specified | — |
| `config.yaml` **and** `settings.json` validated | `settings.json` only | `config.yaml` already covered — item 6 |
| "decide per-check whether a finding blocks or warns" | **everything blocks** | Warning-only steps are the ignorable gate #53 itself warns against. The lever used instead was severity: gate at a level that is clean, rather than gate everything and downgrade |
| — | **prompt-construction lint** | Not in #53. Highest-value item on the list — the only class measured twice in production here |
| — | **vendored-standards drift** | Not in #53. Blocked; #55 |

The two items #53 never mentioned are the highest-value part of the work, and one of the two it did name was already covered.

## Constraints honoured

- **No `paths:` filter, no `changed-files` filter.** The existing header's reasoning binds, and a filtered gate is worse than a missing one because it emits a green check a reviewer reads as evidence.
- **Auxiliary steps in the existing `suite` job**, not a new workflow — per #53's own placement reasoning.
- **Pinned major range**, matching `pytest`/`pyyaml`/`ruff` on the same line. `shellcheck-py` is pinned *tighter* (`<0.11`, not `<1`) for a linter-specific reason recorded in the file: shellcheck is 0.x, so its **minor** releases are the breaking ones — each adds checks, which on a gate means new failures with no commit to this repo. That is exactly the "changes behaviour underneath the gate" case the existing pins exist to prevent.
- **No script was modified to satisfy a linter.** `lint-prompts.sh`, `vendor-standards.sh` and all 37 shell scripts are untouched.
- **Every added step demonstrated green locally** before the PR opened — output below.

### Every step run as CI runs it

Each `run:` block was extracted from the **parsed** YAML and executed under `bash -e` from the repo root — GitHub's `bash -e {0}`, not a hand-retyped approximation. This also proved the heredoc de-indents correctly through the YAML block scalar.

```
======== STEP: Prompt-construction lint ========
✓ prompt lint clean — every prompt block constructs, every MODEL_KEY resolves
EXIT=0
======== STEP: Validate settings.json ========
config/settings.json: valid JSON; 2 hook reference(s) resolve to executable files in config/hooks/
EXIT=0
======== STEP: Bash syntax floor ========
bash -n: checked 37 files
EXIT=0
======== STEP: Shellcheck ========
EXIT=0
```

Dependency line resolves: `pip install --dry-run "pytest>=7,<9" "pyyaml>=6,<7" "ruff>=0.6,<1" "shellcheck-py>=0.10,<0.11"` → `Would install pytest-8.4.2`, shellcheck-py 0.10.0.1 already satisfied.

### Negative controls — every step demonstrated able to go red

A green check that cannot fail manufactures confidence. All seven ran in a `/tmp` sandbox copy; the worktree was confirmed unmodified afterwards (`git status` clean, lint still green).

```
### NC1: settings.json — malformed JSON
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 305 column 1
EXIT=1

### NC2: settings.json — hook path points at a missing file
config/settings.json: PreToolUse: $HOME/.claude/hooks/does-not-exist.sh -> config/hooks/does-not-exist.sh does not exist
EXIT=1

### NC3: settings.json — hook exists but is not executable
config/settings.json: PreToolUse: $HOME/.claude/hooks/block-dangerous.sh -> config/hooks/block-dangerous.sh is not executable
EXIT=1

### NC4: settings.json — hooks restructured away (verified-nothing guard)
config/settings.json: no hook references $HOME/.claude/hooks/ — this check verified nothing
EXIT=1

### NC5: bash -n — unparseable file introduced
SYNTAX ERROR: ./scripts/helpers/broken.sh
EXIT=1

### NC6: shellcheck — error-severity defect that bash -n PASSES
  (bash -n on it: EXIT=0 — passes, as expected)
exit 300
     ^-^ SC2242 (error): Can only exit with status 0-255.
EXIT=123

### NC7: lint-prompts — reintroduce the 3975ed6 unescaped backtick
  (bash -n on it: EXIT=0 — passes, which is the whole point)
✗ research.sh — unescaped backtick(s):
      266: ... (\`run_in_background: false\`) ...
✗ research.sh — prompt block starting at line 217 does NOT construct:
      /usr/bin/bash: line 98: run_in_background:: command not found
EXIT=1
```

**NC7 reproduces the original outage exactly** — `run_in_background:: command not found` — and confirms in the same run that `bash -n` passes the file the lint rejects. That is the empirical case for item 1 over item 4.

### Suite

```
 python/unit            PASS
 python/integration     SKIP (no tests/integration)
 python/e2e             SKIP (no tests/e2e)
RESULT: PASSED
```

337 passed in 3.68s.

## Review findings (build-refine pass)

A separate review run audited this PR. It found **six defects, all in the same class the PR argues against** — a check reporting green having verified less than it claims. All six are FIXED here; the disposition table below covers every finding raised.

### Fixed

1. **The `MODEL_KEY` backstop claim was false.** The step comment said Pass 3's coverage "is covered independently by `test_model_keys_resolve.py`". It is not: that pytest sweeps `modules/assistant` for the Python spelling `MODEL_KEY = "..."` and says so in its own docstring; `lint-prompts.sh` Pass 3 covers the 12 bash workflows. **The two sets are disjoint.** Since Pass 3 is `command -v yq`-gated and self-skips while still printing `every MODEL_KEY resolves`, a runner image without `yq` would drop bash-fleet coverage to zero behind a green tick — with a comment telling the next reviewer not to worry. Comment corrected, and the step now asserts `yq` is present so the skip cannot happen silently on the merge path.

2. **The hook-path check did substring surgery on a shell command line**, and was wrong in both directions. Measured, all 11 cases:

| case | old | now |
|---|---|---|
| `"$HOME/.claude/hooks/x.sh"` (quoted) | trailing quote in path → **FALSE RED**, blocks every merge | passes |
| `~/.claude/hooks/x.sh` | prefix never matched → **FALSE GREEN**, unverified | checked |
| `${HOME}/...` | **FALSE GREEN** | checked |
| `a.sh && b.sh` where b is missing | only `a.sh` checked → **FALSE GREEN** | **red** |
| one canonical + one source-repo path | partial, `checked=1`, green | **red** |
| missing file / not executable / malformed JSON / zero hooks | red | red |

   Now tokenises with `shlex` after normalising every `$HOME` spelling, checks **every** `hooks/*.sh` token, and requires `checked == total` so a partially-unreadable config cannot report clean.

3. **Neither sweep excluded `./.claude/*`**, where this repo's own worktrees live. Measured from the main clone: **148 files swept instead of 37**. An engineer running these lines locally could go red on a sibling branch's WIP for reasons CI never reproduces — and a local run that cannot reproduce CI is one people stop trusting. `testing/suites/python.sh` already carries this exclusion with a comment explaining exactly why; the sweeps now match it.

4. **`bash -n` had no zero-file guard.** A find expression that stopped matching printed `checked 0 files` and exited 0. The `settings.json` step in the same PR guards this explicitly and the file cites the "nothing ran at all" principle three times — this step was not exempt from it. (The shellcheck step turned out to be safe only by accident: `xargs` runs `shellcheck` once with no arguments, which exits non-zero. Noted in the comment rather than left as luck.)

5. **Two docs asserted the opposite of what shipped.** `CLAUDE.md` said *"No tier is gated on the merge path yet — issue #30"* — false, and **#30 is closed**. That is the first file every session in this repo reads, so a false statement there is a false premise loaded into every subsequent session. And `docs/standards/testing/README.md` § The gate described the gate as `run-all.sh` plus `ruff` when it now runs six checks — that file exists *specifically* to stop reviewers re-deriving what the gate covers. Both corrected; the README now carries a CATCHES / DOES NOT CATCH table for all six.

6. **Hardening on a branch-protected job:** `permissions: contents: read` (least privilege), `timeout-minutes: 10` (a hung step held merges for the 360-minute default), and `if: '!cancelled()'` on every check so **one push reports every failing check** rather than only the first. The checks share no data or ordering dependency, so there was nothing to protect by serialising their failures.

Also: recorded the honest cost of `-s bash` — it **overrides** a shebang rather than supplying a missing one, so a future `#!/bin/sh` file would be linted as bash and its POSIX-dialect checks would go quiet. Costs nothing at `--severity=error`, but it is a real hole to close if this is ever ratcheted to warning. And the shellcheck census is now stamped with the version it was measured on (0.10.0), since a pin bump is exactly when those numbers stop being true.

### Rejected, with reasoning

- **"The shellcheck census may have been taken on a different version than the pin."** Reproduced the census independently on shellcheck 0.10.0 — `error 2 bare / 0 with -s bash, warning 40/38, info 84/82, style 86/84`, and 32 × SC2034 — which matches the pin's resolved 0.10.0.1 minor exactly. The green CI run on this branch is the stronger evidence: the pinned version already ran the step on a clean runner.
- **"Steps may be host-coupled via git index mode bits."** `git ls-files -s` reports `100755` for all four executables involved. Already falsified empirically by the green CI run.
- **"`documentation_standard.md:485` forbids an autonomous dispatch filing its own deferrals, so #55/#57 are violations."** The clause exists verbatim, but `docs/standards/documentation/README.md:36` lists **"Deferred Work — GitHub Issues"** under *"Ignore, do not adapt"* for this repo. The section does not bind here. (That carve-out is scoped more broadly than its stated rationale — surfaced below.)

### Surfaced — real, no verified home, for the operator to dispose of

- ~~The `settings.json` validator should arguably be a pytest, not inline YAML.~~ **RULED and DONE — see "Update" section above.** Moved to `scripts/helpers/check-settings.sh` with the NC-A..K cases committed as a pytest.
- ~~Three actions run on mutable tags~~ **RULED and DONE — see "Update" section above.** All three SHA-pinned, tag recorded in a trailing comment, verified green on this branch's own CI before push.
- **`workflow-scripts.md` shebang carve-out names a `lib/` directory that no longer exists.** The two shebang-less sourced files the `-s bash` flag exists for are `activities/run-claude.sh` and `common/shared-prompts.sh`. The reasoning reaches the right answer but rests on a path that moved. Standards are human-in-the-loop, so surfaced not edited.
- **`docs/standards/documentation/README.md:36`'s carve-out is broader than its rationale.** It excludes the whole "Deferred Work — GitHub Issues" section because *"the routing described is platform-specific"*, but that section also contains filing authority and checkbox-before-issue, which have nothing to do with routing. A reviewer spent cycles deriving whether §485 binds — the exact cost these applicability files exist to eliminate.
- **Head commit `a633a96` is subject `wip: ...`**, which is not a conventional-commit type and would land in `main`'s history. Squash-merge with a `ci:` subject resolves it.

### Reported, not fixed (unchanged from the draft)

- **`SC2148` on `activities/run-claude.sh` and `common/shared-prompts.sh`** — sourced libraries with no shebang. Handled CI-side with `-s bash`, which suppresses nothing a shebang would not have. Adding the directive to the two files is a separate ruling.
- **`lint-prompts.sh` Pass 3 prints a clean line for a property it may not have checked** — #57. The `yq` assertion added here means it cannot happen *on the merge path*; the local-run case is what #57 covers.
- **32 × `SC2034` at warning level** — not fixed, not suppressed. They are why warning is not the adopted level.

## Verification

**Every step re-run from the parsed YAML after the fixes** (extracted from the workflow with a YAML loader and executed under `bash -e`, not hand-retyped):

```
======== STEP: Run the master runner ========      337 passed in 3.73s   RESULT: PASSED   EXIT=0
======== STEP: Executability sweep ========        All checks passed!                     EXIT=0
======== STEP: Prompt-construction lint ========   ✓ prompt lint clean                    EXIT=0
======== STEP: Validate settings.json ========     valid JSON; all 2 hook command(s) resolve
                                                                                          EXIT=0
======== STEP: Bash syntax floor ========          bash -n: checked 37 files              EXIT=0
======== STEP: Shellcheck ========                                                        EXIT=0
```

**Negative controls — every rewritten guard demonstrated able to go red.** 11 cases against the validator, plus the two new guards:

```
EXIT=0  NC-A both canonical                        (want 0)
EXIT=0  NC-B quoted        — was FALSE RED         (want 0)
EXIT=0  NC-C tilde         — was FALSE GREEN       (want 0, now actually checked)
EXIT=0  NC-D ${HOME} braces                        (want 0)
EXIT=0  NC-E with args                             (want 0)
EXIT=1  NC-F chained &&, 2nd missing               (want 1)
EXIT=1  NC-G partial / source-repo path            (want 1)
EXIT=1  NC-H missing file                          (want 1)
EXIT=1  NC-I not executable                        (want 1)
EXIT=1  NC-J zero hooks (verified-nothing guard)   (want 1)
EXIT=1  NC-K malformed JSON                        (want 1)
EXIT=1  NC-L bash -n zero-file guard  — was EXIT=0 before this pass
EXIT=1  NC-M yq assertion with yq off PATH
```

The worktree-exclusion defect, measured from the main clone:

```
without ./.claude prune: 148 files
with    ./.claude prune:  37 files
```

The draft run's own negative controls (NC1-NC7, including NC7 reproducing the `3975ed6` outage exactly) remain valid and are unchanged by this pass.

## Files changed

- `.github/workflows/tests.yml` — 4 steps added; hook validator rewritten to tokenise; `./.claude/*` pruned from both sweeps; zero-file guard on `bash -n`; `yq` assertion; `permissions`, `timeout-minutes`, `if: '!cancelled()'`; `shellcheck-py` on the pinned install line; comment block recording item 2's absence; **(this pass)** settings.json step now calls `check-settings.sh` instead of an inline heredoc; `checkout`/`setup-python`/`upload-artifact` SHA-pinned with trailing tag comments; the shellcheck step's adoption-time census corrected to what re-measuring the current tree actually shows (info 44/style 2, not 82/84 — noticed while pinning the actions two steps below it)
- `CLAUDE.md` — corrected a false claim that nothing is gated on the merge path
- `docs/standards/testing/README.md` — § The gate now describes all six checks with their boundaries; **(this pass)** Tier Enforcement row restated for the six-check gate, and the vendor-standards.sh advisory consequence named per the new control-enforcement clause
- `docs/file_structure.txt` — `.github/` added to the map; corrected a stale "No README yet — see issue #35" (#35 is closed and the README exists); **(this pass)** the four new files under `scripts/helpers/`
- **(this pass, new)** `scripts/helpers/check-settings.sh` — the settings.json validator, relocated from an inline CI heredoc; locally runnable, matches `lint-prompts.sh`'s shape
- **(this pass, new)** `scripts/helpers/check_settings.py` — the importable validation logic `check-settings.sh` runs and the pytest suite imports directly
- **(this pass, new)** `scripts/helpers/tests/conftest.py` + `scripts/helpers/tests/unit/test_check_settings.py` — the eleven NC-A..K negative controls, committed and gated by `run-all.sh`


Pumapumapumas:
## Decision Log

- **[Low]** Edited `docs/file_structure.txt` to add `.github/`, which is adjacent to the task rather than in it. Alternatives: surface the gap in the PR body and leave the map alone (the surgical-changes default). Why: the map is absent an entire top-level directory while `CLAUDE.md` calls it authoritative, and this PR makes that directory materially more significant — but it is a pre-existing gap I did not cause, so the call could reasonably go the other way.

- **[Low]** Put the `settings.json` validator inline in the workflow rather than in `scripts/helpers/`. Alternatives: a helper script alongside `lint-prompts.sh` and `vendor-standards.sh`, which is the local precedent for "a check invoked as a gate"; or a pytest test alongside `test_model_keys_resolve.py`, which is the local precedent for "validate a repo config artifact". Why: the definition of done required item 3 to be a *step in the `suite` job*, and a 20-line helper nobody invokes locally is over-engineering for a JSON parse. **The cost is real and worth naming: unlike items 1, 2 and 4, this control is not runnable outside CI** — which cuts against this PR's own thesis that controls should be available locally *and* on the merge path.

- **[Medium]** Passed `-s bash` to shellcheck instead of adding `# shellcheck shell=bash` to the two sourced libraries. Alternatives: edit the two files (forbidden by scope — no script edits to satisfy a linter); gate at `--severity=error` without the flag (red on arrival, blocks every merge on a protected branch); exclude the two files from the sweep. Why: the flag declares a dialect rather than suppressing a defect, touches no file, and is a no-op for the 35 files that already declare bash via shebang. Excluding the files would have been a genuine coverage hole; this is not. Reported SC2148 as a finding rather than letting the flag bury it.

- **[Medium]** Gated shellcheck at `--severity=error` rather than `warning`. Alternatives: adopt warning with a baseline file or excluded codes; adopt warning with `|| true`. Why: measured first — error is 0 findings, warning is 38, of which 32 are `SC2034` unused-variable. #53's own constraint ("a gate that everyone learns to ignore is worse than none") makes the noisy option the wrong one, and `|| true` is that failure in its purest form. The ratchet path is fixing the 32, not suppressing them.

- **[High]** Did not ship item 2 (`vendor-standards.sh --check`), despite it being the second-highest item in scope. Alternatives: gate it behind `if [ -d <clone> ]` so it silently no-ops on runners; wrap in `|| true`; add an `actions/checkout` of the private upstream. Why: measured it exits 1 with no clone present, which is every GitHub runner, and `main` is branch-protected on this job — shipping it as specified blocks every merge in the repo. A conditional skip produces a green check that verified nothing, which is precisely what the rest of this PR argues against. Adding a fetch needs a ruling on repo/ref/auth that scope explicitly reserved. Filed #55 rather than shipping a laundered version.

- **[High]** Ran shellcheck over **every** `.sh` in the tree, where #53 scopes it to `scripts/**` and `config/hooks/*.sh`. Alternatives: the narrower scope as written. Why: same reasoning as the no-`paths:` rule the file already documents — the narrow scope skips `install.sh`, the `testing/` tier and the V2 shims for no measured gain, since error-level is 0 across all 37 anyway. Weighting toward the permanent set (activities/, common/, hooks, helpers, testing, V2 shims — ~21 files) was the stated intent, and full-tree coverage satisfies it without an exclusion list to maintain.

## Deferred Work

- **Gate `vendor-standards.sh --check` on the merge path** — Why deferred: exits 1 on any runner without an `mdc-master-planning` clone, and `main` is branch-protected on this job, so it would block every merge. Making it green requires deciding which repo/ref CI compares against and how a private repo authenticates — a ruling scope explicitly reserved. Tracked at: **#55**. Verified by: `gh issue view 55` → **OPEN**, title *"vendor-standards --check is the drift gate for four MIRROR standard sets and runs on the merge path nowhere — blocked on where CI gets the upstream"*; body contains the reproduction (`HOME=/tmp/nonexistent-home-xyz ... EXIT=1`), the two rejected workarounds, and the three open decisions. Also referenced by name in the `tests.yml` comment block so the omission is visible at the site.

- **`lint-prompts.sh` prints "every MODEL_KEY resolves" when Pass 3 never ran** — Why deferred: scope forbade modifying `lint-prompts.sh`, and the choice between "fail hard without `yq`" and "report the skip honestly" is a ruling, not a mechanical fix. Non-urgent because `test_model_keys_resolve.py` covers the same property and is now gated. Tracked at: **#57**. Verified by: `gh issue view 57` → **OPEN**, title *"lint-prompts.sh prints 'every MODEL_KEY resolves' even when the MODEL_KEY pass never ran — a green tick asserting an unchecked property"*; body quotes the `command -v yq` guard and the summary-line code, names the rename class as the consequence, and proposes the two remedies.

**Surfaced with no tracker, deliberately:** `SC2148` on `scripts/workflows/activities/run-claude.sh` and `scripts/workflows/common/shared-prompts.sh`. Both are sourced libraries with no shebang. `-s bash` handles it CI-side without suppressing anything a shebang would not have, so there is no pending work — if someone wants the directive added to the two files, that is a fresh decision, not a debt. Recording it here rather than inventing a tracker for it.

## Post-Run Reflection

**Friction**

- **Item 2's scope was unbuildable as written, and the task did not know that.** It asserted "Both run clean right now (exit 0), so gating them is a one-line addition each that cannot go red on arrival" — true on the dev workstation, false on a runner, because `vendor-standards.sh` resolves its upstream from a clone on local disk. The scope simultaneously (a) required the step, (b) forbade adding a fetch, and (c) required every step be demonstrated green. Those three cannot all hold. Resolving it meant declining an explicitly-scoped item, which is the right call but is not a cheap one to make autonomously — the temptation to ship `if [ -d ... ]` and call it done was the whole hazard, and it is exactly the shape the same task condemns two paragraphs earlier.
- **The "demonstrated green locally" constraint is what caught it.** Had the instruction been "add the step", item 2 would have merged and blocked the repo. Worth noting as evidence that the constraint earns its cost.
- Measuring shellcheck required installing it; `shellcheck-py` on PyPI turned out to be the clean answer because it pins in the same place and the same way as the existing `pytest`/`pyyaml`/`ruff` line. That was luck rather than guidance — the alternative (`apt-get` or a marketplace action) would have satisfied "pinned major range" far less tidily.

- **The `-s bash` question generalises.** Two files in `scripts/workflows/` are sourced libraries with no shebang, and nothing in `docs/standards/workflow-scripts.md` says whether that is the convention or an accident. Now that a linter reads the tree, "sourced libraries carry `# shellcheck shell=bash` and no shebang" (or the opposite) is worth one line in the standard, so the next person does not re-derive it from a CI flag.
- **`docs/file_structure.txt` had no `.github/` entry at all**, despite `tests.yml` predating this work and `CLAUDE.md` calling the map authoritative. Suggests the map's refresh trigger fires on `scripts/`- and `docs/`-shaped changes but not on repo-root dotfile directories.
- The `tests.yml` header's argument — *"a convention is not a gate"* — turned out to be a reusable test that finds work on its own. Running it across the repo produced this entire PR from a differently-framed issue. It may be worth stating explicitly somewhere as a periodic sweep: *list every control that exists, then check which are invoked by convention only.*

- The build-draft prompt's Stage 4 negative-control requirement was the highest-value instruction in this run. Seven controls were written and all seven fired, including one (NC7) that reproduced a real historical outage and proved the ordering of items 1 and 4 empirically rather than by argument. The instruction currently frames negative controls as applying to "structural/contract/grep-style tests" — **CI gate steps are squarely in that class and are not named.** Adding "CI workflow steps / lint gates" to that list would make the requirement land more reliably, since a gate that cannot go red is the purest form of manufactured confidence.
- Stage 3's EXECUTION-CONTEXT CHECK caught a real defect in draft. The first version of the `bash -n` step used `find ... | while read`, where `fail=1` is set in a subshell and lost — the step would have printed the syntax error and exited 0. Rewriting to `while ... done < <(find ...)` fixed it. The instruction names command substitution and pipelines explicitly; that specificity is what made it fire.
- **Gap:** the prompt has strong guidance on verifying deferral *pointers* by fetch, but none on the prior question — whether a deferral should become a **new** issue versus expanding an existing one. #53 was open and adjacent, but its body genuinely does not cover either deferred item, so two new issues were correct. The `engineering-quality.md` placement questions ("does it have a done-state today, is it an expansion of something that already exists") answer this, but they are not referenced from the Stage 5 deferral section where the decision actually gets made.

Pumapumapumas:
# build-refine pass

Four reviewers (code-reviewer, refactoring-evaluator, standards-auditor, quality-control) on a diff I did not author. Six defects fixed, three rejected with reasoning, five surfaced. Every measured claim in the draft's body was independently reproduced before being trusted.

## Decision Log

*Low-confidence first.*

- **[Low]** Corrected `CLAUDE.md` and `docs/standards/testing/README.md` even though both were stale **before** this PR. Alternatives: leave them (surgical-changes says don't fix adjacent pre-existing problems); surface only. Why: "pre-existing" is a claim I'm required to check rather than accept, and it does not survive here — this PR is what makes `CLAUDE.md:86` maximally wrong, and the testing README's *sole stated purpose* is to be the answer to "what does the gate cover" so nobody re-derives it. A doc whose only job is to describe X, left describing a third of X by the change that expanded X, is in-scope by any reading. Reasonable engineers could still call this scope creep.

- **[Low]** Fixed `docs/file_structure.txt:141`'s stale *"No README yet — see issue #35"* (#35 is closed, the README exists and is now load-bearing for finding 5). Alternatives: leave it, mention it. Why: the file was already open in this PR and the line points at the doc I had just rewritten. Two lines. Deferring it would have cost more to record than to fix.

- **[Medium]** Asserted `yq`'s presence in CI rather than extending `test_model_keys_resolve.py` to sweep the bash fleet. Alternatives: extend the pytest (~6 lines, quality-control's preferred remedy, makes the backstop genuinely real); correct the comment only (cheapest, leaves the coverage gap). Why: the assertion keeps the fix inside the gate file — the artifact this task is about — and makes Pass 3's execution a fact the merge path enforces instead of a property of the runner image. Extending the pytest is the better *long-term* answer and reaches further (it would cover the local case too), but it expands this PR into the test tree for a class already gated. **Judgement call; I would not argue hard against the pytest route.**

- **[Medium]** Kept the `settings.json` validator inline despite three reviewers (refactoring-evaluator High, code-reviewer Warning, quality-control Medium) recommending it move to `config/tests/unit/`. Alternatives: move it — quality-control argued the DoD is satisfied in substance because `run-all.sh` *is* a step in the `suite` job, so a pytest still runs on the merge path. Why: the task's definition of done says *"each of items 1-5 is a step in the `suite` job"* and repeats the placement reasoning twice. That is an explicit operator instruction, and the reviewers' case — which I find genuinely strong — is an argument for the operator to rule on, not for me to act on unilaterally. **I fixed the substance the finding was really about (correctness of the parse) and surfaced the placement.** This is the one disposition most likely to be overturned, deliberately so.

- **[Medium]** Did not SHA-pin `actions/checkout@v4`, `setup-python@v5`, `upload-artifact@v4`, despite quality-control rating it High and the file carrying a ten-line argument for tight pinning. Alternatives: pin all three. Why: three reasons compounding — the lines are pre-existing and untouched by this PR; SHA-pinning carries real upgrade-friction trade-offs that deserve an explicit ruling; and a wrong SHA on this job blocks **every merge in the repo**. The inconsistency quality-control identified is real and I am not disputing it; I am declining to resolve it unilaterally on a branch-protected job.

- **[High]** Rewrote the hook-path check to tokenise rather than patching the specific quoted-path bug. Alternatives: strip surrounding quotes (fixes the reported false-red, leaves the `~/` false-green); leave it, since both current hooks are canonical and neither bug fires today. Why: the reported symptom was one of six behaviours a substring split gets wrong, and "no live failure today" is exactly the reasoning that lets a false-green sit until it matters. `shlex` + expandvars + `checked == total` closes the whole class by construction, and I demonstrated all 11 cases rather than asserting them.

- **[High]** Accepted the draft's non-delivery of item 2 (`vendor-standards.sh --check`) rather than reopening it. Alternatives: build the checksum-manifest substitute now. Why: I independently reproduced the blocker — the script resolves its upstream from `$HOME/Repos/mdc-master-planning` or `/opt/skyy-net/...` and exits 1 when neither exists, which on a runner is every time — and the decline is correct. **But quality-control found the deferral bundled two separable checks**, and the cheap half (local-edit detection via a committed digest, no upstream access needed) is gateable today. Building it here would expand scope into `vendor-standards.sh`, which the task forbids touching; so I amended #55 with the decomposition instead of filing anything new.

## Deferred Work

- **Gate `vendor-standards.sh --check` on the merge path** — Why deferred: exits 1 on any runner lacking an `mdc-master-planning` clone; `main` is branch-protected on this job, so it would block every merge. Tracked at: **#55**. Verified by: `gh issue view 55 --json number,title,state,body` → **OPEN**, title *"vendor-standards --check is the drift gate for four MIRROR standard sets and runs on the merge path nowhere — blocked on where CI gets the upstream"*; body contains the `HOME=/tmp/nonexistent-home-xyz ... EXIT=1` reproduction, both rejected workarounds, and three open decisions. I also reproduced the blocker myself by reading `scripts/helpers/vendor-standards.sh:46-53`. **Amended this pass** with the local-edit/upstream-divergence split — comment posted at `#issuecomment-5226831181`.

- **`lint-prompts.sh` reports "every MODEL_KEY resolves" when Pass 3 never ran** — Why deferred: fixing it means editing `lint-prompts.sh`, which the task forbids, and the choice between "fail hard without `yq`" and "report the skip honestly" is a ruling. **Narrowed this pass:** the new `yq` assertion means it can no longer happen on the merge path; #57 now covers only the local-run case. Tracked at: **#57**. Verified by: `gh issue view 57 --json number,title,state,body` → **OPEN**, title *"lint-prompts.sh prints 'every MODEL_KEY resolves' even when the MODEL_KEY pass never ran — a green tick asserting an unchecked property"*; body quotes the `command -v yq` guard and the summary-line code, and names the rename class as the consequence.

Everything else reached FIXED, REJECTED-with-reasoning, or SURFACED-with-no-pointer. The five surfaced items are in the PR body under *"Surfaced — real, no verified home"*; I deliberately gave them **no tracker** rather than inventing plausible homes, so `review-pr` and the operator pick them up.

## Post-Run Reflection

**Friction**

- **The draft's self-report was accurate, which is itself worth recording.** Every measured claim I re-ran reproduced exactly — the shellcheck census to the finding (`error 2/0, warning 40/38, info 84/82, style 86/84`, 32 × SC2034), the 37-file count, the `vendor-standards.sh` blocker, both issue bodies. The defects were not in what it measured but in what it **asserted without measuring**: the sentence *"It is covered independently by test_model_keys_resolve.py"* was the single highest-severity defect in the PR, and it was one cross-file read away from being caught. **Measured claims were reliable; cross-file coverage claims were not.**

- **The task's own premise was false for one item and the draft had to discover that.** *"Both run clean right now (exit 0), so gating them is a one-line addition each that cannot go red on arrival"* — true on the workstation, false on a runner. The scope simultaneously required the step, forbade adding a fetch, and required green-before-PR; those three cannot all hold. The draft resolved it correctly. Worth noting that the *"demonstrated green locally"* constraint is what forced the discovery.

- **A stray `/tmp/types.py` left by an unrelated process shadowed the stdlib** and made all eight of my first negative-control runs fail with an `ImportError` that looked like my validator was broken. Cost one cycle to diagnose. `python3 -P` (or keeping scratch scripts out of `/tmp`) avoids it — worth knowing for any run that builds sandboxes there.

- **There is no standard governing `.github/workflows/` at all.** `tests.yml` is now ~200 lines, is the merge gate for every change, and has no conformance surface — no rule on pinning discipline, action-SHA policy, where check logic lives, or `permissions`. The `paths:` rule is only reachable via the *testing* standard. Two reviewers hit questions that were unanswerable for this reason. Suggest a trigger-gated item: **when a second `.github/workflows/*.yml` appears**, write the standard.
- **`docs/standards/documentation/README.md:36` excludes a whole section on a rationale that covers only part of it** — *"Deferred Work — GitHub Issues"* is dropped because *"the routing described is platform-specific"*, but that section also carries filing authority and checkbox-before-issue, which are not routing. A reviewer spent real cycles deriving whether §485 binds — the exact cost that file exists to eliminate.
- **`workflow-scripts.md`'s shebang carve-out names a `lib/` directory that no longer exists** (also stale at `:277`, `:287`). The two shebang-less files it is about now live in `activities/` and `common/`.
- The draft's own suggestion stands and I'll second it: *"list every control that exists, then check which are invoked by convention only"* is a reusable sweep that produced this entire PR from a differently-framed issue.

- **Stage 1's "mine the draft's reflection as leads, not conclusions" earned its cost twice over here** — but the *inverse* was the higher-value move and the prompt does not name it. The draft's reflection was honest and its measurements were sound; what needed checking was the confident prose in the **step comments**, which no stage points at. Suggest adding: *in a PR whose comments are load-bearing documentation, verify the comments' factual claims as rigorously as the code* — a false comment in a file reviewers are trained to trust is a higher-severity defect than the same statement elsewhere.
- **The build-draft prompt's negative-control requirement is the highest-leverage instruction in this workflow** (the draft's NC7 reproduced a real historical outage; my NC-A..M caught that the old validator was wrong in six ways). It currently frames negative controls as applying to *"structural/contract/grep-style tests"* — **CI gate steps are squarely in that class and are not named.** The draft flagged this too; two runs converging on the same gap is the evidence threshold.
- **Stage 3's disposition rules gave me no verb for "the reviewers are probably right, but acting on it would override an explicit operator instruction."** FIXED/REJECTED/DEFERRED/SURFACED all fit badly — I used SURFACED and explained, but the taxonomy pushed toward either overriding the operator or dismissing three reviewers. A fifth disposition (roughly *"deferred to operator ruling, with the recommendation stated"*) would fit the settings.json placement finding exactly, and that shape recurs whenever a DoD phrases a *means* and reviewers dispute the means while agreeing on the end.

Pumapumapumas:
# review-pr — disposition pass 1

Verified against the artifacts, not the account. CI is green (`suite` SUCCESS, 17s), the heredoc de-indents, and every measured claim I re-ran reproduced. **The producing run's self-report is unusually accurate** — I found no false measurement in it. What it could not know is that `main` moved underneath it.

## ⚠ ESCALATED — live on `main`, not this PR's, does not hold it

Two defects found while reviewing something else. Both sit in files this PR never touches, merging this PR changes neither one's severity, and both are one edit each. **The PR's verdict is computed as if these were not here.**

**E1 · `docs/standards/documentation/README.md:36` — the repo's applicability file tells every autonomous run that § Filing authority does not bind, so the next build run has documented permission to file its own deferrals.**
The bullet drops the whole *"Deferred Work — GitHub Issues"* section with the rationale *"the routing described is platform-specific."* Only the first sentence of that section is routing. The rest carries **§ Filing authority** (*"No autonomous dispatch may file deferred work as an issue… the PR-review stage triages what was surfaced and files what qualifies"*) and **§ Placement** (checkbox-before-issue, decide-from-the-body) — neither of which is routing, and the first of which describes *this repo's own pipeline by name*. Two runs on this PR already spent cycles deriving whether it binds and concluded it does not.
**Remedy (one line):** narrow the `README.md:36` bullet to the routing sentence, and state explicitly that § Filing authority and § Placement bind here.

**E2 · `docs/standards/workflow-scripts.md` — a binding MUST names a source path that does not exist, so a new workflow script written to the standard dies at `source` time.**
`lib/` appears at lines **118, 121, 264, 273, 277, 287, 541, 623**. Line 277: *"Every workflow script must source the shared `run_claude` helper from `scripts/workflows/lib/run-claude.sh`"*; line 287 shows `source "${SCRIPT_DIR}/lib/run-claude.sh"`. There is no `scripts/workflows/lib/` — the files are `activities/run-claude.sh` and `common/format-stream.sh`. Second consequence, specific to this PR: §118/§121 (*"sourced library files in `lib/` omit the shebang"*) is the **sole justification for this PR's `-s bash` flag**, and with no live subject a reviewer cannot tell whether the two shebang-less files are conformant or accidental.
**Remedy (one line):** correct the eight `lib/` references to `activities/` and `common/`, and state the sourced-library shebang convention against the directories that exist.

---

## What changed under this PR while it was in flight

`origin/main` gained **`9172cba` (11:37:54)** — a re-vendor at `3ffc223` adding **Testing Standard § Tier Enforcement → "A control that no gate runs is not a control (binding)"**. The branch's last commit is `4b1fadb` (11:44:30) and **does not contain it** (`git merge-base --is-ancestor` → false; `grep -c 'A control that no gate runs is not a control'` on the branch → `0`, on `main` → present). The operator's own commit message rules the consequence to this stage: *"Ruled at review, not by stopping the run."*

That clause binds three things this PR is about:
- an automated check that can fail is **on the merge path or documented as advisory — no third state**, and declaring it advisory means **naming what is consequently not covered**;
- **running is not blocking** — a control that cannot fail the merge is advisory in fact;
- a control **MUST establish the freshness of anything it compares against**, and *"a control that cannot establish its own baseline is not made trustworthy by being gated."*

**The six shipped steps are conformant with all three.** Each blocks (`main` is branch-protected on `suite`, and every added step lives in that job), and no shipped step compares against an unrefreshed baseline. The declining of item 2 is not merely defensible — the freshness clause now **forbids** exactly the two workarounds the run rejected, and the operator's commit records that `MDC-Master-Planning` is private with no secrets configured, so the step is impossible as scoped, not just unwise. **Item 2's non-delivery is ratified.**

---

## Dispositions

|---|---|---|---|
| `testing-readme-describes-the-pre-pr-gate` | doc-drift | **HOLD** — redispatch | The applicability row still reads *"tests.yml runs the master runner on every PR"* after this PR made it six checks, and predates the control clause. Fix-in-place, same file the PR already edits. |
| `settings-validator-cannot-be-verified-or-run` | test-gap | **HOLD** — needs-assistance | Its ability to fail lives in a PR body. Overriding the DoD's placement is the operator's call. |
| `gate-job-supply-chain-unpinned` | security | **HOLD** — needs-assistance | Six controls now ride one job whose three actions resolve through mutable tags. The trade is the operator's economics call; I may not self-grant the waiver. |
| `vendor-standards-control-ungated` | standards-implication | **DEFERRED** → [#55](https://github.com/helloskyy-io/Claude-Dot-Files/issues/55) | Verified OPEN, body contains the reproduction and the three open decisions. Ruling delivered below. |
| `filing-authority-released-repo-wide` | standards-implication | **ESCALATED** (E1) | Live on `main`, not this PR's file, one-line remedy. |
| `workflow-scripts-lib-path-does-not-exist` | standards-implication | **ESCALATED** (E2) | Live on `main`, not this PR's file, one-line remedy. |
| `issues-55-57-filed-by-an-autonomous-dispatch` | standards-implication | **NOTED** | Procedural violation is real; it produced no defect. Both issues ratified — see check below. |
| `wip-subject-in-branch-history` | scope | **NOTED** | `main`'s current practice is squash/rebase; the subject does not reach it. |
| `shellcheck-warning-level-32-sc2034` | standards-implication | **REJECTED** | The adopted level is clean with no baseline, no exclusions, no `\|\| true`, and the ratchet path is recorded. Gating at a noisy level is the failure #53 itself names. |

### Overturned from the producing run

**The run rejected quality-control's finding that filing #55/#57 violated `documentation_standard.md:485`, on the grounds that `documentation/README.md:36` excludes the section. That rejection does not survive** — the carve-out's own rationale covers only the routing sentence, and the excluded section names *"the PR-review stage"* as the filer, which is this repo's own pipeline. **The rejection is overturned; its remedy is E1, not a re-filing.** The run flagged the carve-out's over-breadth itself in the same comment, which is the tell: it used a rule to exempt itself and reported the rule as broken in the same breath.

**Ruling on #55/#57 themselves — NOTED, not reversed.** `no_live_defect_check`: I read both bodies end to end. Each states a consequence, carries reproduction evidence (`HOME=/tmp/nonexistent-home-xyz … EXIT=1`; the `command -v yq` guard quoted against the summary line), proposes concrete remedies, is exactly one item, and fails the fix-in-place bar because each needs an operator ruling. `gh issue view` → both **OPEN**. These are the issues the authorized filer would have filed. Re-filing them would be pure waste. Nothing is deferred and nothing breaks if this is never touched again; the *recurrence* risk is carried by E1.

### Ruling delivered — `vendor-standards.sh --check` (the operator's "ruled at review")

The PR's handling is correct and now doubly so. What is left is a ruling the new clauses **force**, and which #55's body — written before them — does not frame:

- `reframe:` not *"how do we make the drift step green on a runner"* but *"which half of the MIRROR guarantee do we actually need enforced — that nobody edited our copy, or that our copy still matches upstream?"* Only the second needs the upstream at all, and only the second can go red for reasons outside this repo.
- `bp:` the new clause is explicit that a control which cannot establish its own baseline is **not** made trustworthy by gating it. That eliminates #55's option 3 as a substitute for upstream comparison — but it does **not** eliminate it as a complete answer to local-edit detection, where the committed digest *is* the baseline and is refreshed by the same act that changes it.
- `recommendation:` **split #55 into the two halves the run already identified.** Ship the committed checksum manifest for local-edit detection — self-contained, needs no secret, cannot go red for upstream reasons, and it is the half `CLAUDE.md`'s MUST-NOT-EDIT rule actually asserts. Leave upstream-divergence as the part that genuinely waits on a secret. **#55's body needs amending to record the freshness MUST and the private-repo/no-secrets fact** before anyone rules from it; per § Placement Q2 that is an expansion of #55, not a new issue, so I have filed nothing.

### No issues filed, and why

Every candidate fails the three-criteria test: E1 and E2 are unrelated but **not substantial** (one edit each → escalated, not filed); `settings-validator…` and `gate-job-supply-chain…` are **the work in hand**; `vendor-standards…` is **already covered** by #55. The one genuinely homeless item — there is no conformance surface for `.github/workflows/` at all — fails **Placement Q1**: its remedy waits on a named trigger (a second workflow file), so it is a phase checkbox, not an issue. It rides `gate-job-supply-chain-unpinned` below.

---

## WHAT HAPPENS NEXT

**1 · redispatch — `build-minor.sh --pr 56`.** Rebase onto `origin/main` (to pick up `9172cba`), then correct `docs/standards/testing/README.md` only: the **Tier Enforcement** applicability row (it still describes the two-check gate this PR replaced, and predates the control clause), and one sentence in **§ The gate** naming what is consequently not covered by the ungated control — *a local edit to a vendored standard merges green today*. The new clause requires that naming; the section currently states the blocker without the consequence.

**2 · needs-assistance — `settings-validator-cannot-be-verified-or-run`.**
- `reframe:` not *"should this validator be inline YAML or a pytest"* but *"the repo binds **mutation evidence** — a guard ships with a demonstration that it fails when the property is violated — and this is the one guard whose demonstration cannot be committed or re-run."* Thirteen negative controls were produced and every one of them lives in a PR body.
- `bp:` a guard whose efficacy is asserted rather than maintained becomes a permanent pass the first time someone edits it — and this is simultaneously the only control in the repo that **cannot be run before pushing**, so no engineer would ever see it happen. That is the exact class this entire PR was built to close, reproduced inside the PR that closes it.
- `recommendation:` **the third option neither reviewer named — `scripts/helpers/check-settings.sh`.** It satisfies the DoD *literally* (the helper is invoked by a step in the `suite` job, matching `lint-prompts.sh`'s existing shape), restores local runnability, and makes the eleven NC-A..K cases committable as a pytest under `config/tests/unit/`, which `run-all.sh` already gates. This is your call because it reinterprets an explicit DoD instruction — but note the DoD phrases a *means*, and all three reviewers plus both runs dispute the means while agreeing on the end.

**3 · needs-assistance — `gate-job-supply-chain-unpinned`.**
- `reframe:` not *"should we SHA-pin actions"* in the abstract, but *"this PR concentrates six of the repo's controls into one job — what now rides on `actions/checkout@v4`, `setup-python@v5` and `upload-artifact@v4` resolving through mutable tags?"* The exposure is unchanged; **what depends on it tripled.**
- `bp:` the file carries a ten-line argument for pinning tightly enough that behaviour cannot change without a commit, and then depends on three references that can. The blast radius is bounded — `contents: read`, no secrets configured — so the realistic outcome is a poisoned green check, which is precisely the currency this job trades in.
- `recommendation:` pin all three to SHAs with the tag in a trailing comment, verified by this branch's own CI going green **before** merge (that retires the wrong-SHA-blocks-every-merge risk, which is the run's only non-waiveable objection of the three). I am explicitly **not** accepting "pre-existing" or upgrade-friction as dispositions — friction is a real cost and it is yours to price, not mine to waive. Related and unfiled: **there is no standard governing `.github/workflows/` at all** — no rule on pinning discipline, where check logic lives, or `permissions`. Trigger-gated on a second workflow file appearing, so it belongs as a phase checkbox rather than an issue.

---

```yaml
pr_review:
  pr: 56
  pass: 1
  attempt: 2
  verdict: HOLD
  converged: false
  findings:
    - id: settings-validator-cannot-be-verified-or-run
      title: the settings.json guard's ability to fail is asserted in a PR body and maintained nowhere, so a later edit turns the repo's only un-runnable control permanently green
      category: test-gap
      consequence: >
        Thirteen negative controls (NC-A..M) proved the validator can go red; none is committed or
        re-runnable. The repo binds mutation evidence ("a guard ships with a demonstration that it
        fails when the property is violated"). An edit to the inline Python that breaks its ability
        to fail leaves the gate permanently green, and because this is the only control in the repo
        that cannot be run locally, no engineer can detect it before pushing.
      disposition: hold
      remedy: fix-in-place
      hold_kind: needs-assistance
    - id: gate-job-supply-chain-unpinned
      title: six of the repo's controls now ride one job whose three actions resolve through mutable tags
      category: security
      consequence: >
        actions/checkout@v4, setup-python@v5 and upload-artifact@v4 are mutable references inside the
        job every merge in the repo depends on. Blast radius is bounded (contents:read, no secrets),
        so the realistic outcome is a poisoned green check — the exact currency this job trades in.
        Exposure is unchanged by this PR; what depends on it tripled.
      disposition: hold
      remedy: operator-action
      hold_kind: needs-assistance
    - id: testing-readme-describes-the-pre-pr-gate
      title: the file whose sole job is stopping reviewers re-deriving which half of the Testing Standard binds ships describing the two-check gate this PR replaced
      category: doc-drift
      consequence: >
        The "What binds today" row reads "Tier Enforcement | YES | tests.yml runs the master runner
        on every PR" after this PR made it six checks, and predates § "A control that no gate runs is
        not a control". § The gate names the ungated control without naming what is consequently not
        covered, which the new clause requires. The next reviewer re-derives — the exact cost this
        file exists to eliminate, and which already cost one review cycle per its own header.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
      reviewed_sha: 9172cba5c3ea5892c03a81c1d6a823441dd113b8
    - id: vendor-standards-control-ungated
      title: the MUST-NOT-EDIT rule on four vendored standard sets is enforced by nothing on the merge path, and the clause ratified today closes the workaround space #55 was written against
      consequence: >
        A local edit to a vendored MIRROR standard merges green today. The re-vendored freshness MUST
        eliminates the non-fetching workarounds, and MDC-Master-Planning is private with no secrets
        configured, so the step is impossible as scoped. #55's body predates both facts; whoever rules
        from it rules on a decision space the standard has since narrowed.
      disposition: deferred
      pointer: https://github.com/helloskyy-io/Claude-Dot-Files/issues/55
      pointer_verified: true
      reviewed_sha: 9172cba5c3ea5892c03a81c1d6a823441dd113b8
    - id: filing-authority-released-repo-wide
      title: the repo's applicability file tells every autonomous run that § Filing authority does not bind, so the next build run has documented permission to file its own deferrals
      consequence: >
        The carve-out's rationale ("the routing described is platform-specific") covers one sentence;
        the section it excludes also carries Filing authority and Placement. With it live, the disposal
        chute the standard exists to close is documented as open in this repo, and the operator's own
        review-pr tooling contradicts the repo's applicability file. Two runs on this PR already
        derived it the wrong way.
      disposition: escalated
      escalation_location: >
        docs/standards/documentation/README.md:36 on origin/main. This PR does not own it — it touches
        tests.yml and three other docs, never this file, and merging changes nothing about its severity.
    - id: workflow-scripts-lib-path-does-not-exist
      title: a binding MUST in the workflow standard names a source path that does not exist, so a new workflow script written to the standard dies at source time
      consequence: >
        scripts/workflows/lib/ does not exist; the files are activities/run-claude.sh and
        common/format-stream.sh. Line 277's MUST and line 287's example both point at the dead path.
        Separately, §118/§121 is the sole justification for this PR's `-s bash` flag, and with no live
        subject nobody can tell whether the two shebang-less files are conformant or accidental.
      disposition: escalated
      escalation_location: >
        docs/standards/workflow-scripts.md:118,121,264,273,277,287,541,623 on origin/main. This PR does
        not own it — the lib/ -> activities/+common/ move predates it and the file is untouched here.
    - id: issues-55-57-filed-by-an-autonomous-dispatch
      title: the producing run filed its own deferrals contrary to § Filing authority, but both issues are the ones the authorized filer would have filed
      consequence: none live — the recurrence risk is carried by filing-authority-released-repo-wide
      disposition: noted
      remedy: none
      no_live_defect_check: >
        Read both issue bodies end to end via `gh issue view 55/57`. Each states a consequence, carries
        reproduction evidence, proposes concrete remedies, is exactly one item, and fails the
        fix-in-place bar (each needs an operator ruling). Both OPEN. Re-filing would be pure waste.
        The producing run's rejection of this finding is OVERTURNED (its carve-out does not cover
        § Filing authority) but the correct remedy is the escalation, not a reversal.
    - id: wip-subject-in-branch-history
      title: a commit subject reading "wip … PRE-REVIEW, not yet audited" is in the branch and would land in main under a non-squash merge
      category: scope
      consequence: none live under current practice — squash-merge with a `ci:` subject is the action
      disposition: noted
      remedy: none
      no_live_defect_check: >
        `git log origin/main --merges -5` — the most recent merge commit dates to PR #20; the last 15
        commits on main are linear, so current practice is squash/rebase and a633a96's subject does
        not reach main under it. Head is now 4b1fadb, not a633a96 as the PR body states.
    - id: shellcheck-warning-level-32-sc2034
      title: gating shellcheck at warning would add 38 findings, 32 of them SC2034, to a gate whose value depends on nobody learning to ignore it
      consequence: none — error level is clean and the ratchet path is recorded
      disposition: rejected
      remedy: reject
  next_steps:
    - item: testing-readme-describes-the-pre-pr-gate
      kind: redispatch
      note: rebase onto origin/main, then correct the Tier Enforcement applicability row and add the not-covered consequence to § The gate
      dispatch_tool: build-minor.sh
      dispatch_context: |
        Rebase (or merge) origin/main into the PR branch so 9172cba is present — it re-vendors
        docs/standards/testing/testing_standard.md at 3ffc223, adding § "A control that no gate runs
        is not a control (binding)" to Tier Enforcement. The branch currently predates it.

        Then edit docs/standards/testing/README.md ONLY, two changes:

        1. The "What binds today" table row for **Tier Enforcement**. It currently reads
           "`.github/workflows/tests.yml` runs the master runner on every PR". That describes the
           pre-PR gate. Restate it against what tests.yml now does (six checks, all in the
           branch-protected `suite` job) AND against what the section now binds — every control, not
           only test tiers, per the newly-vendored clause.

        2. § The gate, the "One control is still off the gate" paragraph. It names
           vendor-standards.sh --check and the blocker but not the consequence. The new clause
           requires an ungated/advisory control be recorded "naming what is consequently not
           covered". Add that: a local edit to a vendored MIRROR standard merges green today.

        DO NOT TOUCH: .github/workflows/tests.yml (the six steps are conformant and CI is green),
        CLAUDE.md, docs/file_structure.txt, docs/standards/testing/testing_standard.md (vendored —
        MUST NOT be edited locally), docs/standards/documentation/README.md and
        docs/standards/workflow-scripts.md (both escalated to the operator, HiL). Do not change the
        gate's behaviour, do not add or remove steps, do not SHA-pin actions (separate ruling).
      precheck: |
        PRECEDENCE: the dispatch_context enumeration governs. The two edits above are the whole
        scope; this precheck gates WHETHER to act, never what to act on.

        Run from the PR branch worktree (NOT main — the README edit lives in this unmerged PR):

          git fetch origin
          git merge-base --is-ancestor origin/main HEAD && echo REBASED || echo NEEDS_REBASE
          grep -n 'runs the master runner on every PR' docs/standards/testing/README.md
          grep -c 'consequently not covered\|merges green today' docs/standards/testing/README.md

        Split stop predicate — these are different states:
        - Row already restated AND the not-covered consequence present (first grep empty, second
          grep >= 1) -> STOP, it is already done.
        - NEEDS_REBASE -> do NOT stop; performing the rebase is step one of the task.
        Different check than the one that surfaced this: the finding came from reading the row
        against tests.yml; this greps for the stale string and the missing consequence directly, so
        a wrong finding fails loud here rather than being executed into an induced edit.
    - item: settings-validator-cannot-be-verified-or-run
      kind: needs-assistance
      why_human: standards
      note: the DoD phrases a means; a binding standard clause cannot be satisfied under that means
      reframe: not "inline YAML or pytest" but "the repo binds mutation evidence, and this is the one guard whose demonstration cannot be committed or re-run"
      bp: a guard whose efficacy is asserted rather than maintained becomes a permanent pass on first edit — and this is the only control that cannot be run before pushing, so nobody would see it happen
      recommendation: |
        Take the third option neither reviewer named: scripts/helpers/check-settings.sh, invoked by
        the existing step. It satisfies the DoD literally (a step in the `suite` job, matching
        lint-prompts.sh's shape), restores local runnability, and makes NC-A..K committable as a
        pytest under config/tests/unit/ which run-all.sh already gates. Yours because it reinterprets
        an explicit DoD instruction — but the DoD phrases a MEANS, and three reviewers plus both runs
        dispute the means while agreeing on the end. If you keep it inline, the honest cost to record
        in the step comment is that its negative controls are not maintained.
    - item: gate-job-supply-chain-unpinned
      kind: needs-assistance
      why_human: standards
      note: no conformance surface exists for .github/workflows/ at all; the pinning trade is yours to price
      reframe: not "should we SHA-pin" but "this PR concentrates six controls into one job — what now rides on three mutable tags?"
      bp: the file argues for ten lines that behaviour must not change without a commit, then depends on three references that can
      recommendation: |
        Pin actions/checkout@v4, actions/setup-python@v5 and actions/upload-artifact@v4 to SHAs with
        the tag in a trailing comment, verified by this branch's own CI going green BEFORE merge —
        that retires the wrong-SHA-blocks-every-merge objection, the only one of the run's three
        reasons that is not a waiver. I am not accepting "pre-existing" or upgrade-friction as
        dispositions; friction is real and it is yours to price, not mine to waive. Related, unfiled:
        there is no standard governing .github/workflows/ — pinning discipline, where check logic
        lives, `permissions`. Trigger-gated on a second workflow file appearing (only tests.yml
        exists today; CodeQL runs via default setup, not a workflow file), so per Placement Q1 it is
        a phase checkbox, not an issue.
  laundered_deferrals:
    caught: 0
    of_total: 2
  homeless_items: 1
  redispatched: false
```

---

## Post-Run Reflection

**Friction**

- **The single highest-value check this pass was `git fetch origin` before anything else**, and the prompt frames it as a guard against prescribing something already done. Here it did the opposite — it revealed a clause ratified **7 minutes before the branch's last commit** that changes what conformance *means* for the artifact under review, plus an operator commit message explicitly routing the ruling to this stage. Suggest widening the CURRENT-TREE CHECK from *"has this already been done"* to *"what landed on the default branch since the branch was cut, and does it change the standard the PR is measured against"* — the second question is strictly larger and is the one that paid here.

- **`ESCALATED` did exactly what its rationale describes.** Both standards defects (E1, E2) are live, neither is this PR's, and under the pre-ESCALATED taxonomy both would have forced a HOLD on a green PR for files it never touches. Worth recording as a confirmed instance.

- **The producing run's reflection asked for a fifth disposition** — *"deferred to operator ruling, with the recommendation stated"* — for the settings.json placement finding. That shape already exists as `hold_kind: needs-assistance` with `reframe`/`bp`/`recommendation`, and it is what I used. The gap is that **`build-refine`'s prompt does not expose it**: the refine stage has FIXED/REJECTED/DEFERRED/SURFACED, and "SURFACED" has no recommendation slot, so a run that correctly declines to override an operator instruction has nowhere to put its reasoning except prose in a PR body. Suggest giving the build stages the same `reframe`/`bp`/`recommendation` triple, so review-pr inherits a structured recommendation instead of reconstructing one.

- **A note on where the defects were.** The run's *measured* claims were all reproducible; the defects it shipped and the ones I found were all in **cross-artifact reasoning** — a comment asserting another file's coverage, a rejection resting on an applicability file's scope, a doc row describing a gate two commits ago. The run said this itself last pass. Two runs converging on it is the evidence threshold: **the review effort is better spent on claims that span files than on claims that span lines.**

**Notes (below the finding bar — no consequence, recorded so they are not re-found)**

- The `ruff --select F821` sweep is scoped to `scripts/workflows/temporal` while this PR deliberately widened the bash sweeps tree-wide on an explicit principle. The only Python outside that scope is `./conftest.py`, whose F821 would fail pytest collection and redden `run-all.sh` — so the class is covered and there is no consequence. Asymmetry only.
- In the `settings.json` validator, the `checked != total` guard is unreachable: every path that skips the `checked` increment appends to `bad`, which exits first. Defensive redundancy, no consequence.
- The PR body states *"Head commit `a633a96`"*; head is `4b1fadb`.
- `lint-prompts.sh` discovers files by explicit directory globs, not a tree walk, so it never needed the `./.claude/*` exclusion the two sweeps got. The worktree-exclusion fix was complete — checked, not assumed.
- No predecessor deferral's trigger is satisfied by this PR: #51's two point at open `sprint.md` milestones (lines 180, 184) and #50's coherence finding is about component-pool candidate routing. Neither is touched here.

Pumapumapumas:
# build — applying review-pr's pass-1 ruling

Rebased onto `origin/main` (picked up `9172cba`, the re-vendored Testing Standard clause), then executed the two doc corrections and the two operator rulings from the disposition comment, in order.

## Decision Log

*Low-confidence first.*

- **[Low]** Did not update the shellcheck step's `-s bash` shebang-carve-out reasoning or touch `workflow-scripts.md`'s dead `lib/` references, even though the settings-validator relocation added a new file to the same tree. Alternatives: none seriously considered — this is E2's territory (escalated, live on `main`, operator's to handle). Why: explicitly out of scope per the task's constraints; flagging only.

- **[Low]** Placed the negative-control pytest at `scripts/helpers/tests/unit/test_check_settings.py` rather than `config/tests/unit/`, which is where the disposition comment's `recommendation:` field mentioned in passing. Alternatives: `config/tests/unit/` as literally written there. Why: the vendored Testing Standard's own binding text is "every library at `lib/<name>/` owns its tests at `<unit>/tests/<category>/`" — `scripts/helpers/` is the owning unit (matches the existing `scripts/workflows/temporal/tests/unit/` precedent exactly), and `config/` isn't. The disposition's mention read as illustrative, not prescriptive, and the task instructed "the placement the repo's testing standard indicates," not a specific path.

- **[Medium]** Extracted the validation logic into an importable `scripts/helpers/check_settings.py` rather than keeping it bash-only (as `lint-prompts.sh` is) and shelling out from pytest. Alternatives: pure-bash `check-settings.sh` with `subprocess`-based pytest cases. Why: this is the first CI helper in the repo pairing a committed pytest suite with its logic, and direct imports (`from check_settings import validate`) let each NC-case assert on the exact exception/message rather than parsing subprocess stdout — cheaper to write and cheaper to read. `check-settings.sh` stays a thin, locally-runnable entry point matching `lint-prompts.sh`'s shape from the outside.

- **[Medium]** Shrank the "Validate settings.json" CI step's comment to a pointer into `check_settings.py`'s docstring, rather than duplicating the full CATCHES/DOES NOT CATCH prose in both places (the "Prompt-construction lint" step keeps its full comment alongside `lint-prompts.sh`'s own header). Alternatives: keep both comments fully self-contained, matching the existing pattern. Why: unlike `lint-prompts.sh`, `check_settings.py` is a real importable module with its own maintained docstring — duplicating the algorithm rationale in two places invites the two copies drifting apart, which is the exact defect class (a comment asserting something not true of the code) two prior review passes on this PR already found and fixed. A reasonable engineer could still prefer the duplicated-comment pattern for consistency with the file's own established style.

- **[Medium]** Fixed a stale shellcheck census (`info 82, style 84` → `info 44, style 2`) in the Shellcheck step's comment, discovered incidentally while adding the SHA-pin comment two steps below it — not one of the three ruled items. Alternatives: leave it and surface it as a new finding instead. Why: verified false against the current tree (`error`/`warning`/`SC2034` counts all still matched exactly, `info`/`style` did not), it's a two-line fix in a file already open for this PR, and it is precisely the "confident prose in a step comment that reviewers are trained to trust" class this PR's own review history flags as highest-severity. Recording a deferral would have cost more than the fix.

## Deferred Work

Nothing new deferred. `vendor-standards-control-ungated` stays at **#55**, untouched — verified via `gh issue view 55 --json state --jq .state` → `OPEN` before pushing, per the disposition's explicit instruction not to touch or close it. The two standards escalations (`docs/standards/documentation/README.md:36`, `docs/standards/workflow-scripts.md`'s `lib/` references) are live on `main`, not this PR's files, and are the operator's to handle separately per the task's constraints — confirmed neither file appears in this branch's diff (`git diff origin/main...HEAD --stat`).

## Post-Run Reflection

**Friction**

- The disposition comment's `recommendation:` field for item 2 named `config/tests/unit/` as the placement in passing, while the task instructions above it said "the placement the repo's testing standard indicates." These pointed in slightly different directions once I actually read the standard's per-unit-ownership clause. Worth noting for future dispositions: a `recommendation:` field that names a specific path reads as more binding than intended if it wasn't checked against the standard itself.

- The disposition comment and PR body both used `config/tests/unit/` / a mixed NC-lettering scheme (NC1-7 for the draft's controls, NC-A..M for the refine pass's, with NC-A..K specifically being the eleven settings-validator cases) — cross-referencing which subset applied to item 2 took a full read of both the disposition comment and the PR body's own negative-control table before the mapping was unambiguous.

- The shellcheck adoption census (now corrected) is the second instance on this PR of a step comment recording a specific measured number that later stopped being true without anything catching it. `bash -n`'s file count already needed the same kind of fix (37→38) as a mechanical consequence of adding a file, but the shellcheck info/style counts drifted for a reason I could not fully reconstruct — the numbers were already wrong before this pass touched the file. Worth considering whether census-style comments should record *how* to re-derive the number (e.g., "run `shellcheck -s bash -f json ... | jq` and diff against this line") rather than only the number itself, so drift is mechanically checkable instead of only noticeable by someone reading closely.

No tooling-level suggestions from this pass — `review-pr`'s disposition comment was unusually actionable and needed no interpretation beyond the one placement question noted above.

---

**Verification:** `./testing/run-all.sh` green (349 tests, was 337 before this pass), all six `suite`-job gate steps demonstrated locally before push, and this branch's own CI (`suite`, `Analyze (actions)`, `Analyze (python)`, `CodeQL`) confirmed green on the final pushed commit (`49e273d`) via `gh pr checks 56 --watch` — retiring the wrong-SHA-blocks-every-merge risk the SHA-pinning ruling named as its one non-waiveable condition.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

Issue #65, opened 2026-08-08
The turn-cap banner asserts 'NOTHING was committed or pushed' without checking, and it was false the first time it mattered


Pumapumapumas:
## What happened

PR #56's redispatch (`build-minor --pr 56`, 2026-08-08) hit the 100-turn cap **while printing the PR URL** — the completion signal, the last step. `run-claude.sh` printed:

```
⚠ RUN TERMINATED AT TURN CAP (100 turns)
  Work is uncommitted at: .../worktrees/build-1786206125
  NOTHING was committed or pushed.
```

**Every deliverable had already been pushed and CI was green.** Verified on `origin/ci-gate-executability-drift`: commit `84e86e7` (both operator rulings — settings guard relocated, three actions SHA-pinned), `49e273d`, +556/-9 across 8 files including `scripts/helpers/check_settings.py` and `tests/unit/test_check_settings.py`; `suite`, CodeQL and both Analyze checks passing; PR body updated; reflection comment posted.

## The consequence

The banner states as fact something it never checks. `run-claude.sh:167` fires on `"subtype":"error_max_turns"` and prints the message without inspecting the tree.

An operator who trusts it goes to salvage a worktree with nothing to salvage — and, worse, may **re-dispatch work that is already merge-ready on the PR**, paying a second full budget and risking a conflicting branch. That is more expensive than saying nothing: a wrong claim about state sends someone somewhere, while an absent one sends them to look.

This is a different defect from the one closed as tracker item T-16 the same morning. T-16 was *the path is not named* — that is genuinely fixed (`dd373d4` prints the worktree path twice). This is *the claim beside the path is unverified*.

## Remedy

The handler already has the worktree path in hand. Before asserting anything, check:

```bash
git -C "$wt" status --porcelain              # is anything actually uncommitted?
git -C "$wt" log origin/<branch>..HEAD --oneline   # committed but unpushed?
```

Report what is there — *uncommitted changes*, *N commits not pushed*, or *everything is pushed; the cap fired after the work landed*. The third case is this one, and it should read as **"check the PR before redispatching"**, not as a salvage instruction.

## Related

[...]

PR #58, opened 2026-08-08
build-draft: characterization tests for block-dangerous.sh (closes #52)

Pumapumapumas:
Closes #52.

A characterization suite for `config/hooks/block-dangerous.sh` — the `PreToolUse` hook that is the only safety control operating *during* an autonomous run. **218 tests, zero changes to the hook itself.**

## Update — PR-REVIEW pass 1 disposition, six findings fixed

`PR-REVIEW disposition — pass 1` held six findings as fix-in-place (all in files this PR already edits) and deferred three to trackers (#61, #63, MDC-Master-Planning#148 — untouched here) plus one operator ruling (issue #52's pytest-for-bats substitution — also untouched here, being recorded by the operator directly). Rebased onto `origin/main` at `9172cba` first.

**build-refine-minor pass:** `origin/main` had moved three more commits past `9172cba` by the time this landed (two workflow-prompt commits plus a standards doc fix — none touching a file this PR edits). Re-rebased onto the current tip (`07eeec5`), re-ran all three required mutation re-demonstrations (findings 1, 2, 4) against the rebased tree — all reproduce identically — and re-ran `./testing/run-all.sh` (560 passed) and bare `pytest` (560 passed) post-rebase. An independent `code-reviewer` pass over the six-fix commit's diff found no Critical or Warning issues; its Info-level notes (pytest exit-2 also covering manual interrupt, the coverage-guard's line-scanner not tracking nested loops, `test_testpaths_matches_discovery.py`'s prune-list re-implementation being a strictly-narrower — and today inert — divergence from `python.sh` at the `.git`/`.claude` anchoring) are pre-existing, already-documented tradeoffs or non-issues, not new defects.

**Every count and transcript below this point that says 218 / 555 predates this update and is left as the historical record of what those runs actually produced.** Current counts: `test_block_dangerous.py` now collects **219** tests (+1), and `./testing/run-all.sh` now reports **560** (+5: the 1 above, plus 4 in a new `test_testpaths_matches_discovery.py`). Bare `pytest` also reports 560, matching `run-all.sh`.

[...]

Issue #64, opened 2026-08-08
Nothing separates what is live from what is being developed — in CDF the working tree IS production, and the MDC ring boundary has no home yet


Pumapumapumas:
Parked deliberately on 2026-08-08 for a session with room to think. **Not blocking anything.** Filed so the reasoning survives; it was worked out under time pressure and deserves a clear head.

## The consequence

**CDF:** `~/.claude/{agents,hooks,rules,skills,commands,settings.json,CLAUDE.md}` symlink to `/home/puma/Repos/claude-dot-files/config/*` — **the git working tree.** So whatever branch is checked out is what every Claude session on this machine is running, and an edit is live before commit, before push. Demonstrated the same day this was filed: prompt edits took effect in two dispatches with no commit and no pull. Convenient while tuning prompts; undefined as an answer to *"which version is live."*

**MDC:** MDC#1 is developed on while live because a second cluster is not affordable. That is a legitimate arrangement — it is a canary ring — but nothing currently enforces the ring boundary, so there is no mechanism preventing MDC#2 receiving something MDC#1 has not survived.

## Why a DEV branch is NOT the answer (this is the part worth not re-deriving)

The operator's initial instinct was a permanent DEV branch: MDC#1 deploys from DEV, MDC#2 from MAIN. **The goal is right and the mechanism is one the industry moved off.** Branch-per-environment is a documented GitOps anti-pattern — Argo CD's own guidance says so:

1. Each branch accumulates environment-specific values, so DEV→MAIN stops merging cleanly
2. Promotion becomes a merge, fusing *is this code good* with *should prod get it* — two decisions, two approvers
3. *"What is running on MDC#2"* becomes a branch diff instead of a lookup
4. **It does not survive N>2.** With server + edge + multiple edges, a hotfix must be cherry-picked into every branch — which is precisely the future this was raised about

**And a DEV branch would deliver ZERO isolation in CDF regardless**, because the symlinks follow the working tree. Checking out `dev` just makes `dev` live.

**Also relevant:** the earlier DEV-branch attempt failed because new-branch dispatches pass `HEAD` as the worktree base, so they treed from whatever feature branch was current. That is a one-line defect, not evidence against branching — filed separately. The evidence that killed the earlier attempt does not support the conclusion it was used for.

## The direction to evaluate

**Environment-per-directory with promotion as a commit** — the pattern that replaced branch-per-environment.

[...]

PR #66, opened 2026-08-08
build-draft: MMF Phase 1 — measure the channel (six experiments, thirteen rulings, one downstream predicate cancelled)

Pumapumapumas:
## Summary

Executes **Phase 1 — Measure the channel before designing it** of the Memory Management Framework. All six experiments (E1, E5, E7, E2, E3, E6) ran against the pinned CLI (**2.1.224**) and the archived artifacts. **39 of 39 checkboxes complete. Thirteen rulings recorded, each naming a downstream consequence.** This phase measures; nothing was implemented.

Draft committed **per experiment** (six pushes). The refine pass then re-took three rulings — see below.

## The refine pass reversed this PR's highest-consequence ruling

**E7's draft ruling would have cancelled a predicate that works.** It measured the finding-id delta over the `pr_review:` block's **cumulative** id set. `review-pr.sh:221` has each pass restate every prior id with its `disposition` updated in place, so that set is **monotone by construction** — "0 of 7 empty" and "0 of 25 dropped" observed the reporting schema, not the fleet. Verified directly against PR #58: all 10 of pass 1's ids reappear in pass 2, seven flipped to `disposition: fixed`.

Re-measured over the **open** subset (`disposition` not in `fixed`/`deferred`/`rejected`/`noted`):

| | all ids | open subset |
|---|---|---|
| pairs with an empty delta | **0 of 7** | **2 of 7** |
| blocks whose set is empty | 0 of 14 | **1 of 14** — PR #42 pass 2 |

PR #42 pass 2 is the only `MERGE` and the only `converged: true` in the archive. So:

- **Phase 5's predicate is RE-SCOPED, not replaced.** The mechanism was right; the *set* was wrong.
- **The heuristic and the computed signal AGREE, 1 of 1** — the draft reported them as disagreeing, which was the load-bearing premise for replacing the mechanism.
- **`findings[].severity` leaves E6's envelope.** It appears on **0 of 195** archived findings. `findings[].disposition` replaces it — present on all 195.
- **The stopping condition is the open set being EMPTY, not UNCHANGED.** At #58 pass 2→3 the open set was unchanged *and non-empty*; stopping there would have been wrong. Both halves now measured.

Both the superseded ruling and the correction are kept in the doc, because the failure mode — reading a set's monotonicity as a fleet property when the schema guarantees it — is the reusable lesson.

### Two other rulings re-taken

- **E5's denominator.** The third strict-negative was the draft run's **own in-flight log**, adjudicated in the table as a "truncated log" fleet observation. It has since completed and now matches strictly. Corrected to **72 completed logs / 50 in-scope / 2 strict-negatives**. The **0 miss count is unchanged under either accounting**, so no ruling moves — but the tool now lists envelope-less logs separately, since any run invoking it has its own open log in the directory.
- **E6's field list**, which inherited E7's `severity` error.

**No E1, E2 or E3 ruling moved**, and every measured figure in those experiments reproduced exactly on independent re-run.

## The rulings

| Exp. | Ruling | Downstream consequence |
|---|---|---|
| **E1(a)** `is_error` | no-op | `is_error == (exit != 0)` on **8 of 8** forced modes. Removed from Phase 3's envelope |
| **E1(b)** `num_turns` vs cap | no-op | Reported **`num_turns: 2` against a cap of 1**. Telemetry only, never a branch input |
| **E1(c)** `subtype` | confirms | `error_max_turns` and `error_max_budget_usd` **both exit 1** — only `subtype` separates them |
| **E1(d)** `.result` on error | **changes** | The `result` **key is absent** on every error subtype. The prose channel's failure mode is *non-existence*, not misparsing |
| **E1(e)** auth ≡ SIGTERM | **changes** | Identical envelope. **And neither exited on its own** — both hit our 120 s `timeout` after 9 retries. Nothing in `run-claude.sh` wraps the CLI in a timeout |
| **E1(f)** `permission_denials[]` | **changes** | Forced denial exited **0**, `is_error: false`, `subtype: success`. Required field |
| **E1(g)** transport | confirms | `structured_output` wins on availability, isolation and Temporal replay cost |
| **E5** prose-grep misses | no-op (for the defect argument) | **0 adjudicated misses / 50** completed in-scope logs, **0 / 14** for the VERDICT token, loose set **identical** to strict |
| **E7** convergence delta | **changes** | **Re-scoped to the open subset (2/7 empty, 1/14 blocks empty), not replaced.** `disposition`, not `severity`, is its input |
| **E2** partial records | confirms + changes | 0 of 4 forced deaths left a partial record — one Phase 3 requirement **dropped**. A *silent* success-path absence found — one **added** |
| **E3(a)** disagreement policy | no-op | Off-diagonal cells empty **by construction**, not by small N |
| **E3(b)** to-do bit | **changes** | Typed verdict disagreed with PR disposition in **6 of 7**; **31 of 38** PRs have no typed verdict at all. `open` owns it |
| **E6** envelope union | **changes** | **Nine fields, not the roadmap's "roughly five"** — enumerated across both fleets |

## Asserted facts checked; four had drifted

1. **`--max-turns` is no longer in `claude --help` on 2.1.224.** It still parses, but the flag **every dispatch passes** (`run-claude.sh:139`) is undocumented on the installed binary — and `§Runtime Verification` verified two flags the design *might* use while omitting the one the fleet depends on.
2. **`--json-schema` takes an inline JSON string, not a file path.**
3. **`.claude/logs/` held 73, not the 60 recorded 2026-08-07** (74 by the end of the refine pass — it grew twice during this PR).
4. **`build-phase.sh` is not a parent.** It invokes no child. Two bash parents, not three.

Verified unchanged: `run-claude.sh:159`'s 0.9% (4/443) cite, the `COMPLETION_PATTERN` counts (11 bash / 21 language-agnostic), all `build.sh`/`build-minor.sh`/`review-pr.sh`/`routing.py` line citations. **`claude_code_integration_surface.md` §5/§7 checked, not assumed: `Last validated: 2026-07-25`, `Revalidate: high — 4 weeks` → due 2026-08-22, current today.** Its version anchor is 2.1.220 while the installed CLI is 2.1.224, and E1 found **ten envelope fields its §7 does not list** — flagged for its revalidation sweep, not edited here.

## Downstream phase amendments (every one)

- **`phase3_typed_exit_record.md`** — step 1 (E1: `is_error` out, `num_turns` telemetry-only, `permission_denials` required; E2: every required field must be one the child can always fill); step 5 (**one requirement DROPPED**, **one ADDED**); step 6 (two rows ruled by E3). *Phase 3 references E6's field list by pointer rather than restating it, so the `severity`→`disposition` correction flows through without an edit.*
- **`phase5_convergence_stopping.md`** — step 2 (premise confirmed on the added direction; **the "0 dropped" half withdrawn** as a schema artifact); step 3 (**predicate re-scoped to the open subset**, `severity` explicitly ruled out); step 4 (one mode re-answered, **one new mode added**: a pass that closes nothing and adds nothing).
- **`roadmap.md`** — six Phase-1 checkboxes ticked, status line updated.
- **`docs/file_structure.txt`** — new `scripts/helpers/measure/` entries, plus two stale lines corrected (the component was still annotated `PLANNED` / "Five experiments").
- **Phase 4 deliberately NOT amended** — E1(c)/(e) change what it reads, but its checklist already accommodates both.

## What E5 feeds `D-007` (not ruled here — it is the operator's)

- **The miss count is 0/14 for the token and 0/50 completed in-scope logs**, loose set identical to strict.
- **`D-007`'s premise "ours has none" is FALSE as stated.** A write-time gate already exists — `run-claude.sh:201-204` fails the run loud when the pattern is absent from `.result`. What it does *not* cover is the parent's wider surface: `build.sh:277` greps the child's whole console. **The ruling `D-007` needs to make is the narrower one of whether that surface mismatch is worth closing.**

## Scope

`sprint.md` untouched · `docs/standards/` untouched · `memory-management-framework/research/` read-only, cited only · **no implementation**.

**Sprint-item candidate for human-reviewed editing (surfaced, not written):** `sprint.md:143-145` still reads `📋 QUEUED, NEEDS PLANNING` / `Phase doc: not yet written` for this component. A roadmap and five phase docs now exist and Phase 1 is measured. Flipping that line is the operator's edit.

## Tests

- **`./testing/run-all.sh` → 627 passed, RESULT: PASSED** (574 before this PR; 615 after the draft; +12 from the refine pass).
- **`./testing/suites/python.sh unit helpers` → 52 passed** (12 pre-existing + 40 new).
- `scripts/helpers/lint-prompts.sh` → clean.
- **CI on the pushed branch: all four checks green** (`suite`, `Analyze (python)`, `Analyze (actions)`, `CodeQL`).

**Negative controls — 11 written, 11 fired**, via the repo's existing `testing/scripts/mutate.sh`. Five from the draft (verdict anchoring, strict-vs-loose gap, absent-vs-empty `.result`, absent-vs-false `converged`, fence matching) plus six from the refine pass:

| # | Property falsified | Result |
|---|---|---|
| NC6 | `open_ids` reverts to a hold-only test (fails open on `escalated`/unknown) | **2 failed** |
| NC7 | per-finding entry span unbounded, so a `disposition` leaks across findings | **2 failed** |
| NC8 | `PR_OR_ISSUE_URL` narrowed back to pull-only | **1 failed** |
| NC9 | unparseable-line count reported as always-zero | **3 failed** |
| NC10 | `last_whole_match` reverts to `findall` (tuple shape) | **2 failed** |
| NC11 | `last_whole_match` returns the first match instead of the last | **1 failed** |

## Tooling fixes (both tools are kept to be re-run, so each would corrupt a *future* measurement)

- `replay_pr_review_blocks.py` — extracts per-finding `disposition`/`category` and computes `open_ids`; **unknown disposition counts as OPEN** (fail-safe: the archive already carries a sixth value, `escalated`, that a hold-only reading would have scored as closed).
- `replay_completion_predicate.py` — `PR_OR_ISSUE_URL` for `plan-revision`/`plan-new`, whose gates accept a STOP **issue** URL (verified: no current count changes, both archived logs used `pull/`); `last_whole_match` replacing a `findall()` that returned a tuple for verdict rows and a string for URL rows; unparseable-line counting; in-flight logs listed for adjudication; the worktree-default caveat documented.
- Both test files flattened from `class Test` grouping to the sibling `test_check_settings.py`'s convention — `class Test` appears nowhere else in the repo.

## Surfaced for the operator (measured, deliberately not fixed — this phase measures)

1. **Auth failure is an unbounded hang.** Both auth runs retried 9× toward `max_retries: 10` and exited only because *we* wrapped them in `timeout(1)`. `run-claude.sh` has no timeout. **No issue covers this** — raised naked, no pointer. E1(e).
2. **The turn-cap rate's only re-measurable sample runs 3–4× the recorded figure.** `run-claude.sh:159` records 0.9% (4/443, 3 from April); the archive shows **2 in 72 completed logs (2.8%)** / **2 in 50 in-scope (4.0%)**, both from August. Different populations, so **not** a claim the rate rose — but that comment's own reopen condition is "if the rate climbs". E5.
3. **A run died at its turn cap *after* opening a PR** (`build-draft-minor-20260808-122206`). Corroborates **issue #65**, open on exactly this. E5.
4. **Two `system` subtypes missing from `liveness_signal_measurement.md` §2.3**: `api_retry` and `permission_denied`. An *addition*, not a contradiction; its next revalidation sweep is the right home.
5. **`converged`, `attempt` and `hold_kind` have zero programmatic readers** across both fleets. E6.


Pumapumapumas:
## Decision Log

Ordered low-confidence first.

- **[Low]** Ruled E6's envelope at **nine** fields rather than trimming toward the roadmap's "roughly five". Alternatives: hold `permission_denials` out (no *code* consumer today, only an operator one); collapse `completion_ref` from three fields to a bare URL string and keep the parsing. Why: the checklist's bar is *"a field with no NAMED consumer does not enter"* — not "no code consumer" — and E1(f) measured that the operator question *"did a dispatch get blocked?"* is unanswerable from anything else. `completion_ref.kind` exists to delete `plan_revision_workflow.py`'s `rfind` positional tie-break, which is a real code deletion, not a nicety. **A reviewer could reasonably argue nine is too wide; the excluded-fields table is where to attack it, and it is deliberately as long as the included one.**

- **[Low]** Adjudicated **0 of 25** added finding-ids as restatements under a new slug, so E7 reports the stable-id convention as holding cleanly. Alternatives: score the four near-miss pairs as borderline and report "0 clear, 4 borderline". Why: I read the full finding **bodies** for the two closest pairs (`quality-control-…-no-slot` vs `security-lens-…-no-slot`, and `merge-drops-the-model-key-guard` vs `merge-drops-executability-guard`) and both are genuinely different lines with different fixes. **The residual risk is that the other three were adjudicated on slug and title alone, not on body** — stated in the doc, and a re-adjudication would be cheap.

- **[Medium]** Excluded **22 of 73** archived logs from E5's denominator as anachronistic (retired `revision`/`revision-major`, and `review-runs` which declares no pattern). Alternatives: report 73 with a footnote; report both. Why: applying today's predicate to a log from a workflow that declared none measures a rule that was not in force, and 5 of the 8 raw strict-negatives are in that group — reporting 73 would have produced "8 non-matches" as the headline against a real figure of 3, all three of them correct rejections. The exclusion is **named with its membership**, not folded silently into a smaller denominator.

- **[Medium]** **Ran E2 instead of closing it**, against the checklist's explicit offer to close it if E1 selected `structured_output`. Alternatives: close it and cite E1(g), which was sanctioned and would have saved four API calls and a turn. Why: the offer rested on the assertion *"`structured_output` … has no partial-record class"* — an unverified claim about a vendor binary, which is the exact class this phase exists to stop trusting. It cost four haiku calls. The assertion held, **and** the run found the silent success-path absence (model declines to call the `StructuredOutput` tool → exit 0, `subtype: success`, no record), which is now an added Phase 3 fail-safe rule. Closing E2 would have shipped Phase 3 without it.

- **[Medium]** Ruled E3(a) on a **structural** reason rather than on the observed count. Alternatives: write "0 of 14, revisit at larger N", which is the conventional shape. Why: the off-diagonal cells cannot be populated at *any* N — the asserted verdict rides in the same envelope key E1(d) measured as absent whenever `is_error` is true. A count-shaped justification would invite a future reader to re-take a measurement that can only ever return zero. Phase 3's amendment says this explicitly so the reasoning survives the edit.

- **[Medium]** Amended **Phase 5's step 3 to replace the predicate** rather than annotating it as at-risk. Alternatives: leave the checklist and add a warning note; leave it entirely and let Phase 5 discover it. Why: the phase doc mandates that a *changes-the-design* ruling amends the affected phase's checklist **before it starts**, and 0-of-7 with a strictly-growing set is not a marginal signal. I kept the set-difference computation as an input to step 4's windowed check rather than deleting it, because that check genuinely needs it. **The severity-based replacement is named but not designed here** — designing it would be Phase 5's work leaking into Phase 1.

- **[Medium]** Did **not** amend Phase 4, though E1(c) and E1(e) both change what it reads. Alternatives: amend it for symmetry with Phases 3 and 5. Why: its checklist is written at an altitude that already accommodates "read `subtype`" and "auth failure needs the stream events", so an amendment would restate a Phase 1 ruling rather than redirect any step. Named explicitly in the phase summary so the omission is visible as a decision rather than an oversight.

- **[High]** Used a **benign copyright decline** ("print the lyrics of Hey Jude") to measure the model-decline row of E1 rather than authoring a policy-violating prompt. Alternatives: record the row un-measured. Why: it produces the observable the row is for — whether a model-level decline leaves any envelope trace — without writing a harmful request. It does not (exit 0, `subtype: success`, decline prose in `.result`), which is the finding.

- **[High]** Kept the two replay tools in `scripts/helpers/measure/` instead of deleting them as one-shots. Alternatives: delete both, per the brief's "or delete them if they were one-shot". Why: neither is one-shot — E5's number is what `D-007` is waiting on and its denominator grew 60→73 in a single day; E7's is what Phase 5 will re-run. The README states that reasoning in the directory itself so the next reader does not have to reconstruct it. The E1/E2 **forcing** scripts *were* one-shot (they burn API calls) and were left in `/tmp`.

## Deferred Work

- **`run-claude.sh` has no timeout, so an auth failure hangs unboundedly** — Why deferred: this phase measures; fixing the incumbent runtime is an implementation change explicitly out of scope, and the right sizing (a `timeout` wrapper vs. reading `system/api_retry` from the stream) is a design call. Tracked at: **nowhere — SURFACED WITH NO POINTER.** Verified by: `gh issue list --repo helloskyy-io/Claude-Dot-Files --state all --limit 40 --search "timeout auth retry hang"` → no issue covers this; and `grep -n "timeout" scripts/workflows/activities/run-claude.sh` → **no match**, confirming the gap is live. I am not filing an issue because the two placement questions both point away from one: it has a done-state today (so not a phase checkbox) but it is squarely **Phase 4's** subject matter — Phase 4 owns what parents read from failure envelopes, and E1(e) is already written into this PR as its input. Raising it naked here so it is picked up rather than laundered.

- **The turn-cap rate may have climbed above the 0.9% the code comment reopens on** — Why deferred: the two denominators are different populations (443 runs vs 73 archived logs), so this is a signal to watch, not a measurement that settles it. Tracked at: **`scripts/workflows/activities/run-claude.sh:163-165`**, the existing comment's own reopen clause. Verified by: `git show origin/main:scripts/workflows/activities/run-claude.sh | sed -n '156,166p'` → the block is live on the default branch and reads *"Reopen only if the rate climbs under unattended/pooled operation, and even then the fix is louder signalling, not resume."* This is an **expansion of something that already exists** (placement question 2), and E5's cross-check is now recorded in the phase doc's incumbent-surfaces table beside the cite. No new issue.

- **`liveness_signal_measurement.md` §2.3's event vocabulary is missing `system/api_retry` and `system/permission_denied`** — Why deferred: it is a **vendored-discipline** matter — research papers are critic-gated evidence and this phase must not edit the pool. Tracked at: the paper's own revalidation clause. Verified by: `sed -n '10,12p' docs/development/fleet-reliability/research/raw/liveness_signal_measurement.md` → `Last validated: 2026-08-07` / `Revalidate: high — 3 weeks`, i.e. **due 2026-08-28**, an unfired trigger with a date. Placement question 1 says a trigger-gated item is not an issue; both event shapes are recorded verbatim in this PR's phase doc for the sweep to pick up.

- **Nine `pr_review:` envelope fields need a cross-language schema owner** — Why deferred: it is `phase3_typed_exit_record.md` step 6's third open question and belongs to that phase, not this one. Tracked at: **`docs/development/memory-management-framework/phase3_typed_exit_record.md`**, step 6, third bullet. Verified by: `git show origin/main:docs/development/memory-management-framework/phase3_typed_exit_record.md | grep -n "Cross-language schema ownership"` → present on the default branch at the *"Rule between a single language-neutral schema artifact both trees load, and two declarations plus a cross-tree conformance test"* bullet. Unchanged by this PR; E6's field list is its input.

## Post-Run Reflection

**Friction**

- **The brief's warning that "a bash-shaped grep silently measures half this fleet" earned its place, and it bit exactly where predicted.** E6's single highest-value branch point — `plan_revision_workflow.py:86-95`, where a run may complete via an **issue** URL and the parent disambiguates by `output.rfind()` position — is Python-only and invisible to every bash-shaped pattern. Without that warning the envelope would have shipped with a `completion_ref` that only models PRs.
- **The phase doc's `§Runtime Verification` block verified the wrong flags**, and this is a generalisable trap. It checked `--json-schema` and `--output-format` — flags the *design might use* — and omitted `--max-turns`, the flag *every dispatch already passes*. `--max-turns` has since vanished from `claude --help`. **A runtime-verification block should start from what the code already depends on, then add what the design proposes.**
- **`--json-schema` taking an inline string rather than a path** cost two failed runs. Nothing in the cited research says which, because "the flag exists" is where flag-availability research usually stops. Worth carrying into the integration-surface paper's revalidation.
- **`build-phase.sh` was named as a parent in E6's checklist and is not one.** A one-line check at plan time (`grep -c 'children/' build-phase.sh`) would have caught it. Minor, but it is the third planning-asserted fact in this task that did not survive contact.

- **`.claude/logs/` grew 60 → 73 in a single day, and every count taken over it dates that fast.** Several docs now carry counts over this corpus. A convention of writing counts as `N as of YYYY-MM-DD` (which this phase doc now does) is cheap; a `scripts/helpers/measure/` tool that re-derives them is cheaper still, which is the argument for keeping the two added here.
- **`testing/scripts/mutate.sh` is excellent and under-advertised.** Its own docstring records that four runs on a sibling repo hand-rolled the same loop before it existed. I found it only because the dispatch brief happened to mention `mutate.sh` in passing. It deserves a line in the Testing Standard's README next to the efficacy rule, so a run reaches for it instead of rebuilding it.
- **Three shipped `pr_review:` keys — `converged`, `attempt`, `hold_kind` — have zero programmatic readers.** That is fine and intentional today, but it is invisible: nothing distinguishes "human-facing by design" from "was supposed to be wired and never was". Phase 3's publishable/internal classification is the natural place to make that explicit, and E6's ruling now feeds it.

*Flagging for the architecture session; not editing.*

- **`run-claude.sh` wraps `claude` in no timeout, and an auth failure therefore hangs.** Measured this run: 9 `system/api_retry` events at `error_status: 401` against `max_retries: 10`, exiting only because the measurement harness supplied its own `timeout(1)`. Under unattended operation a credential expiry becomes a stuck dispatch with no upper bound. This is a claude-dot-files-level change and the fix has two shapes worth weighing (a wall-clock `timeout` wrapper vs. reading `system/api_retry` from the stream the fleet already captures) — which is why it is surfaced rather than fixed.
- **`--max-turns` is undocumented on the installed CLI while every dispatch passes it.** Not currently a fault — it works — but it is an undocumented dependency of the whole fleet on a vendor binary that changes weekly. A single assertion in `lint-prompts.sh` or the settings ship-gate (`claude --help` still parses `--max-turns`, or a one-turn probe) would turn a future silent breakage into a loud one at ship time.
- **The dispatch brief's "commit after EVERY experiment" instruction worked exactly as intended and is worth generalising.** Six independent units of work, six commits, six pushes — a cap death at any point would have left everything before it. **The property that made it safe is that each unit's output was self-contained** (a ruling that stands alone). Build-phase-shaped workflows over checklist docs could state that as a default rather than relying on the brief to say it each time.

Pumapumapumas:
# build-refine pass — Decision Log, Deferred Work, Reflection

Stages 1–5 all ran. Stage 2a dispatched code-reviewer, refactoring-evaluator and standards-auditor in parallel; Stage 2b ran quality-control sequentially with their findings.

## Decision Log

Ordered low-confidence first.

- **[Low]** **Re-scoped E7's predicate rather than declaring the draft ruling simply wrong and reverting to the original phase-doc text.** Alternatives: revert step 3 to its pre-draft state and let Phase 5 rediscover the open-subset distinction; or leave the draft's severity redirection with a warning note. Why: neither the draft's reading (set-difference is decorative) nor the original (empty delta over all ids) is correct, and the measurement supports a third thing — the mechanism with a different input set. **The residual risk is that I have now specified more of Phase 5 than a measurement phase should**, so I deliberately stopped at naming the input set and the empty-vs-unchanged distinction, both of which are *measured facts*, and left the predicate's construction to Phase 5. A reviewer could reasonably argue the "one archived positive observation" caveat should have been strong enough to leave the redirection out entirely.

- **[Low]** **Kept the superseded E7 ruling in a `<details>` block rather than deleting it.** Alternatives: delete it (the corrected ruling is what binds); move it to a commit message only. Why: the failure mode — reading a monotone-by-construction set's monotonicity as a fleet property — recurred *twice in this one PR* (E7's cumulative ledger, E5's self-observed log directory), which is evidence it is a repeatable trap rather than a one-off slip. The doc's own closing gotcha is about denominator honesty; this is the same class and had no entry. **A reviewer could argue this bloats a doc that is already 550 lines.**

- **[Medium]** **Treated `escalated` and unknown dispositions as OPEN rather than as closed.** Alternatives: `disposition == "hold"` (what the archive's own vocabulary suggested at first read, and what my initial measurement used). Why: the fail-safe direction is the cheap one — an unrecognised disposition emptying the open set would report convergence never observed. It changes no current number (all 195 archived findings carry a known disposition), and the archive already contains a sixth value (`escalated`, 2 occurrences) that a hold-only reading would have mis-scored. This is bounding what a *future* re-run can silently conclude.

- **[Medium]** **Rejected refactoring-evaluator's High-priority proposal to import `routing._VERDICT` instead of copying the regex.** Alternatives: import it, as `review_pr_helper.py` does for its own consumers. Why: **a replay tool must pin the predicate it replays.** Importing the live pattern would make a re-run in a year silently measure a *changed* rule against the *same* archived logs and report it as the same number — the tool's whole stated purpose is a comparable number over a growing denominator. The drift a copy risks is loud (a test asserts parity with the shipped ERE); the drift an import risks is silent. **I did fix the real defect the finding exposed** — the false "verbatim from routing.py:43" comment and the tuple-shaped output it caused. The reasoning is now in the module so the next reviewer does not re-raise it.

- **[Medium]** **Corrected E5's denominator to 50 while stating that the 0 holds under either accounting.** Alternatives: leave 51 with a footnote, since no ruling moves. Why: the doc's closing line is *"sample-size honesty is a requirement, not a courtesy,"* and a measurement citing its own unfinished run as an external data point is exactly what that sentence is against. The correction costs a paragraph and the alternative is a permanent record that reads as a fleet observation.

- **[High]** **Extracted `last_whole_match` rather than leaving the findall→finditer fix inline.** Alternatives: leave it inline as originally written (correct, but in `main()`, which no test reaches). Why: I attempted a mutation on the inline version and it could not fail a test — the fix had no guard. Extracting it made two negative controls possible, both of which fire. A fix nothing can falsify is the shape this repo's Testing Standard exists to reject, and I would have been shipping one.

- **[High]** **Corrected `docs/file_structure.txt:97,99` even though it is adjacent to, not caused by, this PR's changes.** Alternatives: surface it as pre-existing drift. Why: the PR edited lines 174–184 of the same file while leaving the two lines describing the component it just measured saying `PLANNED` and "Five experiments." That is self-description drift caused by this PR's own work, not pre-existing condition — the lines only became wrong because of it.

## Deferred Work

- **`run-claude.sh` has no timeout, so an auth failure hangs unboundedly** — Why deferred: this phase measures; fixing the incumbent runtime is out of scope, and the right sizing (a `timeout` wrapper vs. reading `system/api_retry` from the stream) is a design call. Tracked at: **nowhere — SURFACED WITH NO POINTER.** Verified by: `gh issue list --repo helloskyy-io/Claude-Dot-Files --state open --limit 30` → 13 open issues (#26, #28, #36, #38, #55, #57, #59, #60, #61, #62, #63, #64, #65); none concerns dispatch timeouts or auth retry. `grep -n "timeout" scripts/workflows/activities/run-claude.sh` → **no match**, confirming the gap is live. Not filing: it is squarely **Phase 4's** subject matter (Phase 4 owns what parents read from failure envelopes) and E1(e) is already written into this PR as its input. Raised naked so it is picked up rather than laundered.

- **The turn-cap rate may have climbed above the 0.9% the code comment reopens on** — Why deferred: the two denominators are different populations (443 runs vs 72 archived logs), so this is a signal to watch, not a measurement that settles it. Tracked at: **`scripts/workflows/activities/run-claude.sh:163-165`**, the existing comment's own reopen clause. Verified by: `sed -n '156,166p'` on the live file → the block reads *"Reopen only if the rate climbs under unattended/pooled operation, and even then the fix is louder signalling, not resume."* This **expands something that already exists** (placement question 2); E5's cross-check is recorded in the phase doc's incumbent-surfaces table beside the cite. No new issue.

- **`liveness_signal_measurement.md` §2.3's event vocabulary is missing `system/api_retry` and `system/permission_denied`** — Why deferred: research papers are critic-gated evidence and this phase must not edit the pool. Tracked at: the paper's own revalidation clause. Verified by: `sed -n '10,12p' docs/development/fleet-reliability/research/raw/liveness_signal_measurement.md` → `Last validated: 2026-08-07` / `Revalidate: high — 3 weeks`, i.e. **due 2026-08-28** — an unfired trigger with a date, so placement question 1 says not an issue. Both event shapes are recorded verbatim in this PR's phase doc for the sweep to pick up.

- **Nine `pr_review:` envelope fields need a cross-language schema owner** — Why deferred: it is `phase3_typed_exit_record.md` step 6's third open question and belongs to that phase. Tracked at: **`docs/development/memory-management-framework/phase3_typed_exit_record.md`**, step 6, third bullet. Verified by: `git show origin/main:docs/development/memory-management-framework/phase3_typed_exit_record.md | grep -n "Cross-language schema ownership"` → present on the default branch at the *"Rule between a single language-neutral schema artifact both trees load, and two declarations plus a cross-tree conformance test"* bullet. Unchanged by this PR; E6's field list is its input.

- **`sprint.md`'s MMF entry is stale** (`QUEUED, NEEDS PLANNING` / `Phase doc: not yet written`) — Why deferred: `sprint.md` is human-in-the-loop and binding; an autonomous dispatch may not edit it. Tracked at: **surfaced in the PR body for human-reviewed editing**, per the governance rule's own mechanism. Verified by: `sed -n '143,145p' docs/development/sprint.md` → the lines are live and read as quoted; `git diff main...HEAD --name-only | grep sprint` → no match, confirming this PR did not touch it.

### Findings NOT deferred — dispositions

- **`FENCE` could truncate on a nested backtick fence** (code-reviewer, Info) — **REJECTED.** Checked every archived block: none contains a nested fence, and the extractor's own docstring explains the deliberate regex-over-YAML-parser choice (hand-edited blocks a strict parser would drop). Changing the fence logic to defend an unobserved case risks the extraction that currently works on 14 of 14 blocks. Re-raise on the first block that triggers it.
- **`run-claude.sh` greps all `result` events; `read_log` keeps the last** (code-reviewer, Info) — **REJECTED as a defect, verified as a fact.** One archived log has two `result` events (`revision-major-20260410-115958`), and it is in the excluded/retired set, so no number moves. The "last match wins" framing in the doc describes the child gate correctly. Recording it here rather than in the doc because it changes nothing a reader would act on.
- **The "Amended by Phase 1 E-n" blockquote pattern may accumulate** (refactoring-evaluator, Low) — **REJECTED for now, with the trigger named.** At 3–4 callouts across 2 docs, each localised to the step it amends, the pattern is readable and carries the traceability it exists for. It needs folding into requirement text when a single step draws amendments from *two different phases*; that has not happened.
- **`docs/standards/documentation/README.md:36` might have been edited by this branch** (standards-auditor, Low, tool-limited) — **REJECTED, verified.** `git diff main...HEAD --name-only` returns ten files, none under `docs/standards/`. The auditor could not run git; the check it asked for comes back clean.

## Post-Run Reflection

**Friction**

- **The single highest-value finding in this pass came from an agent that could not verify it.** quality-control flagged, as an explicit *question not a finding*, that it could not tell whether `pr_review:` blocks are cumulative — and said so rather than guessing. One `gh pr view 58` settled it, and it inverted the PR's headline ruling. **The lesson is about the shape of the report, not the tool access:** an agent that writes "I could not check X, and here is what turns on it" hands the orchestrator a lead; one that quietly omits X hands it nothing. Worth stating explicitly in review-agent prompts that a named unverifiable is a first-class output.
- **Two of this PR's three wrong rulings came from measuring an artifact the measurement was inside of** — a log directory the measuring run was writing to, and a cumulative ledger read as if it could shrink. Both looked like clean zeros. Neither was falsifiable from within the draft's own framing, because in both cases the schema guaranteed the answer. **"What would this measurement look like if the property were false?" would have caught both**, and it is a cheaper check than re-measuring.
- **The brief's warning that a bash-shaped grep measures half the fleet earned its place twice** — once as the draft reported (E6's `plan_revision_workflow.py` issue-URL branch point), and once in the refine pass, where the *same* Python-only issue-URL path turned out to be missing from the replay tool's pattern set. The same fact was found from the parent's side and missed from the tool's side, in one PR.
- **`mutate.sh` rejected a multi-line mutation and was right to.** Its error explains that `grep -F` treats an embedded newline as alternation, so presence and ambiguity checks cannot be trusted. That is a well-placed guard; noting it because the failure looked at first like a harness limitation and is actually the harness refusing to certify something it cannot check.

- **The `pr_review:` block's cumulative shape is undocumented and load-bearing.** Nothing in `review-pr.sh` states that pass N restates every prior id — it falls out of the "reuse each prior finding's stable id slug verbatim" instruction at `:221`. Two separate consumers in this PR read the resulting monotonicity as a fleet property. One sentence in the Stage 5 emission block ("the block is cumulative: restate every prior finding with its disposition updated") would make the schema self-describing.
- **`disposition` is the fleet's most useful shipped signal and nothing reads it.** Six values across 195 archived findings, complete coverage, and it is what distinguishes an open finding from a closed one — yet the convergence discussion reached for a `severity` field that has never existed. Worth a line wherever the block's schema is described.
- **`.claude/logs/` grew twice during this one PR** (73 → 74). Every count taken over it dates within a day. The `N as of YYYY-MM-DD` convention this phase doc adopted is cheap and worth generalising.

*Flagging for the architecture session; not editing.*

- **`run-claude.sh` wraps `claude` in no timeout, so an auth failure hangs.** Measured: 9 `system/api_retry` events at `error_status: 401` against `max_retries: 10`, exiting only because the measurement harness supplied its own `timeout(1)`. Under unattended operation a credential expiry becomes a stuck dispatch with no upper bound. Two fix shapes worth weighing (a wall-clock `timeout` wrapper vs. reading `system/api_retry` from the stream the fleet already captures), which is why it is surfaced rather than fixed.
- **`--max-turns` is undocumented on the installed CLI while every dispatch passes it.** Not currently a fault, but an undocumented dependency of the whole fleet on a vendor binary that changes weekly. A single assertion in `lint-prompts.sh` or the settings ship-gate would turn a future silent breakage into a loud one at ship time.
- **A measurement tool run from a worktree resolves `<repo root>` to the worktree**, where `.claude/` does not exist because it is globally gitignored. Any future tool reading `.claude/logs/` inherits this. Worth a documented helper for "the main checkout's root, not this worktree's" rather than each tool re-deriving `parents[N]`.
- **The draft's per-experiment commit discipline worked exactly as intended and made this correction cheap.** Six self-contained units meant the refine pass could re-take E7 without disturbing E1–E3. **The property that made it work is that each unit's output stands alone** — worth stating as a default for build-phase-shaped workflows over checklist docs, rather than relying on the brief to say it each time.

Pumapumapumas:
# PR Review — disposition pass 3

**Verdict: HOLD — needs-assistance.** Two live defects (one in the authoritative roadmap, one in a kept measurement tool), one preventive comment fix, and two items that genuinely need the operator's pen (a standards surface and the sprint file).

**On the evidence base first, because it changes how the rest reads.** I re-ran both tools and re-derived every headline figure independently. `replay_pr_review_blocks.py` reproduces **14 blocks · 7 pairs · 195 findings · 1/14 blocks with an empty open set · 2/7 empty open deltas**, and the disposition census matches the code comment value for value (`hold` 37 · `fixed` 74 · `deferred` 58 · `rejected` 21 · `noted` 3 · `escalated` 2). Every line citation I spot-checked — `review-pr.sh:186`, `:221`, `build.sh:277`, `build-minor.sh:281`, `run-claude.sh:201-204`, `phase3:105`, `direction.md:67` — resolves exactly as written. **This PR's measurements hold up.** What did not propagate is the *correction*, and that is where both live defects sit.

---

## Findings

|---|---|---|---|
| `roadmap-carries-withdrawn-e7-ruling` | correctness | **HOLD** — redispatch | `roadmap.md:3` still reads *"Phase 5's empty-delta predicate was measured as never-firing (0 of 7 pass-pairs) and **replaced**"* — the exact ruling the refine pass reversed. `git log origin/main..HEAD -- roadmap.md` returns **only `f547108`** (the draft's E6 commit); the refine commit `c070927` swept `phase1`, `phase5` and `file_structure.txt` and never opened the roadmap. |
| `replay-tool-scores-patternless-workflows` | correctness | **HOLD** — redispatch | `replay_completion_predicate.py:146-154` buckets every unrecognised workflow into `pr_url`. `review-runs.sh` declares no `COMPLETION_PATTERN` at all (`run-claude.sh:201` gates on `[[ -n … ]]`, so it has no gate). I ran the tool: **all 3 `review-runs` logs score `strict_result: false`**, and with the 2 retired-workflow logs, **5 of the 8 reported strict-negatives are artifacts of the tool's own bucketing.** |
| `escalated-open-on-a-false-premise` | correctness | **HOLD** — redispatch | `replay_pr_review_blocks.py:63-66` asserts *"`hold` and `escalated` both leave work outstanding."* `review-pr.sh`'s own spec says the opposite: an escalated item is terminal and *"never contributes to a HOLD."* Verified number-neutral today (see below), so this is the rationale, not the number. |
| `exit-protocol-transport-marked-unmeasured` | standards-implication | **HOLD** — needs-assistance | `docs/standards/exit-protocol.md` landed on `main` as `d88e178` **after** this branch's base (`3e0522a`) — invisible from inside the worktree. Its §2 Transport row carries `⟨PHASE 1 · E1⟩` while its banner states *"a section still carrying `⟨PHASE N⟩` markers is a section whose answer has not been measured."* E1(g) answers it. |
| `sprint-mmf-entry-stale` | doc-drift | **HOLD** — needs-assistance | `sprint.md:143-145` reads `📋 QUEUED, NEEDS PLANNING` / *"Phase doc: not yet written — writing it is the planning step"* while a roadmap, five phase docs and one measured phase exist. Correctly surfaced by the producing run, not written. `sprint.md` is binding HiL — the operator's pen. |
| `run-claude-auth-hang` | correctness | **DEFERRED** | Re-homed. Producing run said *"nowhere — SURFACED WITH NO POINTER … squarely Phase 4's subject matter."* **That placement fails on Phase 4's body.** A valid home exists → `docs/development/sprint.md:176`. |
| `liveness-paper-missing-two-events` | doc-drift | **DEFERRED** | Verified §2.3 (L194-318) contains neither shape (the L159 hit is in **§2.2**, the *documented* surface, not §2.3's *observed* one). Trigger dated → `liveness_signal_measurement.md` revalidation, due **2026-08-28**. |
| `cross-language-schema-owner` | scope | **DEFERRED** | `phase3_typed_exit_record.md:105` — opened and confirmed present on the branch and on `main`. |
| `turn-cap-rate-signal` | friction | **NOTED** | Real, nothing broken. The cite's own reopen clause names the remedy as *"louder signalling, not resume"* — already shipped. |
| `fence-nested-backtick-truncation` | correctness | **REJECTED** (upheld) | Producing run's rejection stands. No archived block contains a nested fence; the regex-over-YAML choice is documented and extraction works on 14 of 14 blocks. Named re-raise trigger. |
| `run-claude-greps-all-result-events` | correctness | **REJECTED** (upheld) | Re-verified: `jq -r 'select(.type=="result") | .result // ""'` concatenates, `grep -qE` matches either — the child gate is "any result event carries it." The doc's "last match wins" describes the *parent*'s `tail -1`. No number moves. |
| `amended-by-blockquote-accumulation` | doc-drift | **REJECTED** (upheld) | 3-4 localised callouts across 2 docs, each attached to the step it amends. Trigger named (a step drawing amendments from two different phases); not reached. |
| `standards-readme-possibly-edited` | standards-implication | **REJECTED** (upheld) | Re-verified independently: `git diff main...HEAD --name-only` → ten files, none under `docs/standards/`. The auditor's tool-limited concern comes back clean. |

---

### Why `roadmap-carries-withdrawn-e7-ruling` is the headline

It is not doc tidiness. `docs/standards/exit-protocol.md` § Provenance — merged to `main` yesterday — states: ***"That roadmap is the authority until this document is ratified — where the two disagree today, the roadmap wins."***

So the roadmap is the **declared authority**, and its status line currently tells a Phase 5 engineer that the empty-delta predicate never fires and was replaced. This PR's own correction measured that the replacement target — `findings[].severity` — appears on **0 of 195** archived findings. The failure mode is precise: the engineer reads the authority, builds against a field that has never existed, and this PR's most expensive correction is spent.

The refine pass caught the error, corrected `phase1` (including a `<details>` block preserving the lesson), corrected `phase5`'s steps 1-4, and corrected `file_structure.txt:98`. It corrected every surface **it opened**. The roadmap was the one it did not.

### Why `replay-tool-scores-patternless-workflows` is not a nit

`scripts/helpers/measure/README.md:18-20` states the directory's own contract:

> *"every excluded artifact is named as excluded rather than quietly dropped from a denominator. **The exclusions are the part a reader cannot reconstruct.**"*

The tool does not honour it. The exclusion of `review-runs` / `revision` / `revision-major` lives in one paragraph of a **568-line** phase doc; the tool itself silently applies `pr_url` to all of them and reports **8** strict-negatives where the honest in-scope figure is **2**. The tool is kept *specifically* to be re-run for `D-007` over a growing denominator — a future operator re-runs it and reads 8.

This is the **identical defect class the refine pass already fixed once**: `plan-revision`/`plan-new` were being scored against a predicate they do not declare, and `PR_OR_ISSUE_URL` was added for exactly that reason. The sibling case — a workflow that declares *no* predicate — was left behind. There is also **zero test coverage on the bucket mapping**: neither test file imports `VERDICT_WORKFLOWS` / `ISSUE_ALTERNATIVE_WORKFLOWS` or exercises `main()`'s classification, which is why the first instance shipped and the second survived.

### Why `escalated-open-on-a-false-premise` is a comment fix and not a number fix

I tested it rather than reasoning about it. Re-ran the tool's corpus with `escalated` in and out of `CLOSED_DISPOSITIONS`:

| reading | blocks with empty open set | empty open-delta |
|---|---|---|
| `escalated` OPEN (shipped) | 1 / 14 | 2 / 7 |
| `escalated` CLOSED (per spec) | 1 / 14 | 2 / 7 |

Identical — both escalated findings sit in PR #56 pass 1, a block with other open findings. **No number moves, so nothing shipped here is wrong.** What is wrong is the stated reason, and `phase5_convergence_stopping.md` step 3 explicitly defers *"which of its six values count as closed"* to Phase 5 — which will read this tool and inherit the false premise. One comment, corrected now, costs nothing.

### Why `run-claude-auth-hang` is deferrable after all

The producing run refused to invent a pointer and said so plainly — that is the honest behaviour, and I am not scoring it as laundered. But its placement reasoning does not survive reading the bodies:

- **Phase 4 does not own it.** `phase4_fleet_migration.md`'s eight completion requirements are entirely about migrating routing onto the typed record. Nothing about the incumbent runtime failing to exit.
- **This component's own roadmap already ruled it out.** `roadmap.md:133`: *"The stall/liveness axis belongs to Fleet Reliability, not here."* An unbounded hang **is** the stall axis.
- **The home exists.** `sprint.md:176` — Fleet Reliability § ***"Three cheap guards — credential expiry**, false completion, and a safety-hook wiring test"*, and `:178`'s three-legged liveness predicate. Its research pool already carries `fleet-reliability/research/raw/credential_expiry_detection.md`.

So: DEFERRED to an existing sprint item, no issue filed (criterion 3 fails — already covered). Two independent absence checks confirmed no issue duplicates it (`gh issue list --state open` → 13 issues, none matching; `--state all` filtered on `timeout|hang|auth|retry|401` → only #44, unrelated). *Why not now:* the sizing call the producing run named — wall-clock `timeout` wrapper vs. reading `system/api_retry` from the stream — is precisely what that sprint item's planning step decides, and E1(e) is its input.

---

## WHAT HAPPENS NEXT

**1 — Redispatch (`build-minor.sh --pr 66`).** Three scoped corrections, all in files this PR already owns. No new experiments, no re-measurement.

**2 — Operator ruling: `docs/standards/exit-protocol.md`** *(needs-assistance — standards are HiL)*

> **reframe:** Not *"is the scaffold incomplete?"* — it is a scaffold and says so. The real question is: **does merging this PR make a statement in a standard on `main` false, and who is allowed to fix it?**
> **bp:** A doc that self-describes its own completeness state must not assert "unmeasured" about something measured — a stale completeness marker is worse than no marker, because readers act on it. The Exit Protocol invites phase engineers to write answers in; standards governance requires a human gate; the governance rule's own clarification says **the mechanism is a PR**.
> **recommendation:** Replace the §2 Transport row's `⟨PHASE 1 · E1⟩` marker with E1(g)'s ruling (`structured_output`, on isolation and Temporal replay cost, not availability alone) plus a link to `phase1_measure_the_channel.md`. Leave the Field-list row's `⟨PHASE 3⟩` marker alone — that one is genuinely still open. Your call whether this rides this PR as a standards edit (the PR gate satisfies HiL) or waits for Phase 3; I lean **now**, because the marker becomes false the moment this merges and nobody re-reads a scaffold.

**3 — Operator ruling: `sprint.md:143-145`** *(needs-assistance — sprint file is binding HiL, no autonomous edit)*

> **reframe:** Not *"is the sprint line wrong?"* — it plainly is. The question is **what a dispatch does when it reads a stale sprint line**, and the answer is that it re-plans a component that already has a roadmap, five phase docs and one measured phase.
> **bp:** The sprint file is the operator's cross-domain sequencing surface and every edit is human-reviewed. Surfacing the candidate rather than writing it is exactly correct, and the producing run did that.
> **recommendation:** Flip `📋 QUEUED, NEEDS PLANNING` → an in-progress status, and replace **"Phase doc:** not yet written — writing it is the planning step" with a pointer to `memory-management-framework/roadmap.md`, noting Phase 1 measured 2026-08-08. One-line edit, no redispatch, no re-review.

*Nothing in the runway requires re-running an experiment. Items 2 and 3 are pen strokes; item 1 is a scoped fix.*

---

```yaml
pr_review:
  pr: 66
  pass: 3
  attempt: 2
  verdict: HOLD
  converged: false
  findings:
    - id: roadmap-carries-withdrawn-e7-ruling
      title: Phase 5 gets built against a field that has never existed, because the roadmap — the declared authority — still carries the reversed E7 ruling
      category: correctness
      consequence: >-
        roadmap.md:3 states Phase 5's predicate was "measured as never-firing (0 of 7 pass-pairs)
        and replaced". The refine pass reversed exactly that: the predicate is RE-SCOPED to the open
        subset (2/7 empty, 1/14 blocks empty) and the replacement target it named — findings[].severity —
        appears on 0 of 195 archived findings. docs/standards/exit-protocol.md § Provenance (on main as
        of d88e178) declares "That roadmap is the authority until this document is ratified — where the
        two disagree today, the roadmap wins". A Phase 5 engineer reading the authority builds the
        withdrawn severity-based predicate, spending this PR's most expensive correction.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
      reviewed_sha: d88e178f1c63afc23bd416a4aa949b946d41d211
    - id: replay-tool-scores-patternless-workflows
      title: a re-run of the kept replay tool reports 8 completion misses where the honest in-scope figure is 2, because it scores workflows that declare no predicate
      category: correctness
      consequence: >-
        replay_completion_predicate.py:146-154 buckets any unrecognised workflow into pr_url.
        review-runs.sh declares no COMPLETION_PATTERN (run-claude.sh:201 gates on [[ -n … ]], so it has
        no gate at all), and revision/revision-major are retired. Running the tool scores all 3
        review-runs logs strict_result:false — 5 of its 8 reported strict-negatives are artifacts of
        its own bucketing. The tool is kept expressly to be re-run for direction row D-007 over a
        growing denominator, and scripts/helpers/measure/README.md:18-20 states its contract as
        "every excluded artifact is named as excluded … the exclusions are the part a reader cannot
        reconstruct". The exclusion exists only as prose in a 568-line phase doc, so a future re-run
        reads a miss rate 4x the real one. Same defect class the refine pass fixed for plan-revision
        via PR_OR_ISSUE_URL; the sibling case was left behind, and the bucket mapping has no test.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: escalated-open-on-a-false-premise
      title: Phase 5's "which dispositions count as closed" ruling inherits a false premise from the tool it is told to read
      category: correctness
      consequence: >-
        replay_pr_review_blocks.py:63-66 asserts "hold and escalated both leave work outstanding".
        review-pr.sh's own spec states an escalated item is terminal and "never contributes to a HOLD".
        phase5_convergence_stopping.md step 3 explicitly defers "which of its six values count as
        closed" to Phase 5, which reads this tool. If escalated is ruled open on this premise, a review
        whose only surviving findings are escalated is scored non-converged and the loop runs to its
        bound. Verified number-neutral today — re-ran the corpus with escalated in and out of
        CLOSED_DISPOSITIONS and got 1/14 blocks and 2/7 deltas either way — so the defect is the
        recorded reason, not the shipped number.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: exit-protocol-transport-marked-unmeasured
      title: the Exit Protocol tells a Phase 3 engineer the transport question is unmeasured, so a settled experiment gets re-run
      consequence: >-
        docs/standards/exit-protocol.md landed on main as d88e178, AFTER this branch's base 3e0522a,
        so it was invisible from inside the worktree. Its §2 envelope table carries "Transport |
        ⟨PHASE 1 · E1⟩ — judged on isolation and replay cost, not just availability", while its status
        banner states "A section still carrying ⟨PHASE N⟩ markers is a section whose answer has not
        been measured". E1(g) answers it (structured_output wins on availability, isolation and
        Temporal replay cost). Merging this PR is what makes the banner's rule false for that row, so
        the item is this PR's; a Phase 3 engineer orienting from the protocol re-opens a transport
        ruling that cost forced-failure API runs to establish.
      disposition: hold
      hold_kind: needs-assistance
      reviewed_sha: d88e178f1c63afc23bd416a4aa949b946d41d211
    - id: sprint-mmf-entry-stale
      title: a future dispatch re-plans a component that already has a roadmap, five phase docs and one measured phase
      category: doc-drift
      consequence: >-
        sprint.md:143-145 reads "📋 QUEUED, NEEDS PLANNING" and "Phase doc: not yet written — writing
        it is the planning step". The roadmap and five phase docs landed in #54 (3b3fb0c) and Phase 1
        is now measured. The sprint file is the operator's cross-domain sequencing surface; a dispatch
        or standup reading it concludes the planning step is outstanding and re-plans finished work.
        The producing run surfaced this correctly rather than editing the file, per binding governance.
      disposition: hold
      remedy: operator-action
      hold_kind: needs-assistance
    - id: run-claude-auth-hang
      title: an expired credential turns any unattended dispatch into an unbounded hang
      category: correctness
      consequence: >-
        E1(e) measured 9 system/api_retry events at error_status 401 against max_retries 10, exiting
        only because the measurement harness supplied its own timeout(1). grep -n timeout
        scripts/workflows/activities/run-claude.sh returns no match. Under unattended or pooled
        operation a credential expiry becomes a stuck dispatch with no upper bound.
      disposition: deferred
      pointer: "docs/development/sprint.md:176 — Fleet Reliability § 'Three cheap guards — credential expiry, false completion, and a safety-hook wiring test' (also :178, the three-legged liveness predicate)"
      pointer_verified: true
    - id: liveness-paper-missing-two-events
      title: Fleet Reliability's liveness design would be built from an observed-event vocabulary missing the two shapes that signal a credential stall
      category: doc-drift
      consequence: >-
        liveness_signal_measurement.md §2.3 (the OBSERVED event surface, lines 194-318) lists neither
        system/api_retry nor system/permission_denied; the line-159 occurrence is in §2.2, the
        DOCUMENTED surface, which is a different claim. api_retry is present in 14 archived logs and
        non-empty permission_denials arrays are present in archived logs today, so §2.3's own stated
        method (glob plus ripgrep over .claude/logs/*.jsonl) re-run at revalidation surfaces both.
        Not material to any ruling in this PR — E1 measured both shapes directly rather than relying
        on the paper — so it rides the sweep rather than holding this PR.
      disposition: deferred
      pointer: "docs/development/fleet-reliability/research/raw/liveness_signal_measurement.md — 'Last validated: 2026-08-07 / Revalidate: high — 3 weeks', due 2026-08-28; trigger-gated with a date, so per placement question 1 it is not an issue"
      pointer_verified: true
    - id: cross-language-schema-owner
      title: the bash and Python envelopes diverge with every test in both trees green, unless one owner is named
      category: scope
      consequence: >-
        E6's nine-field envelope must be declared once across two fleets. An intra-tree test cannot
        catch a cross-tree divergence — the same duplicated-vocabulary defect that produced routing.py,
        one level up. Ruling it is Phase 3's, not Phase 1's; E6's field list is its input.
      disposition: deferred
      pointer: "docs/development/memory-management-framework/phase3_typed_exit_record.md:105 — step 6, 'Cross-language schema ownership' bullet; confirmed present on the branch and on origin/main"
      pointer_verified: true
    - id: turn-cap-rate-signal
      title: nothing is decided wrongly by the turn-cap rate flag, because the remedy its own cite names is already shipped
      category: friction
      consequence: >-
        No consequence, which is why this is noted rather than deferred. E5 observed 2 error_max_turns
        in 72 completed logs (2.8%) and 2 in 50 in-scope (4.0%) against run-claude.sh:159's recorded
        0.9% (4/443), but the denominators are different populations so no rate claim is established.
      disposition: noted
      remedy: none
      no_live_defect_check: >-
        Read run-claude.sh:156-165 on the branch. The comment's reopen clause reads "Reopen only if the
        rate climbs under unattended/pooled operation, and even then the fix is louder signalling, not
        resume" — and the louder signalling (the turn-cap banner at :167-190) is already shipped and
        firing. Separately confirmed issue #65 is open on the banner's false 'NOTHING was committed'
        assertion, not on the rate, so it neither covers nor contradicts this.
    - id: fence-nested-backtick-truncation
      title: FENCE would truncate a pr_review block containing a nested code fence, losing findings from the corpus
      category: correctness
      consequence: >-
        Would silently shorten an extracted block and drop findings from every count. Rejection upheld:
        no archived block contains a nested fence, extraction works on 14 of 14, and the regex-over-YAML
        choice is documented as deliberate (hand-edited blocks a strict parser would drop). Changing the
        fence logic to defend an unobserved case risks the extraction that currently works. Re-raise
        trigger is named and correct — the first block that triggers it.
      disposition: rejected
      remedy: reject
    - id: run-claude-greps-all-result-events
      title: a log with two result events would let the child gate pass on the wrong one
      category: correctness
      consequence: >-
        Would mean a run passing its completion gate on a stale result. Rejection upheld and
        independently re-verified: run-claude.sh:203 uses jq -r 'select(.type=="result") | .result //
        ""', which concatenates every result event, and grep -qE then matches if ANY carries the
        pattern — the child gate's semantics are "any result event carries it", which is correct. The
        doc's "last match wins" phrasing describes the PARENT's tail -1 at build.sh:277. The one
        archived two-result log is revision-major-20260410-115958, in the retired/excluded set, so no
        number moves.
      disposition: rejected
      remedy: reject
    - id: amended-by-blockquote-accumulation
      title: repeated "Amended by Phase 1 E-n" callouts would bury the requirement text a phase engineer executes against
      category: doc-drift
      consequence: >-
        Would make a checklist unreadable at the point of use. Rejection upheld: 3-4 callouts across 2
        docs, each localised to the step it amends, still readable and carrying the traceability they
        exist for. The named folding trigger — a single step drawing amendments from TWO different
        phases — has not occurred, and it is the right trigger.
      disposition: rejected
      remedy: reject
    - id: standards-readme-possibly-edited
      title: an unnoticed edit to a vendored standard would break the MIRROR drift gate
      consequence: >-
        Would fail vendor-standards.sh --check and put an unreviewable local edit into a vendored
        corpus. Rejection upheld and independently re-verified rather than taken on report: git diff
        main...HEAD --name-only returns ten files, none under docs/standards/. The standards-auditor
        raised it explicitly as tool-limited (it could not run git), which is the correct shape for an
        unverifiable — the check it asked for comes back clean.
      disposition: rejected
      remedy: reject
  next_steps:
    - item: roadmap-carries-withdrawn-e7-ruling
      kind: redispatch
      note: Propagate the refine pass's corrected E7 ruling into roadmap.md, the one surface it never opened.
      dispatch_tool: build-minor.sh
      dispatch_context: |
        Fix three scoped items on branch build-draft/mmf-phase1-measure-channel. No new
        measurements, no re-running experiments, no re-adjudication of any ruling.

        (1) docs/development/memory-management-framework/roadmap.md:3 — the Status line states
            the WITHDRAWN E7 ruling as Phase 1's headline: "Phase 5's empty-delta predicate was
            measured as never-firing (0 of 7 pass-pairs) and replaced." The shipped ruling is the
            opposite. Replace it with the corrected one, which is already written verbatim at
            phase1_measure_the_channel.md:551 and :316 — the predicate is RE-SCOPED to the OPEN
            subset, not replaced: empty 2 of 7 pairs, the open set empties completely on 1 of 14
            blocks (PR #42 pass 2, the archive's only MERGE and only converged: true), and
            `disposition` — not `severity` — is its input, because severity appears on 0 of 195
            archived findings.
        (2) Same file, line 81 (the Phase 5 blurb): the clause "did this pass find anything not in
            the previous pass's result?" is the framing phase5_convergence_stopping.md step 1 now
            amends to "is anything still open?". Add the open-subset qualifier so the roadmap and
            the phase doc state the same question. Do NOT touch the checkboxes at :82-85 — they are
            written at planning altitude and remain correct.
        (3) scripts/helpers/measure/replay_completion_predicate.py — add a fourth classification for
            workflows with no live COMPLETION_PATTERN declaration, so the tool stops asserting misses
            against a predicate that was never in force. Concretely: a NO_PATTERN_WORKFLOWS set
            covering `review-runs` (exists, declares no pattern — verified: run-claude.sh:201 gates
            on [[ -n "${COMPLETION_PATTERN:-}" ]], so it has no gate) and the retired `revision` /
            `revision-major` (no such script in scripts/workflows/). Give those rows
            pattern_kind "none", leave strict_*/loose_* null rather than false for them, and print
            them under their own header in main() the same way the in-flight list is printed. This
            is the SAME defect the refine pass fixed for plan-revision via PR_OR_ISSUE_URL — the
            sibling case. Add a unit test on the bucket mapping itself (it currently has ZERO
            coverage: neither test file imports VERDICT_WORKFLOWS or ISSUE_ALTERNATIVE_WORKFLOWS),
            with mutation evidence per the Testing Standard. Update the E5 table in
            phase1_measure_the_channel.md ONLY if a number changes — it should not: the doc already
            excludes these 22 logs in prose and reports the corrected 50 / 2 figures.
        (4) scripts/helpers/measure/replay_pr_review_blocks.py:63-66 — the comment asserts "hold and
            escalated both leave work outstanding". review-pr.sh's spec says an escalated item is
            TERMINAL and "never contributes to a HOLD". Correct the comment: escalated is terminal
            per the spec; it is counted OPEN here as a deliberate fail-safe pending Phase 5's ruling
            on which of the six values count as closed (phase5 step 3 owns that ruling); and the
            choice is verified number-neutral — 1/14 blocks and 2/7 deltas either way. Do NOT change
            CLOSED_DISPOSITIONS or any test — the shipped numbers are correct.

        DO NOT TOUCH: docs/development/sprint.md (binding human-in-the-loop), docs/standards/
        (human-in-the-loop; the exit-protocol.md item is a separate operator ruling), any E1/E2/E3/E5/E6
        ruling, the <details> superseded-ruling block at phase1:331-342 (it is deliberately preserved),
        or memory-management-framework/research/.
      precheck: |
        RUN FROM THE PR BRANCH (build-draft/mmf-phase1-measure-channel), not main — items 1 and 2
        live in this unmerged PR and a check run from main reports them absent. All paths below are
        relative to the branch worktree root.

        PRECEDENCE: the dispatch_context enumeration above GOVERNS. This precheck gates whether to
        act; it never widens or narrows what to act on. If a check below disagrees with the
        enumeration, follow the enumeration and report the disagreement.

        Item 1+2 (different check than the grep that surfaced it — verify against the SHIPPED ruling
        rather than searching for the withdrawn wording):
          diff <(grep -o 'RE-SCOPED[^.]*' docs/development/memory-management-framework/phase1_measure_the_channel.md | head -1) \
               <(grep -o 'RE-SCOPED[^.]*' docs/development/memory-management-framework/roadmap.md | head -1)
          - roadmap side EMPTY  -> proceed; the corrected ruling has not reached the roadmap.
          - roadmap side NON-EMPTY and matching -> STOP for item 1, it is ALREADY DONE.
          - phase1 side empty   -> STOP AND REPORT: the finding's own premise is broken, do not
                                   invent a correction.

        Item 3 (different check — assert against the tool's OUTPUT, not its source):
          python3 scripts/helpers/measure/replay_completion_predicate.py <main-checkout>/.claude/logs \
            | python3 -c 'import json,sys; d=sys.stdin.read(); rows=json.loads(d[d.index("["):]); print(sorted({(r["workflow"], r["pattern"]) for r in rows if r["workflow"] in ("review-runs","revision","revision-major")}))'
          - any of the three still paired with "pr_url" -> proceed.
          - all three paired with "none" -> STOP for item 3, it is ALREADY DONE.
          - PASS THE LOG DIR EXPLICITLY. The default resolves to the worktree root, where .claude/
            does not exist (globally gitignored), and the tool then reports an empty corpus. An
            empty corpus is NOT "already done" — it is a mis-invocation; re-run with the main
            checkout's path.
          - NOTE these are two DIFFERENT stop states and only the second is done-ness. A corpus that
            happens to contain no review-runs logs is "not yet observable", NOT "already fixed":
            proceed with the fix anyway, because the classification is what is being added, not the
            count.

        Item 4:
          grep -c 'terminal' scripts/helpers/measure/replay_pr_review_blocks.py
          - 0 -> proceed. - non-zero -> read the comment; if it already states escalated is terminal
            per the spec, STOP for item 4, it is ALREADY DONE.

        Regression gate for all items: ./testing/run-all.sh must pass, and
        `python3 scripts/helpers/measure/replay_pr_review_blocks.py` must still report 14 blocks,
        7 pairs, 195 findings, 1/14 empty open sets and 2/7 empty open deltas. Any movement in those
        five figures means the fix changed a measurement and is wrong.
    - item: exit-protocol-transport-marked-unmeasured
      kind: needs-assistance
      why_human: standards
      reframe: >-
        Not "is the scaffold incomplete" — it is a scaffold and says so. The question is whether
        merging this PR makes a statement in a standard on main FALSE, and who is allowed to fix it.
      bp: >-
        A document that self-describes its own completeness state must not assert "unmeasured" about
        something measured; a stale completeness marker is worse than no marker because readers act
        on it. The Exit Protocol invites phase engineers to write answers in, standards governance
        requires a human gate, and that rule's own clarification names the PR as the mechanism.
      recommendation: |
        Replace the §2 envelope table's Transport row marker ⟨PHASE 1 · E1⟩ with E1(g)'s ruling —
        structured_output, decided on isolation and Temporal replay cost rather than availability
        alone — plus a link to phase1_measure_the_channel.md. Leave the Field-list row's ⟨PHASE 3⟩
        marker alone; that one is genuinely still open, and §3's precedence row is likewise Phase 3's.
        My recommendation is to do it NOW rather than at Phase 3: the marker becomes false the moment
        this PR merges, and nobody re-reads a scaffold. A PR against docs/standards/exit-protocol.md
        satisfies human-in-the-loop; a self-merge would not.
    - item: sprint-mmf-entry-stale
      kind: needs-assistance
      why_human: sprints
      reframe: >-
        Not "is the sprint line wrong" — it plainly is. The question is what a dispatch DOES when it
        reads a stale sprint line, and the answer is that it re-plans a component that already has a
        roadmap, five phase docs and a measured phase.
      bp: >-
        The sprint file is the operator's cross-domain sequencing surface and every edit is
        human-reviewed; a dispatch surfaces the candidate and never writes it. The producing run did
        exactly that, which is why this is a pen stroke and not a finding against the run.
      recommendation: |
        Flip sprint.md:143's "📋 QUEUED, NEEDS PLANNING" to an in-progress status, and replace
        :145's "Phase doc: not yet written — writing it is the planning step" with a pointer to
        memory-management-framework/roadmap.md, noting Phase 1 measured 2026-08-08. One-line edit,
        no redispatch, no re-review needed.
  laundered_deferrals:
    caught: 0
    of_total: 5
  homeless_items: 0
  redispatched: false
```

---

## Post-Run Reflection

**Friction**

- **The pass counter and the review history disagree, and the prompt has no branch for it.** This is dispatched as pass 3, but `gh api …/issues/66/comments` returns **2 comments and neither contains a `pr_review:` block** (the 5 `pr_review:` string matches are the phase doc quoting the schema). So there were no prior findings to carry forward and every id here is new — which means the invariants that exist specifically for pass ≥2 (carry vanished items forward, detect re-laundering, verify my own prior prescriptions) had nothing to operate on. I seeded `attempt: 2` from the two attesting reflection comments per the pass-1 rule. Worth a line in the prompt: *if the pass counter is >1 but no prior block exists, seed as pass 1 and say so* — otherwise a future reader of this block sees `pass: 3` with all-new ids and reasonably concludes the earlier passes' findings were dropped.
- **The two-directional current-tree check earned its place, and the direction that paid was the one the prompt describes second.** `docs/standards/exit-protocol.md` landed on `main` ~5 minutes before the refine pass posted its comment, against a worktree cut 75 minutes earlier. Nothing in the PR could have referenced it, `git` reports the merge CLEAN because the textual overlap is one line in `file_structure.txt`, and it is the file that establishes the roadmap as the *declared authority* — which is what turns finding 1 from doc tidiness into a build-the-wrong-thing risk. **The arriving-rules direction found the thing that made another finding matter, not a finding of its own.** That is a different value than "catches non-conformance to a rule the author could not read," and the prompt only names the latter.
- **The deleted-artifact sweep has no sibling for a *reversed ruling*, and that is what this PR needed.** The sweep asks "this file was deleted — where did its assertions go?" Nothing asks "this ruling was withdrawn — which surfaces still assert it?" The refine pass corrected every surface it opened (`phase1`, `phase5`, `file_structure.txt`) and missed the one it did not. A one-line enumeration discipline — *list every file that states the ruling, not every file you edited* — would have caught it, and it is the same shape as the deleted-artifact rule: **enumerate by name, map each to a destination.** Worth adding as a second trigger on that mandatory sweep.
- **Running the tools was worth more than reading them.** Re-executing `replay_pr_review_blocks.py` reproduced all five headline figures and the full disposition census, and re-executing `replay_completion_predicate.py` is the only reason the `review-runs` bucketing surfaced at all — it is invisible in the diff, where it reads as a correct three-way `if/elif/else`. The `escalated` question likewise resolved from a 12-line script rather than an argument. When a PR ships an executable measurement, the executable is the artifact to check.

**Notes (below the finding bar — no consequence, recorded so they are not re-derived)**

- `phase1_measure_the_channel.md:~177` says strict is *"verbatim from the six PR-URL declarations"* while the table below it reads *"PR-URL predicate (9 workflows)"* and the tool comment says *"16 pull-only declarations"*. Three counts of three different things in one section; no ruling turns on any of them.
- PR body reports `python.sh unit helpers → 52 passed`; it is 53 on the branch today. Almost certainly a count taken before the final commit.
- `replay_completion_predicate.py`'s reconstructed "console" is assistant text turns only, while the real console also carries the `.result` text `run-claude.sh` echoes. This makes the reconstruction *narrower* than reality, so it could only manufacture false disagreement — and the observed result is agreement 50 of 50, which the doc already caveats as luck-of-the-corpus. Nothing to fix.

**Tooling-level suggestions (claude-dot-files)** — *flagging for the architecture session; not editing.*

- The producing run's own reflection asks for one sentence in `review-pr.sh` Stage 5 stating that the `pr_review:` block is **cumulative**. I second it from the other side: I authored this block and the instruction that makes it cumulative (`:221`, "reuse each prior finding's stable id slug verbatim") never says so in the emission block. Two consumers in this PR read the resulting monotonicity as a fleet property and one of them inverted a headline ruling.

PR #84, opened 2026-08-12
research-minor: cross-node memory prior art (persistent-memory-protocol)

Pumapumapumas:
## research-minor: cross-node memory prior art

**This is a MINOR cycle — one paper, no synthesis, no topic list.**

**Question:** How do mature systems carry durable memory across nodes that come and go? Evaluated against the operator's thesis — "workers are ephemeral like containers; what matters is stashed on persistent storage; short-lived processes are stitched together by a common set of data passed on" — as evidence to test, not a premise to confirm.

**Feeds:** a not-yet-taken planning decision on whether cross-node memory persistence becomes its own component or a phase of the Memory Management Framework. This paper is the validating evidence for that decision; per its own §5.1 it explicitly does not take it — no source found bears on component-vs-phase, which is a preference/scope ruling, not a research question.

**Destination context:** `docs/development/persistent-memory-protocol/` did not exist before this PR — this is a new component. It extends `docs/guide/memory-model.md`, which documents this fleet's single-machine Kind 1 model and explicitly has no answer for a second node (everything currently joins through one run log on one filesystem).

**Paper:** `docs/development/persistent-memory-protocol/research/raw/cross_node_memory_protocol.md` (901 lines)

**Sources:** 28 (23 external, 5 local) — above the proportional floor for a single-concern topic; the paper states why (§6): four load-bearing facets in one artifact, each needing 4–6 primaries, plus the peer-reviewed CRDT/Dynamo/local-first lineage for facet 4. Raw sources fetched via `curl` throughout rather than rendered pages, so quoted spans are exact returned bytes.

**What it covers — four facets of one question, per the dispatch's explicit framing:**
1. Identity across nodes (CloudEvents, W3C Trace Context/OpenTelemetry, Temporal workflow-id/run-id, content-addressing) — load-bearing, answered well enough for a later design to start from.
2. What lives at the edge vs. the centre, and what happens to work in flight when the centre is unreachable.
3. Binding-independence — what a store must be to serve as the durable record, checked against a non-git edge case (Home Assistant / SQLite / MQTT retained topic).
4. Reconciliation on return — LWW, vector clocks, CRDTs, OT, manual-merge-with-conflict-surfacing, and specifically what changes when writes carry reasoning rather than just values.

**What it explicitly does not cover** (§5, honest-boundary + test plan):
- The component-vs-phase decision itself (operator ruling, not research — §5.1, test-plan item 8).
[...]

PR #82, opened 2026-08-11
build-draft: Phase 6 — three readers for three unread observables, and the counter that could never move

Pumapumapumas:
## Summary

Builds **Phase 6 of the Memory Management Framework — *Read what it writes*.**

Three phases each appended a parent-written observable to the same per-run JSONL log under `.claude/logs/` — **`parent_route`** ([Phase 3](docs/development/memory-management-framework/phase3_typed_exit_record.md)), **`convergence`** ([Phase 5](docs/development/memory-management-framework/phase5_convergence_stopping.md)) and **`run_resources`** (Phase 4's telemetry, `0feacc3`) — and **no committed tool read any of the three.** Verified at the start of the phase: `grep -rln 'parent_route\|run_resources\|"convergence"' scripts/helpers/measure/` returned nothing. That is [`memory-model.md` §1.2](docs/guide/memory-model.md)'s own rule failing — *a record with no reader is not memory, it is exhaust* — in the shape this repo's `loose_ends.md` already ran for ten weeks.

| Piece | Status |
|---|---|
| `scripts/helpers/measure/run_log.py` | **NEW.** The surface's ONE declaration — member event types, join key, publish classification, per-figure commit cutovers, a structural sub-agent recount. Not a replay tool, and named as not-one in the README and the file map |
| `replay_run_resources.py` | **NEW.** Four figures: the `measured: false` rate, `peak_anon` by workflow, the subagent-count-vs-content-volume knob question, and the overlapping-run aggregate |
| `replay_parent_route.py` | **NEW.** The computed abstention arm grouped by reason, and the prose-shadow agreement rate with **C-060**'s conditioning printed beside it |
| `replay_convergence_events.py` | **NEW.** Candidate **C-059** built — the LIVE convergence corpus, plus Phase 5's activation conditions 1 and 2 with their denominators |
| Emission-side fixes | `run_claude` gains a **required** `workflow_key`; `run_id` now carries the bare nonce so the surface's join key actually joins; the sub-agent counter matched `Task` while this CLI spawns `Agent` |
| `sprint.md` | **UNTOUCHED** — operator-only |

### Three payload defects the readers exposed, one of which the plan did not know about

[...]

PR #86, opened 2026-08-12
research-minor: state-passing between claude -p children (persistent-memory-protocol)

Pumapumapumas:
## MINOR research cycle — one paper, no synthesis

**Question:** How does a workflow pass what it knows between `claude -p` child processes that share no memory — on ONE machine, within a single local dispatch?

**Destination:** A not-yet-taken decision on whether cross-`claude -p` state-passing becomes a phase of the Memory Management Framework or its own component. This paper is validating evidence for that decision; it does not take it (§5.1 states why no source could).

**Paper:** `docs/development/persistent-memory-protocol/research/raw/state_passing_between_workflow_children.md`

**Sources:** 18 external ([S1]–[S18]), under the 20-source ceiling. 7 are Temporal's own docs (read as this fleet's substrate, not an alternative); excluding that set leaves 11 across the four facets, inside the Research Standard's "proportionally fewer for small topics" target. Plus 14 local repository artifacts enumerated separately as the object of study. `Critic: not-yet-verified — 2026-08-12`, as the standard requires — a separate fresh-context run verifies it.

**What it explicitly does not cover:** multi-machine anything (identity schemes, placement, CAP, CRDTs/reconciliation), per-machine/per-edge stores, or a design/schema for this fleet. It also does not rule on the phase-vs-component question — that's named as categorically outside what research can settle (§5.1).

### The prior paper is superseded, not deleted

`cross_node_memory_protocol.md` answered a different (multi-machine) question because its brief restated the operator's "short-lived processes" framing as "nodes that come and go." It now carries a header noting it's superseded, why, and which three findings carried forward (re-grounded in primary evidence in the new paper rather than inherited). Nothing else in its body was touched. `docs/file_structure.txt`'s one-line description of the `persistent-memory-protocol/` research pool was corrected to match.

### Key findings (see the paper for full evidence)

- A hard prompt-size ceiling nobody had named: the prompt is a single `execve` argument, capped by Linux `MAX_ARG_STRLEN` at 131,072 bytes; the `review-pr` research-type template already consumes 76,226 bytes (58%) of it before the unbounded `CONTEXT_BLOCK` is substituted in.
- The operator's consolidation question ("one location or many?") is answered against its implied direction: mature systems run multiple channels deliberately, but each has a single join key spanning them. This fleet's `run_id` join key is complete on exactly one child type and one field wide elsewhere — consolidate the *address*, not the *store*.
- The sharpest fleet consequence: `.claude/logs/` (the run-log JSONL surface) is 175 files / 262 MB / 125-day span with no pruning code and is gitignored, so it's neither pruned, nor durable, nor (until recently) read. This finding is explicitly stated as surviving the strongest counter-case against the paper's own thesis.
- Of the 8 boundary-crossing channels this fleet's code actually uses (enumerated from `research_workflow.py`, `build_workflow.py`, `run_log.py`, `exit_record.py`), 5 are covered by neither the Kind-1 nor the Kind-2 taxonomy in `memory-model.md` — recorded as unsettled evidence rather than a ruling on whether the classification cuts in the right place.


Pumapumapumas:
## Decision Log

- **[Low]** Left `cross_node_memory_protocol.md`'s machine-parseable header block (`Revalidate: high — 6 weeks`, etc.) untouched rather than converting it to the standard's `Revalidate: retired — <N> months` + `Superseded by:` form. Alternatives: fully apply Research Standard §3's superseded-artifact rule (change the header fields too). Why: the dispatch instruction was explicit and narrower than the standard — "Add a header... Do not otherwise edit its body" — so I added a prose note under the title recording the supersession and left the header block alone. Net effect: the paper-currency table will keep reporting this paper as "current" on its stale multi-machine content until a future pass reconciles the header to the standard's retired-artifact form. Flagging this discrepancy for the architecture session rather than resolving it myself, since amending header-block semantics is a standards-conformance question, not a research-content one.
- **[Medium]** Updated `docs/file_structure.txt`'s one-line description of `persistent-memory-protocol/` (it described the superseded "nodes that come and go" framing). Alternatives: leave it, since the task didn't explicitly ask for it. Why: proactive-doc-management applies — the description was actively misleading after this PR, and it's a one-line mechanical correction inside the file the standard already requires be kept current.
- **[Medium]** The research-analyst wrote its first draft of the new paper to the main checkout's working tree (`/home/puma/Repos/claude-dot-files/docs/...`) instead of this worktree, because the dispatch prompt (and my own agent prompt, copying it) gave an absolute path that happened to coincide with the main checkout's path rather than the worktree-rooted equivalent. Caught before commit: `ls` on the worktree path came back empty, traced the file to the main checkout via `find`, confirmed it was untracked and uncommitted there, copied it into the correct worktree path, and removed the stray copy from the main checkout so it wasn't left as untracked noise on `main`. No content was lost or duplicated; verified via `git status --porcelain` on both trees post-fix.

## Deferred Work

- **Reconcile `cross_node_memory_protocol.md`'s header block to the Research Standard's superseded-artifact form** (`Revalidate: retired — <N> months` + a `Superseded by:` line under the header block itself, per §3) — Why deferred: this dispatch's explicit instruction was "add a header... do not otherwise edit its body," which I read as scoped to the prose note I added, not the machine-parsed header fields; changing those is a standards-conformance edit outside a research-minor cycle's write boundary (raw/ content only) and better done deliberately. Tracked at: surfaced here in this PR body/comment — no existing tracker covers it. Verified by: `grep -n "^Revalidate:" docs/development/persistent-memory-protocol/research/raw/cross_node_memory_protocol.md` on this branch → still reads `high — 6 weeks`, confirming the header was not touched.

## Post-Run Reflection

**Friction:**
- The workflow dispatch's "Research dir:" line and the code-surface paths in the brief (`scripts/helpers/measure/run_log.py`, `modules/assistant/exit_record.py`, `modules/assistant/research/research/research_workflow.py`) were given as repo-root-relative paths without the worktree prefix, and two of the three didn't exist at the literal path given (they're actually under `scripts/workflows/temporal/modules/assistant/...`) — had to `find` them before dispatching the analyst. More consequentially, when I built the analyst's own prompt I passed the *absolute* main-checkout path for the output file rather than translating it to the worktree-rooted absolute path, and the analyst wrote there faithfully. This is exactly the trap `terminal-output.md`/the workflow's own "Large-file reading" rule warns about for `Write` calls, but the rule as stated only says "use an ABSOLUTE worktree-rooted path" — it doesn't call out that a dispatch's stated absolute path may itself be worktree-unaware and need translation before being handed to a sub-agent.

**Tooling-level suggestions (claude-dot-files):** This is a claude-dot-files-level change worth bringing to the architecture session: the research-minor (and likely research/research-family) workflow prompt states output paths as bare absolute paths (e.g. `/home/puma/Repos/claude-dot-files/docs/...`) without stating whether that path is worktree-rooted or main-checkout-rooted. A one-line addition — "this path is relative to your worktree root; verify with `pwd` or by comparing against your worktree's absolute path before using it in a sub-agent dispatch" — would have caught this before dispatch instead of after, and it's a generic hazard for any workflow that forwards a dispatch-supplied path straight into a subagent prompt.

Pumapumapumas:
# research-verify — PR #86

**Result: both papers verified and fit to merge.** Minor cycle, so Stages 2 and 3 emit `SKIPPED — minor cycle, no synthesis exists`; confirmed by `find docs/development/persistent-memory-protocol -type f`, which returns the two papers in `raw/` and nothing else. The paper is the deliverable.

| Paper | Final verdict | Rounds |
|---|---|---|
| `state_passing_between_workflow_children.md` | **PASS-WITH-FIXES**, closed at round 3 of a 3-round ceiling, 0 blocking surviving | 3 |
| `cross_node_memory_protocol.md` (inserted supersession block only) | **PASS-WITH-FIXES**, 1 blocking fixed; body carries its recorded `Critic: PASS` from PR #84 and is byte-unchanged | 1 |

## What was verified, and how

Round 1 was the full pass on the new paper: **all 18 external sources resolve, and every quoted span was confirmed byte-exact by raw retrieval** — `git clone --depth 1 --filter=blob:none --sparse` plus `grep -F` for the eight git-hosted source repos, and `curl` plus `grep -F` for the three sources that are not git-hosted or are impractical to clone ([S13] `docs.claude.com`, [S15] `sqlite.org`, [S17] `torvalds/linux`). **No WebFetch established any quote.** No fabricated source, no miscited quote, no confidence inflation — in any round.

`cross_node_memory_protocol.md` was scoped to the 20-line supersession block this PR inserted. Its body was **not** re-verified: `git diff main...HEAD` shows a single hunk at the top of the file, so the body is byte-unchanged under the `Critic: PASS` it already carried.

## Blocking findings, fixed

1. **§8's raw-versus-rendered source arithmetic** claimed sixteen raw and two rendered against a citation list of **seventeen and one**, and contradicted the same section's own counting note. Re-enumerated and corrected.
2. **Four internal cross-references read `§4.3.3` where the content they point at is `§4.3.4`** (Observations A and B). Found at round 2 — a defect round 1 missed, in text round 1 had repaired. Each was checked against what its own sentence claims before being changed; two now name the Observation letter.
3. **A misattributed finding in the supersession block.** It claimed *three* findings of the prior paper survive the scope correction. The third — "parent-written observables that went unread for weeks" — **is not in that paper.** A `grep` of its ~1,100-line body for `unread`, `nobody reads`, `no reader`, `exhaust` returns nothing outside the block itself; the finding is native to the *superseding* paper, derived there from `run_log.py`'s docstring. The claim is withdrawn in the block, and **the analyst found the same defect had propagated into the new paper's own scope note** and corrected it there too.

Also fixed: three off-by-one line citations, one incomplete `${RUN_ID}` occurrence claim, one internal-arithmetic slip, a dangling `C-EX` reference, and an undisclosed quote normalisation.

## Two findings worth an operator's eye

**The superseded paper's header was non-conformant, and the consequence was live.** Research Standard §3 binds the form: a superseded paper carries `Revalidate: retired — <N> months` plus a one-line `Superseded by:` note, and is excluded from refresh cycles. The header still read `high — 6 weeks` with no `Superseded by:` line — so **this cycle's own currency table listed the paper as `current (due 2026-09-23)`: a retired paper sitting in the live refresh queue.** Now `retired — 6 months`, with the interval set at the top of the low band because a paper that will never be refreshed wants the value that fires least often. `Last validated:` and the existing `Critic: PASS` line were left byte-unchanged — supersession changes a paper's *standing*, not the accuracy of its recorded verification.

*I did not recompute any currency verdict.* The table was authoritative on currency and I used it verbatim; what changed is the artifact's conformance to a binding header rule, which is a different axis.

**One quote in the supersession block has no retrievable record.** The operator sentence *"IT HAS NOTHING TO DO WITH EDGES AND PLACEMENT OR MOVING CONTAINERS."* is quoted verbatim. Searched `git log --all -p -S` across the full history, every file in the tree, and the bodies and comments of PRs #84 and #86 — **no instance outside the block itself.** This is not evidence it was not said; an utterance in a dispatch prompt is genuinely outside a git-based verifier's reach. The quote is kept and now carries a sourcing note naming the three searches and warning that a third artifact must not re-cite it as traceable. **UNVERIFIABLE, flagged, not blocking.**

## One measurement that is a finding, not a defect

The paper cites `.claude/logs/` at 175 files / 262 MB. That directory is gitignored and machine-local — **it does not exist in the worktree at all**, so the figure is not reproducible from the repository. Five readings of the identical enumeration were taken on 2026-08-12 across the run's passes: **175, 176, 176, 177, 178 files and 262–265 MB**, all with the same 125-day span. They are now tabulated in §4.4, attributed by pass, with the rows this paper measured itself distinguished from those recorded as reported — rather than averaged into a single figure. The stable shape is stated as a floor (`175+ files, 260+ MB`) so it cannot go stale on the next dispatch. **The spread is not a defect in the measurement; it is the measurement's finding** — the surface grows one file per dispatch with no pruning code.

## Decision Log

- **[Low]** Dispatched a *scoped* critic on `cross_node_memory_protocol.md`'s 20-line inserted block rather than skipping it. Alternatives: skip it under the "paper unchanged since its recorded verdict → do not re-dispatch" rule, since the body was untouched and carried a PASS; or re-verify the whole paper. Why: the block is new claim surface with checkable assertions, and the rule protects *unchanged* papers — this one changed. It earned itself, returning a blocking provenance defect that had also propagated into the new paper.
- **[Low]** Closed the last completeness gap with an analyst turn I verified by `grep` myself, rather than spending a fourth critic round or rejecting it as de minimis. Alternatives: reject (content was already verified correct); spend a round the ceiling does not allow. Why: the remedy was one clause, and my own disposition rule says an item whose remedy is already written gets applied. **The weakness is that I verified this one edit rather than a critic** — flagged deliberately. The analyst also went past the relayed item: it found the phrase *"both citation sites"* was itself the under-count (four sites, not two) and enumerated them, rather than fixing the reported symptom and leaving the wrong number two lines below.
- **[Medium]** Scoped rounds 2 and 3 to the repaired spans, re-cloning nothing. Alternatives: full re-verification each round, as the prior cycle did. Why: Research Standard §3 states the burden — *"A repaired span is a new claim"* — and `verify.md` itself names re-verifying-everything as the real defect the 3-round ceiling stands in for. **Scoping did not cost detection:** round 2 caught a blocking defect round 1 had missed. Placed as candidate C-075.
- **[Medium]** Conformed the superseded paper's header to `retired — 6 months`, which changes what the refresh gate computes and moves it out of the queue the currency table placed it in. Alternatives: leave it, since the table was handed to me as authoritative and I was told not to recompute. Why: the table is authoritative on *currency*; §3 binds the *form* of a superseded artifact's header, and the live interval was actively wrong. Different axis, so no conflict — but it changes a machine-parsed value, which is why it is here rather than buried in the diff.
- **[Medium]** Accepted two analyst overrides of critic claims. The analyst contradicted the round-2 critic's `.claude/logs/` reading with its own measurement rather than transcribing the relayed number, and contradicted the round-3 critic's attribution of the §5.2 edit to round 3 (it was round 1, evidenced against `git show HEAD:<paper>`; no intermediate commit existed, so the critic was reading a cumulative diff and could not see the boundary). Why: both overrides carried evidence, and relayed remedies are proposals to be checked.
- **[Medium]** Kept the paper's `Critic:` field long (~46 lines). Alternatives: compress to a verdict word and date. Why: the standard requires a paper read on its own to carry its verification evidence and sets no cap, and the field separates four things a downstream consumer cannot otherwise reconstruct — what round 1 established, what the correction pass re-derived itself, what each round fixed, and what stays flagged.

## Deferred Work

- **Scoping a correction round's re-verification to what changed** — Why deferred: it is a change to `research_verify`'s prompt contract, not to the artifacts this run was dispatched to verify; building it inside a verify run would be a tooling change made incidentally inside an evidence run. Tracked at: `docs/standards/architecture/research/candidates.md`, row **C-075**, `decision` blank (untriaged is the truth) and `status: open`. Verified by: `git show HEAD:docs/standards/architecture/research/candidates.md | grep "^| C-075"` on this branch → row present, reading `decision=[  ] status=[ open ]`, titled *"Scope a correction round's re-verification to the spans that CHANGED, so `research_verify` stops re-downloading a corpus it has already verified"*. **On this branch; lands at merge** — `finding-routing.md` §4 requires a producing run to place its own proposal in its own PR, which by construction is not on the default branch yet. If this PR closes unmerged, the row goes with it.

Nothing else was deferred. Every other finding from all four verification passes was fixed in this PR.

## Post-Run Reflection

**Friction**

- **The currency table and Research Standard §3 point opposite ways for a superseded paper, and nothing reconciles them.** The table is declared authoritative and "do not recompute" — but it listed a paper as `current (due 2026-09-23)` *because* that paper's header was non-conformant to the §3 rule that would have excluded it. A run that treats the table as settling the question inherits the defect and ships it. It cost a paragraph of reasoning to establish that conformance and currency are different axes; a sentence in the currency-table preamble saying so would remove the ambiguity.
- **The "paper unchanged since its recorded critic verdict → do not re-dispatch" rule has no case for a paper whose *body* is unchanged but which gained new claim surface.** Read literally it points at skipping; read by intent it points at verifying the new text. Here it mattered — the added block carried a blocking defect that had also propagated into the other paper.
- **No intermediate commits meant round attribution could not be derived from `git`.** The round-3 critic mis-attributed an edit's round because it was reading a cumulative diff with no boundary in it. The analyst caught it, but only because it held its own round-by-round span lists in context. A fresh actor could not have.

- The `.claude/logs/` figures are only citable as a lower bound because the surface is gitignored and unpruned. §7 of the paper already proposes the experiment that would settle whether a retention rule is needed; worth noting that **five enumerations on a single day produced four distinct values**, which is a stronger argument for the experiment than the paper's original single figure was.

- **C-075 is the substantive one** — placed rather than described, per the rule that a proposal surfaced only in a PR body dies at merge.
- The instruction to relay critic remedies as proposals the analyst may override **earned itself twice on this run**: the analyst rejected a relayed figure in favour of its own measurement, and rejected a relayed round attribution with evidence. Both would have written a false statement into the artifact if applied as given.
- Instructing the critic to supply verdict *content* while the analyst renders the header line also earned itself: the header the analyst wrote is more conservative than the relayed content in one place (it says round 2 re-derived *nine* of ten round-1 repairs and found all nine holding, where round 2 actually found the tenth holding for its own target too). Conservative drift in that direction is the correct failure mode for a verdict line.

Pumapumapumas:
# REVIEW-PR — PR #86, pass 1

**Verdict: `MERGE`.** All six blocking checks pass, zero blocking findings, and I am recommending nothing be changed inside either paper. A clean research PR returns MERGE with nothing owed — this is that.

The two items the producing run itself flagged for an operator's eye both terminated correctly *inside this PR*: the superseded paper's non-conformant header was **fixed** by the verify pass (I re-ran the authoritative parser to confirm), and the `verify.md` scoping proposal was **placed** as `C-075` rather than described in a PR body.

---

## The six blocking checks — what each one actually ran

This is a fresh-context integrity pass, not a second opinion on research that already passed three `research-critic` rounds. What I ran is the layer the critic does **not** cover: internal link integrity, header machine-parseability against the real parser, platform accuracy, and table rendering.

| # | Check | Result | How |
|---|---|---|---|
| 1 | Citations resolve, claims match | **PASS** | All 18 sources present, `[S1]`–`[S18]`, no gap, no reuse. **All 19 cited URLs sweep live today**: 18× HTTP 200 direct; `docs.claude.com/en/docs/claude-code/cli-reference.md` returns 301 → `code.claude.com/docs/en/cli-reference.md`, HTTP 200 at **exactly 106,498 bytes** — the byte figure §6 G1 states. Three spans re-verified byte-exact by raw retrieval in a context that never saw the critic's work: `#define MAX_ARG_STRLEN (PAGE_SIZE * 32)` [S17]; NDJSON §3.2's *"If the JSON text is not parsable, the parser SHOULD raise an error"* [S18]; and three separate Airflow spans [S8] — *"Tasks are entirely isolated and may be running on entirely different machines"*, *"only designed for small amounts of data; do not use them to pass around large values, like dataframes"*, and *"XComs therefore can't be used to persist state across task retries"*. Plus [S13]'s `--json-schema` sentence, byte-exact against the redirect target. |
| 2 | Internal links resolve | **PASS** | Every relative link in **both** papers extracted and stat-ed. New paper: 4 (`cross_node_memory_protocol.md`, `memory-model.md`, `run_log.py`, `temporal.md`) — all resolve. Superseded paper: 3 — all resolve. **0 broken.** |
| 3 | Header block machine-parseable | **PASS** | Not asserted — **run through the authoritative parser.** Lifted `_RETIRED` / `_REVALIDATE` / `_LAST_VALIDATED` out of `research_activities.paper_currency` and applied them to the pool: `cross_node_memory_protocol.md` → **`RETIRED — excluded from refresh`**; `state_passing_between_workflow_children.md` → `validated 2026-08-12, interval 6 weeks, due 2026-09-23, current`. Research Standard §3's required `**Superseded by:**` line confirmed present by anchored regex. |
| 4 | Honest-boundary section present | **PASS** | §5 with four subsections — §5.1 (the decision no source could settle), §5.2 (the strongest case *against the paper's own thesis*, with three named weakenings each carrying its own strength), §5.3 (the `--resume` alternative stated fairly, with its weak reason marked weak), §5.4 (where the comparative evidence does not transfer). §6 carries five gaps G1–G5, each with its search method. |
| 5 | No false statement about our platform | **PASS** | Every platform claim I could mechanically check, checked. **Byte counts exact**: `disposition.md` 72,002 + `criteria_research.md` 3,112 + `core_corpus_rule.md` 1,112 = **76,226**, the figure §2.3 rests its 58%-of-ceiling finding on. `research_workflow.py` is **95 lines** as claimed; `:38` `notes: list[str] = []`, `:42` `worktree_add`, `:48` `pr_url = write.run_write(`, `:56` `pr_number_from_url(..., expected_repo=slug)`, `:77-78` the returned dict — all exact. `run-claude.sh:142` `claude -p "$prompt"`, `:155` `--json-schema`, `:299` the completion contract — exact. `MEMBER_EVENT_TYPES` at `run_log.py:58`, `JOIN_KEY = "run_id"` at `:64` — exact. **The `grep` claims re-run rather than trusted**: 12 `prompts/` directories exist and `RUN_ID` appears in **exactly one file**, at `disposition.md:255` and `:420` — both line numbers correct. `exit_record.schema_argument()` is supplied from **exactly one site**, `review_pr_workflow.py:135`. `render_prompt` substitutes **exactly six** placeholders. `memory-model.md` §7.2 enumerates **fourteen** items with **three** marked ⚠ as having no yaml field (rows 3, 4, 11) — both counts exact. `.gitignore` line 2 is `.claude/`. |
| 6 | Source-count floor met | **PASS** | 18 external against Research Standard §3's `10–20 credible sources` band (`research_standard.md:93`). **And against the ceiling that landed the same day**: `write_minor.md:11` binds **20 cited sources**, corrected up from a briefly-set 12 by `74536bb`. I checked the direction the instruction asks for *and* the other one — `85774f9`, `74536bb` and `94f07ae` are all **ancestors of HEAD** (merge-base is `89eed0d`), so the branch was cut after both bound corrections and the paper's *"binding ceiling for this dispatch was 20 and is not breached"* is measured against the live rule, not a stale one. The 3-round verify ceiling (`verify.md:24`) likewise post-dates nothing here; the verify pass closed at round 3 of 3. |

**Table rendering, run because a table break is invisible in `git diff`.** Rendered **both** changed markdown files in full through `POST /markdown -f mode=gfm` and read the output, not the diff: the paper renders **7 `<table>` / 7 `</table>`**, `candidates.md` renders **10 / 10**, and **zero** pipe-rows fell out as `<p>|` orphans in either. `candidates.md`'s prose count was checked against its own rows: it declares *"Twenty-seven rows are untriaged… C-047 through C-075"* and enumeration returns exactly 27 (C-047–C-062, C-064–C-069, C-071–C-075).

**Sweeps that returned nothing, stated so absence reads as checked rather than skipped.** *Deleted-artifact sweep:* the PR has **exactly 3 deleted lines** in total and I enumerated all three — the old `Revalidate:` line, the old `file_structure.txt` description, the old "Where things stand" paragraph. Each is replaced in place; nothing is dropped. `cross_node_memory_protocol.md`'s three hunks all sit at `@@ -1`, `@@ -12`, `@@ -96` — the header region — so its ~1,100-line body is byte-unchanged under the `Critic: PASS` it already carried, exactly as claimed. *Completion-checkbox sweep:* **N/A** — no `[ ]` → `[x]` flip anywhere in the diff, and no planning artifact touched. *Merge-commit sweep:* **N/A** — no merges in the PR. *Attestation:* 2 commits, 2 reflection comments, both substantive and both attested; **no gap**.

---

## Dispositions

|---|---|---|---|
| `cross-node-header-nonconformance` | standards-implication | **FIXED** | The write pass deferred this; the verify pass fixed it, and I verified the fix rather than the account of it. Research Standard §3 (`research_standard.md:84`) binds `Revalidate: retired — <N> months` plus a `Superseded by:` note for a superseded paper. Both now present, and — the part that actually mattered — **the live consequence is gone**: the paper was sitting in the refresh queue as `current (due 2026-09-23)`, and the authoritative parser now returns `RETIRED — excluded from refresh`. |
| `c075-verify-scoping-proposal` | deferral | **DEFERRED** | Pointer **verified present**: `candidates.md` row `C-075`, `decision` blank, `status: open`, in this PR's diff. Correctly classified a **proposal**, not a defect — it adds a capability `verify.md` does not have — so `candidates.md` is its home and an Issue would have been wrong. The index paragraph was updated with it (26→27), which is the failure mode that file documents against itself. Why now isn't the time: acting on it is a `plan-sprint` decision, not a research run's. |
| `file-structure-map-line` | doc-drift | **FIXED** | The map's one-line description still carried the superseded "nodes that come and go" framing, which this PR makes actively wrong. Now describes the pool accurately, and its granularity matches its siblings (`planning-and-agents/`, `workflow-decomposition/` are likewise unexpanded) — so no new inconsistency. Proactive-doc-management applies and the run was right not to leave it. |
| `worktree-path-write-hazard-recurrence` | friction | **NOTED** | **Second consecutive occurrence** — PR #84 and PR #86 hit the identical failure: the orchestrator composed the analyst's absolute output path from the project-root convention rather than the worktree root, and the analyst wrote the paper into the **main checkout**. Both times the workflow's own mandated pre-commit `git status` / `ls` check caught it before commit. Preventive only, and the recurrence is the part worth an operator's eye. |

**Nothing vanished.** Every item either run raised reaches a terminal state above. **Predecessor check (PR #84):** its `component-vs-phase-ruling` hold closed into `C-074`, which I confirmed live on `origin/main`; `test-plan-items-1-7` was rejected at its pass 2; `c073-normalisation-proposal` and the three map/cap items are fixed. **No #84 deferral carries a trigger this PR fires.**

### On the `worktree-path-write-hazard-recurrence` NOTED, stated in full because NOTED is the disposition most easily abused

**No live defect — here is the check I ran, not an impression.** `git status --porcelain` on the main checkout `/home/puma/Repos/claude-dot-files` returns **empty**, and `find … -name 'state_passing_between_workflow_children.md' -not -path '*/worktrees/*'` returns **nothing**. The stray write was fully reverted; the paper exists once, in the right place.

**Nothing breaks if this is never done.** The guard that caught it is the thing that makes it preventive: the mandated pre-commit check fired on both occurrences, which is a guard working twice, not a guard missing twice.

**No work item falls to this PR.** The remedy is a line in the research-family workflow prompts — a different subtree, unrelated to this artifact, and a change the research cycle's own design principle says must not be made incidentally inside a research PR. The producing run placed it on its designated surface (Post-Run Reflection → *Tooling-level suggestions (claude-dot-files)*), which `claude-dot-files-governance.md` names as **the** canonical channel for this class with the operator as its reader. PR #84's pass 2 ruled the identical mechanism `noted` under `deferral-trigger-has-no-watcher`, and I am consistent with it rather than re-litigating it.

**What I am adding beyond that ruling:** the occurrence *count*. This is now twice in two consecutive PRs of the same workflow family, which is the CPI signal the reflection channel alone does not surface. Recorded here so the architecture session sees a pattern rather than two independent anecdotes.

### One thing I checked and demoted rather than filed

`candidates.md`'s "Where things stand" paragraph explains the `C-063` numbering skip but not the `C-070` one. I chased it: `C-070` is explained **inline at the `C-072` row** (*"Renumbered from C-070 on 2026-08-12 — PR #82 merged first and took that ID"*), the declared count of 27 is correct, and the sentence never claims C-063 is the *only* skip — so nothing is false and nothing breaks. A bare discrepancy with no consequence is a note, not a finding, and it does not enter the runway.

---

## Why MERGE and not HOLD

Six blocking checks, six passes, zero blocking findings. The one open item is preventive, is on its designated surface, and would not have blocked this PR on any pass. Research also carries a staleness cost nothing else does — this paper's `Revalidate:` window opened today, and holding it for a workflow-prompt suggestion would burn that window against standards work that research is supposed to *precede*, not follow.

Two things I deliberately did **not** treat as findings, because they are this PR's deliverable rather than defects in it: the `C-075` candidate, and the paper's four facet answers with their standards-adjacent implications. Research proposes; a planner rules. Converting a research run's intended output into holds is the category error this review type exists to avoid.

```yaml
pr_review:
  pr: 86
  run_id: 5818e5ec0fd140c4a34400da6b13d954
  pass: 1
  attempt: 2
  verdict: MERGE
  converged: false
  findings:
    - id: cross-node-header-nonconformance
      title: a superseded paper sat in the live refresh queue, so a retired multi-machine framing would be re-validated as current evidence
      consequence: >-
        Research Standard §3 excludes a superseded paper from refresh cycles via its header. Left
        non-conformant, `render_paper_currency` reported `current (due 2026-09-23)` for a retired
        paper — putting stale multi-machine material back into a future refresh cycle's input set
        and inviting a third artifact to cite it as current evidence.
      disposition: fixed
      remedy: fix-in-place
    - id: c075-verify-scoping-proposal
      title: without a placed row, the correction-round scoping that saved this cycle's verify budget dies at merge and the next run re-downloads the corpus
      category: deferral
      consequence: >-
        `verify.md` names this defect and deliberately does not fix it, bounding damage with a
        3-round ceiling instead. The scoping existed only in this run's hand-written dispatches, so
        a proposal left in a PR body means the next verify run reinvents it or pays the full
        re-fetch cost the ceiling was set against.
      disposition: deferred
      pointer: docs/standards/architecture/research/candidates.md row C-075 (decision blank, status open) — in this PR's diff, lands at merge
      pointer_verified: true
    - id: file-structure-map-line
      title: the authoritative repo map described the pool by the framing this PR supersedes
      category: doc-drift
      consequence: >-
        `docs/file_structure.txt` is the authoritative map. Left unchanged it would describe the
        persistent-memory-protocol pool as backing "carrying memory across nodes that come and go"
        — the exact framing this PR retires — so a reader orienting from the map would size the
        work on multi-machine scope that §5.1 explicitly rules out.
      disposition: fixed
      remedy: fix-in-place
    - id: worktree-path-write-hazard-recurrence
      title: two consecutive research PRs wrote their paper into the main checkout because a dispatch-supplied absolute path was worktree-unaware
      category: friction
      consequence: >-
        Second consecutive occurrence across PR #84 and PR #86, same mechanism. Caught both times
        by the mandated pre-commit check, so nothing shipped wrong; the consequence if the guard
        ever misses is a paper committed to `main` outside its PR. Preventive, and the recurrence
        count is the CPI signal the reflection channel alone does not surface.
      disposition: noted
      remedy: none
      no_live_defect_check: >-
        Ran `git status --porcelain` on the main checkout `/home/puma/Repos/claude-dot-files` — empty.
        Ran `find /home/puma/Repos/claude-dot-files -name 'state_passing_between_workflow_children.md'
        -not -path '*/worktrees/*'` — no results. The stray write is fully reverted and the paper
        exists exactly once, in the worktree. Remedy is a workflow-prompt line in a different
        subtree, already placed on the governance-designated Tooling-level-suggestions channel;
        PR #84 pass 2 ruled the identical mechanism `noted` under `deferral-trigger-has-no-watcher`.
  laundered_deferrals:
    caught: 0
    of_total: 2
  homeless_items: 0
  redispatched: false
```

**Note on `laundered_deferrals: 0 of 2`, since one of the two had an invalid pointer.** The write pass's deferral of the header reconciliation pointed at *"surfaced here in this PR body/comment — no existing tracker covers it"*, and the reviewed PR is never a valid pointer. It is **not** scored as laundered, and the distinction is the one the metric exists to draw: laundering is burying an item behind a **plausible-looking** pointer, and this run did the opposite — it disclosed the absence of a tracker in the same sentence. It was a scope misread (a fixable item dispositioned as a deferral), and the very next pass in the same PR fixed it. Nothing was buried, so nothing is charged.

---

## Post-Run Reflection

**Friction**

- **The `laundered` / `homeless` taxonomy has no member for the shape I actually met.** The write pass raised a legitimate item, honestly stated it had no tracker, and pointed at the PR — which is the literal definition of LAUNDERED, while being the opposite of laundering in intent. It was not HOMELESS either, because a home existed (the fix), which the next pass proved by taking it. The real shape is *a fixable item mis-dispositioned as a deferral*, and the schema cannot say that. I scored it 0 and wrote a paragraph explaining why; a run under turn pressure would have scored it 1 and charged a producing run for disclosure. Worth a third classification, or one clause saying a deferral whose pointer *discloses its own absence* is a mis-disposition rather than a laundering.

- **The post-branch-cut instruction only asks one of the two questions it needs to.** The prompt tells me to check whether a prescription has *already been done*, and adds the reverse check (rules that arrived after the cut) as a measured afterthought. Here the reverse check was the load-bearing one: two commits (`85774f9`, `74536bb`) set and then *corrected* a source ceiling and a round ceiling on the same day this branch was cut, and a 12-source cap — had it still been live — would have made this 18-source paper non-conformant. It turned out both were ancestors of the merge-base, so the paper is measured against the live rule. But I only knew that because I ran `git merge-base --is-ancestor` on each; nothing asked me to, and the framing that got me there was luck, exactly as the prompt's own note concedes.

- The research-type criteria's *"a clean research PR returns MERGE with ZERO findings"* line did real work on this review. Without it the four items above read as four findings and this is a HOLD; with it, three are terminal-on-arrival and one is preventive. Recording that the instruction earned itself, since the criteria file asks for specifics.
- The header-parseability check is the highest-yield of the six and is the only one stated in a form that invites *assertion*. Running the artifact through `research_activities.paper_currency`'s own regexes — rather than eyeballing the header — is what turned "header looks conformant" into "the refresh gate now excludes it", which is the claim that actually matters. Worth stating in the criteria as the required form of check 3, as PR #84's pass 1 also did by instinct rather than instruction.

PR #87, opened 2026-08-13
plan-revision: plan the persistent memory protocol as its own component

Pumapumapumas:
Turns `docs/development/persistent-memory-protocol/research/synthesis.md` — 15 adopted concepts from the 2026-08-12 operator+PM session — into a buildable plan. All 15 are phased; none deferred out.

## The structural decision: NEW COMPONENT, and the reasoning

Placed as `docs/development/persistent-memory-protocol/` with `roadmap.md` + phase docs, **not** as a 7th phase of the six-of-six-complete Memory Management Framework. Against §0's own test — *does this stand up a new domain, or extend an existing one?*:

- **MMF states its own scope**: *"the typed record and its schema; what a parent may route on without a model in the loop; how that relates to the durable human-readable record already in git; and the fail-safe contract when the record is absent or malformed."* All four describe **a handoff between two steps of one run**. This component's subject is a **store** — root path, bag per run, checksum manifest, content-addressed cache, snapshots, rotation, edge identity, cross-machine aggregation. No MMF clause reaches any of it.
- **It inverts authority over MMF's own outputs** — `candidates.md` becomes a projection of the journal.
- **The journal is not a Kind 1 record**, so it cannot be that interface's next binding: no to-do bit (property 4), no selection rule, immutable and never edited. It is the substrate Kind 1 surfaces are *rebuilt from*.

**The strongest counter-reading is stated in the roadmap, not omitted**: `memory-model.md` frames Kind 1 as substrate-free and names edge devices explicitly, so "this is Kind 1's third binding" is a serious read. It is rejected on the property test above. Two weaker draft arguments were **dropped at review and recorded as dropped** — "MMF is six-of-six complete" (MMF itself added a 6th phase to a shipped component 3 days earlier) and "MMF's binding is git" (contradicted by the substrate-free framing).

**`C-074`'s `decision` stays BLANK** — that flag is `plan-sprint`'s write and the ruling is the operator's. Its Note now points at the reasoning. If ruled the other way the remedy is mechanical: phase docs move into `memory-management-framework/` and renumber from 7.

## A task premise that did not hold

The brief asserted **"Documentation Standard §0 is binding."** It is not, here. §0 exists (`documentation_standard.md:69`, marked `(binding)`) but is nested inside `## Development Planning Files`, which `docs/standards/documentation/README.md:36` excludes wholesale (*"assumes the master-planning layout"*). The plan cites **`sprint.md:28`** for authority and treats §0 as a pattern — the same treatment the MMF roadmap gives `§Runtime Verification`. Substance is identical, so the decision is unchanged. Independently re-verified line-by-line by standards-architect.

This is the **third** instance of one hazard in that vendored doc; the README records two earlier ones, and upstream fixed the last by promotion. Placed as **`C-078`**.

Also corrected: the brief cites "the synthesis's §5.1" for the component-or-phase trap — that passage is in `cross_node_memory_protocol.md` §5.1, not the synthesis.

## Planning changes

- **`roadmap.md`** — 8 phases in rollout order (6 before 5), the component ruling, a 15-concept→requirement coverage map, dependencies, and § *Open inputs*
- **`phase1_the_run_bag.md`** — configurable root, folder per run keyed by `run_id` (never by path), subfolder per concurrent child, BagIt RFC 8493, bag lifecycle, payload spec, no-database with its revisit trigger
- **`phase2_content_store.md`** — bernstein §4.6: store the bytes, hash them, re-verify quoted spans offline. Tier 1, ranked and then never placed once already
- **`phase3_the_emit_rule.md`** — if any store gets it the journal gets it verbatim; destination is a field not a format; event admission contract; stable `edge_id`
- **`phase4_rebuild_is_a_test.md`** — replay and diff; two arms (synthetic fixture on the merge path, live journal host-only)
- **`phase6_cpi_reads_the_journal.md`** — added at review (see below)
- **`candidates.md`** — C-077, C-078 appended; C-074 Note amended; untriaged count 28→30 re-derived by command
- **`file_structure.txt`** — set-equal to the tree (`test_file_structure_map_covers_the_tree.py` passes)

**Sequencing the synthesis fixed, left intact:** CPI on Edge1 until a second edge produces runs; S3 after; no database; BagIt not an invented format; snapshots before rotation.

## Deviations from plan

- Phase docs for **5** phases, not 3 — the rule became "a phase doc exists iff the phase has no gate outside this component", stated in the roadmap along with its divergence from `sprint.md`'s pickup test.
- **8 phases, not 7** — review split Phase 6.
- `candidates.md`'s untriaged count updated (that file's own rule makes the number load-bearing); the undocumented C-070 gap recorded, since the count was re-derived by command.

## Review

**architect** — 3 Critical, all fixed. (1) The journal silently absorbed `.claude/logs/`, an MMF-Phase-6-declared surface with three committed readers → Phase 1 requirement 7 now forces the answer. (2) No event identity against Temporal's at-least-once model → folded into the new admission contract. (3) Phase 4's rebuild targets are the two stores with out-of-run human writes → ruled explicitly. 5 Warnings fixed (bag lifecycle, `bag-info.txt` not `bagit.txt`, the component argument's weak checks, per-step emit, the Phase 6 split); 2 Info fixed.

**security-auditor** — 5 High, all fixed. Capture-time secret handling (**the one place the plan was strictly weaker than today** — `.claude/logs` is equally unredacted but deletable) → redaction event class + capture-time filter. Replay path containment. CI fixture provenance. `edge_id` assignment authority (spoofable as drafted). Credential epoch. 9 Medium fixed: egress/ingress gate split, provenance field, Phase 7 checkbox wording, digest-only paths, fetch policy, read-path invariant, root `0700`/`0600`, originating-project field, manifest over-claim.

**planner** — 3 Critical, all fixed. Model-issued writes are invisible to a tree search (the PR body and comments would go unemitted while requirement 1 read as met). Phase 4 has no journal in CI. Phase 4 has no replay baseline. 8 Warnings fixed. Verified the coverage map independently: 13/15 mapped to a numbered requirement, 2 to prose — map now annotates which.

**standards-architect** — 0 Critical. 2 Warnings fixed (`state_passing §4.3.1` → `§4.3.3`, propagated to 2 files; superseded-paper citation caveat). 1 surfaced (`sprint.md`, below). Confirmed clean: `candidates.md` conformance, no vendored standard edited, `file_structure.txt` matches disk.

**quality-control** *(override authority)* — 2 integration Criticals, both fixed. Ruled the envelope findings **one** decision, remediated as one requirement. Ruled Phase 4 the disqualifying gap and pointed at this repo's own precedent (`tests.yml:146-155` refuses skip-when-absent, issue #55) — applied. Cleared the "phase doc iff unblocked" rule as a real trigger-gated placement, not deferral in disguise. Verdict: *"revise these two sections, not redesign the component."* Both revised.

**Cross-reference consistency: PASS.** All relative links resolve, all anchors resolve, no stale `§4.3.1`. `./testing/run-all.sh` → **PASSED (1120 passed, 3 skipped)**.

## Operator calls this plan does NOT make

1. **Storage budget and snapshot cadence** (Phase 5) — two numbers that trade against each other. Requirement stays unchecked with prose.
2. **What an edge actually is** (Phase 1) — HAOS may have no user account. Root stays a config value so it does not block.
3. **Phase 7's egress and ingress rulings** — two rulings, not one; both unchecked boxes.
4. **A `sprint.md` entry.** `sprint.md` is operator-only and has no `## Sprint: Persistent Memory Protocol` row (`grep` returns nothing), so the component is planned but unreachable from the file operators start at. **Surfaced as a sprint-item candidate; not written.** Suggested placement: after *Memory Management Framework*, gated on `C-074`'s ratification.


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Wrote phase docs for 5 of 8 phases, under a rule this run invented: *a phase doc exists iff the phase has no gate outside this component.* Alternatives: write all 8 (the DoD asks for a buildable plan), or write none (`sprint.md:30` conditions doc-writing on **sprint pickup**, and this component has no sprint row — read literally, none of these docs should exist yet). Why: gated phases are exactly where "a detailed plan for work that has not started is a guess that ages badly" bites hardest. **The divergence from the cited line is stated in the roadmap rather than left invisible** — planner flagged it as a re-derivation, correctly, and the fix was to name it as a choice. Quality-control cleared it as trigger-gated placement, not deferral in disguise.

- **[Low]** Split Phase 6 into a reader (Phase 6) and a poller (Phase 8) **during review**, adding a phase to a plan that argued carefully for seven. Alternatives: keep one phase and name a trigger that forces the reader if the Temporal server slips (planner's own fallback); or leave it. Why: three reviewers independently found the draft gating this component's **only consumer** behind an unbuilt server, for four phases of producers — the exact failure the plan cites MMF Phase 6 against, committed by the plan citing it. QC read the tension as narrower than "reproduces the failure" (PMP *named* the gap and pointed at a placed phase, which is MMF's conformant case, not its failure case) — so this is a Warning-grade fix I chose to make anyway, because the split costs a roadmap row and the alternative costs the plan its own argument.

- **[Medium]** Ruled the persistent memory protocol its **own component**, and left `C-074`'s `decision` blank. Alternatives: a 7th phase of MMF; or leave the artifact homeless pending the ruling. Why: MMF's stated scope sentence covers a *handoff*; this covers a *store*, and no clause of the former reaches the latter. Implementing the decision while leaving the flag blank keeps both rules intact — `standards-governance.md` reserves the flag to the operator, and the brief forbids leaving the plan homeless. The remedy if reversed is mechanical and is written down.

- **[Medium]** **Dropped two of my own three supporting arguments at review and recorded them as dropped.** Architect showed both were contradicted by MMF's own artifacts — MMF added a 6th phase to a "complete" component three days before this plan, and `memory-model.md` frames Kind 1 as *substrate-free*, so "MMF's binding is git" argues the opposite of what I cited it for. Alternatives: defend them, or delete them silently. Why: they were the two that read best and held worst, and the counter-reading they left unengaged ("this is Kind 1's third binding") is the strongest objection to the whole ruling. Replaced with the property test, and the counter-reading is now stated **in the roadmap before the operator rules**, not just rebutted.

- **[Medium]** Treated the brief's *"Documentation Standard §0 is binding"* as **evidence, not instruction**, verified it, found it false-as-cited, and built on `sprint.md` instead. Alternatives: take it at face value (the workflow's Stage 1 warns this class has already produced a wrong binding-standard citation in a phase's sizing). Why: §0 is nested inside a section the vendored README excludes wholesale. Independently re-verified by standards-architect line-by-line. Substance unchanged, citation corrected, hazard placed as `C-078`.

- **[Medium]** Fixed the Phase 3 envelope findings as **one** requirement rather than four. Alternatives: four separate requirements matching how the findings arrived (architect's dedupe, security's authority, security's epoch, security's provenance). Why: quality-control's integration call — they are four symptoms of one undecided question (*what makes an event admissible*), and remediating them separately risks four small fixes that do not cohere into one identity model. I took the integration reading over the four narrow ones.

- **[Medium]** Applied `tests.yml`'s issue-#55 precedent to Phase 4 rather than inventing a CI strategy. Alternatives: commit a real journal fixture (security: verbatim transcripts into git history permanently), or skip-when-absent (this repo explicitly refuses that: *"a green check that verified nothing, which is worse than the gap it hides"*). Why: the repo had already litigated the identical dilemma; re-deciding it would have been the easy-path signature of reaching for a familiar approach without checking it fits.

- **[Low]** Updated `candidates.md`'s untriaged count 28→30 and recorded the previously-unexplained **C-070** gap. Alternatives: leave the paragraph alone (surgical-changes discipline). Why: that file's own rule says *"blank is only honest while this paragraph says how many there are"*, so my two rows made the number wrong. Having re-derived the count by command rather than incrementing it, the C-070 gap was in the output and recording it was one clause.

## Deferred Work

- **Research cycle on journal format at volume, and on the egress + ingress rulings for records crossing a trust boundary** — Why deferred: it is a research question, not a planning one, and the ingress half hard-gates Phase 7, which is itself gated on a second edge that does not exist. Tracked at: `docs/standards/architecture/research/candidates.md` row **C-077**. Verified by: `grep -c "^| C-077 " docs/standards/architecture/research/candidates.md` → `1` **on this branch; lands at merge**; row reads `| C-077 | A full research cycle on journal format at volume, and on redaction/classification for records crossing a trust boundary — the second GATES cross-edge aggregation |`, with `decision=[ ]` and `status=[ open ]` confirmed by field-split (`awk -F'|'`). Blank `decision` is deliberate — that flag is `plan-sprint`'s.

- **Promote Documentation Standard §0 out of the excluded *Development Planning Files* section, upstream** — Why deferred: `docs/standards/documentation/` is a vendored MIRROR that must not be edited here; the amendment goes to MDC-Master-Planning and re-vendors. Tracked at: `candidates.md` row **C-078**. Verified by: `grep -o "^| C-078 | [^|]*|"` → `| C-078 | Promote Documentation Standard §0 (component vs phase) out of the excluded *Development Planning Files* section upstream, so a layout-independent binding rule stops being swallowed by a layout-specific exclusion |` **on this branch; lands at merge**; `decision=[ ]`, `status=[ open ]`.

- **A `sprint.md` entry for this component** — Why deferred: `sprint.md` is operator-only under `standards-governance.md`; a dispatch surfaces, never writes. Tracked at: `docs/development/persistent-memory-protocol/roadmap.md` § *Open inputs*, item 4. Verified by: `grep -n "A \`sprint.md\` entry for this component" roadmap.md` → line **215**, reading *"`sprint.md` is operator-only and this plan does not write it. Without a row there the component is planned but unreachable from the file operators are told to start at"* — **on this branch; lands at merge**. It is placed in a durable file rather than only in the PR body, because a PR body dies at merge.

*(Everything else the five reviews raised was fixed in this PR. Nothing was left as a bare pointer.)*

## Post-Run Reflection

**Friction**

- **The brief asserted a binding standard that is not binding here** (`Documentation Standard §0`), and separately attributed the component-or-phase trap to *"the synthesis's §5.1"* when that passage is in `cross_node_memory_protocol.md` §5.1. Both were one grep from being caught, and Stage 1's own warning is what caught them. Worth noting the brief also said *"Anything this brief claims about the artifacts is checkable; check it"* — that sentence did its job.
- **The pool's `synthesis.md` disclaims itself as evidence in its own header** while being the run's designated primary input. That is the right call by the synthesis, but it makes every citation two-hop: read the synthesis for the decision, then cite what it points at. It roughly doubled the reading, and it is the correct cost.
- **`sprint.md` had no row for the component being planned**, so the local convention this plan cites for authority (*"phase docs are written when a sprint is picked up"*) could not be satisfied literally. Resolvable, but it meant inventing a rule mid-build and defending it — planner and QC both landed on it.

- The `state_passing §4.3.1` → `§4.3.3` mis-citation propagated to two documents from one draft. Given how much of this repo's persuasive weight rests on exact `§N` citations, a mechanical `§`-reference checker over `docs/` would be cheap and would catch the class that already has a named row (`C-065`).
- `candidates.md`'s "Where things stand" paragraph is a hand-maintained count over a table in the same file. It has been wrong before (the file says so). It is derivable — `grep -oE "^\| C-0[0-9]{2}" | sort -u | wc -l` — and a test asserting the paragraph against the table would retire the class.

- **The workflow's Stage 4a dispatch does not tell reviewers which brief premises the run found FALSE.** I passed that context by hand this run and standards-architect independently re-verified the §0 correction, which was worth having. Making "premises the run rejected, and on what evidence" a named element of the reviewer dispatch would make that verification systematic rather than a judgement call by the orchestrating run.
- **Stage 4b's value was concentrated in one move the narrow lenses structurally cannot make**: ruling that four findings from three agents were *one* decision, and that remediating them separately would produce four fixes that do not cohere. The prompt asks QC to "look for meta-patterns"; it might ask more specifically for **finding consolidation** — *which of the quad's findings are one finding, and what is the single remedy* — since that is the judgement the sequential position exists to buy.
- The workflow left me on a **detached HEAD**, so `git push -u origin HEAD` failed with *"You must fully qualify the ref"* and I had to create the branch by hand. Minor, but every run on this path pays it.

PR #85, opened 2026-08-12
build-draft: split plan_sprint into triage_candidates and a narrowed plan_sprint

Pumapumapumas:
Splits `plan_sprint` into **`triage_candidates`** (rules the candidates) and a narrowed **`plan_sprint`** (places what they ruled, maintains the plan).

`plan_sprint` did two jobs in one run — *"triage research candidates and keep the sprint plan current"*, its own docstring — and **nothing could be sequenced between them.** Feature planning and scaffolding belong in that gap, and while both jobs shared a dispatch that was structurally impossible rather than merely unbuilt.

**It fixes an ordering defect on its own:** the sprint plan used to be updated *before* anything estimated the work, so its hour totals landed ahead of the estimates they depend on. `plan_sprint` now runs last.

## What changed

**New workflow, mirroring `plan_sprint`'s standalone-and-composable shape exactly** — no new pattern invented:

| Precedent (`plan_sprint`) | New (`triage_candidates`) |
|---|---|
| `plan_sprint.sh` | `triage_candidates.sh` |
| `run_plan_sprint.py` | `run_triage_candidates.py` |
| `run_plan_sprint()` imported by `plan_project` | `run_triage_candidates()` imported by `plan_project` |
| `plan_sprint/prompts/plan_sprint.md` | `triage_candidates/prompts/triage_candidates.md` |

Both render end-to-end against the real queue — the current figures are under *Verification* below.

**Sequencing in `plan_project`:** `triage_candidates -> research(per NEW component) -> plan_sprint -> review-pr`. The loop-back goes to `plan_sprint` alone — rulings are made once, and re-running triage would re-litigate settled dispositions at a full opus dispatch per pass.

**Shared fragments used, not duplicated.** Both workflows interpolate `decision_log_and_reflection` and `headless_execution_guard` via `act.shared_prompt()`. Two tests enforce it: one that each workflow interpolates them, one that neither prompt contains a line copied verbatim from them (with a positive control proving the probe can find a copy).

# Refine passes — 2026-08-12

Four correction passes have run since the draft, newest first below. The first enforced the boundaries the split had left on prose. The second, after `review-pr` returned **HOLD** with three items, treated each as an **instance** and closed its **class**. The third did the same to the two items held at pass 2. The fourth made the PR mergeable again, and then found that the layer nobody had looked at was the one this PR added last.

[...]

PR #88, opened 2026-08-13
build-draft: plan_candidates — scaffold what the ruled candidates need, and correct the chain that made research inert

Pumapumapumas:
## Summary

Builds **`plan-candidates`** — the child that gives every ruled `ship` candidate somewhere to be built — and corrects the guide's chain, which had the ordering backwards in a way that was load-bearing.

`triage-candidates` decides that something should be built. A `ship` row is a **decision, not a place**, and both children after it need a place before they can run: research is commissioned per component pool, and `plan-feature` plans phases *into* a component. This is the step between the ruling and the work.

Shape mirrors `triage_candidates` exactly — shim (`scripts/plan_candidates.sh`), runner (`scripts/run_plan_candidates.py`), workflow module (`modules/assistant/plan/plan_candidates/`) — standalone-runnable and composable. Verified against the real tree:

```
$ ./scripts/workflows/temporal/scripts/plan_candidates.sh --dry-run
  Candidates : 75 total · 27 ruled `ship` (the working set)
  Components : 16 on disk, 1 with a charter
  Max turns  : 150 (estimate — nothing has measured this workflow)
  Prompt     : 27920 bytes rendered, 0 placeholders remaining
```

---

## The structure-versus-substance boundary — the answer, with reasoning

The brief asked for this to be reasoned out rather than dictated, so here it is in full.

**`plan-feature` owns roadmap and phase CONTENT.** The Documentation Standard § 0 forbids the obvious alternative too: every planning doc is a new component or a new phase of one, **no third category, no standalone design doc**, and rule 1 says *"a file exists only when it carries binding content. An empty cell is a reserved name, never a stub doc."*

**The answer: a `ship` row gets exactly one of three outcomes, and only the third creates anything.**

| Outcome | When | What is created |
|---|---|---|
| Already tracked | An open issue or existing sprint section covers it | **Nothing.** Report which |
| **Extends an existing component** | The work falls inside a component that exists | **Nothing.** Name the target; `plan-feature` writes the phase doc and roadmap row |
| Stands up a NEW domain | No existing component covers it | `docs/development/<slug>/roadmap.md` carrying the CHARTER |

**The honest "no scaffolding needed" case the brief asked me to name is the middle row, and it is expected to be the majority of today's 27.** The prompt makes it a first-class outcome so a run does not invent a domain to have something to create: *"A run that scaffolds one component and correctly reports that twenty-six other candidates extend things that already exist has done this job well."*

**What a charter is** — and why it is binding content rather than a stub:

| A charter states | A plan states — NOT this workflow's |
|---|---|
| What the component IS — the domain it stands up | How the work breaks into phases |
| What it explicitly is NOT — the boundary against neighbours | What each phase achieves |
| Which `C-NNN` rows it derives from | Milestones and completion criteria |
| Which differentiator in `problem-statement.md` it serves | **Estimated hours** |
| What it depends on | Sequencing against other work |

**Why that is content and not a stub:** it is the one thing neither downstream child can supply for itself. `plan-feature` plans phases *inside* a scope; it does not choose the scope. Research is commissioned *per component pool*, so it needs the component to already mean something. The scope boundary is the gap, and the charter is it.

**And the cost of not having it is on disk, measured rather than argued:**

```
$ find docs/development/fleet-reliability -type f
docs/development/fleet-reliability/research/README.md
docs/development/fleet-reliability/research/topics.md
docs/development/fleet-reliability/research/synthesis.md
docs/development/fleet-reliability/research/raw/{5 papers}

$ git log --oneline -2 -- docs/development/fleet-reliability
3b88897 research: fleet-reliability pool — 5 papers, all verified (#51)
755935a docs: scaffold the fleet-reliability research folder
```

Five verified research papers commissioned into a component **no document defines**. The parent's research step created that folder with a `mkdir` at `plan_project_workflow.py:158` because nothing ahead of it wrote what the component was. **The pipeline already produces the exact empty shell the standard forbids** — which is why an empty component directory now fails this workflow's run.

**Counted, not asserted:** 16 of 17 directories under `docs/development/` have no `roadmap.md`; `memory-management-framework` is the only one that does. That measurement decided one design detail — the shell guard is scoped to directories the *run added*, because a tree-wide sweep would fail every dispatch over work nobody in it did.

---

## Authorization — declared, and observed

| You MAY | You MAY NOT | Observed by |
|---|---|---|
| Create `docs/development/<slug>/roadmap.md` | Touch `sprint.md` at all | `FORBIDDEN_PATHS` |
| Write the charter into it | Write or edit any phase doc | `FORBIDDEN_PATHS ^docs/development/[^/]+/phase[^/]*\.md$` |
| Append a blank-`decision` proposal row | **Edit a roadmap that already exists** | `own.roadmaps_edited` over disk-read digests |
| | **Write phases, epics or hour estimates** | `own.phase_planning_in` over created roadmaps |
| | Leave a component directory with no `roadmap.md` | `own.shells_without_a_charter` |
| | Set `decision` on any candidate | `act.rulings_this_run_had_no_right_to` |
| | Set `status` in the candidates file | `act.statuses_this_run_had_no_right_to` |
| | Append to or edit `direction.md` | absence of an exception in `PERMITTED_PATHS` |
| | Edit anything else under `docs/standards/` | `FORBIDDEN_PATHS` less `PERMITTED_PATHS` |
| | Decide which sprint or phase a component belongs in | **JUDGEMENT**, with the reason stated |
| | Delete anything | `act.ids_deleted` ×2, `act.grants_that_vanished` |

**Two rows have no analogue in either sibling, and they are the ones that matter.** This is the only workflow in the family that legitimately creates a file under `docs/development/`, so:

- **"edit a roadmap that already exists"** is invisible to any path rule — `worktree_state` reports a created roadmap and an edited one identically (absent before, present after). So the roadmap map is read from **disk** on both sides and compared by content digest.
- **"plan phases inside the roadmap you created"** is invisible to any snapshot — that file is the workflow's own output. So its **content** is scanned, matching the *claim shape* (`phase\d`, a figure with a unit) rather than a vocabulary, because the charter is *required* to say that phases are `plan-feature`'s.

`sprint.md` is untouched and unreachable. Nothing sets `decision` or `status`.

---

## The guide correction

`docs/guide/workflows.md:39` drew `research(component)` **before** `[plan-candidates]` and showed one `research(component)` step instead of the conditional `research-write` → `research-verify` pair. Corrected to the operator's ordering, with the error and its consequence recorded in place — a step drawn before the thing that produces its input has no input, which is exactly why the research step was inert.

**The research step's input is reachable again.** It read added `## Sprint:` headings out of the sprint diff; with `plan-sprint` moved behind it, nothing ahead of it added one. It now reads charters added between the pinned base and HEAD — same discipline (the parent reads the artifact, never a child's account), and a better brief besides: a heading is a name, a charter states scope and boundary.

## Deviations from plan

Two, both forced rather than chosen:

1. **§10.1 rule 3 promotion.** `candidate_decisions` and `_rulings_this_run_had_no_right_to` gained a second consumer, so they moved from `plan_sprint_activities` back up to `plan_activities`. `workflow-scripts.md` § Location states the test mechanically — *"consumer count decides, never taste"* — and it is a count, not a one-way ratchet. Duplicating them would be the drift `normalise_cell` exists to record.
2. **`new_sprint_sections` and `component_dir` deleted.** Rewiring the research input orphaned both; `scaffolded_components` and `component_pool` replace them. Deleted, not commented out.

## Not built (brief put these out of scope)

`plan-feature` — not created, not stubbed, and never referenced as existing. Per-phase hour estimation is its job and is explicitly forbidden here. No `--project` / `--feature` flags. No re-sequencing of `plan_project` beyond inserting this child and making the research input reachable.

## Standards

**Industry best practice:** a pipeline stage creating a target for later stages writes the *contract* they consume and nothing a later stage owns.
**Project standard:** Documentation Standard § 0 (binding, vendored) + rule 1.
**Alignment:** aligned — both say the artifact must carry real content, and the content must be the part this stage owns.

**Vendored standards untouched.** One gap surfaced and placed rather than edited (below).

## Sprint-item candidates surfaced

None. Nothing in this change implies a sprint edit, and `sprint.md` is neither read nor written by anything here.

---

# REFINE PASS (2026-08-13) — what the review changed

The draft's guard machinery held up: five mutations against it, all five caught by the named tests. **Every defect found was at a seam between this module and the repository** — the code enforcing something narrower than the prompt claimed, or telling the model something false about the tree. Not one was an internal logic error.

## ⚠ RULING REQUIRED — two documents disagree about what a component's first file is called

**This is the one item the refine pass did not resolve, and it is the highest-value thing in this PR for a human to look at.**

| Says | What it says |
|---|---|
| `docs/guide/workflows.md`, Documentation Standard § 0 (vendored) | a component is `<slug>/roadmap.md` + `phase{N}_*.md` — **what this workflow writes** |
| **`docs/development/sprint.md:24,28`** — the operator's own file, which no dispatch may edit | *"A component is a folder, not a file. `<name>/<name>.md` is its phase doc"* · *"A component that outgrows one phase gets its own `roadmap.md`… **One phase needs no roadmap; do not create one to be tidy.**"* |

**Counted: 11 of this repo's 16 components follow `sprint.md`; 1 follows the other shape.** A component being chartered has no phase count yet — that is `plan-feature`'s to determine — so `roadmap.md` quietly pre-decides *"this will outgrow one phase"*, and `shells_without_a_charter` **fails any run that writes the `sprint.md` shape instead**.

**Why this was surfaced rather than settled:** resolving it means either amending `sprint.md` (human-only, `standards-governance.md`) or changing this workflow's primary artifact against the two documents that specify it. Both are operator calls. It is recorded in `plan_candidates_workflow.py`'s module docstring so the next reader hits it before copying the shape.

**Nothing is duplicated while the question is open** — `component_inventory` now marks a component defined by its own `<slug>.md` as ALREADY DEFINED and tells the run not to charter it.

## Fixed — the guard enforced less than the prompt claimed

**1. `FORBIDDEN_PATHS` left most of `docs/development/` open while the prompt said it was enforced.** The pattern was `^docs/development/[^/]+/phase[^/]*\.md$` — **it matches exactly 6 files in this repo.** Unwatched: the 11 component docs named `<slug>/<slug>.md` (the phase docs the row is actually about), `cpi-decisions.md`, every review artifact, every component's `research/` pool. `triage-candidates`, running one step earlier in the same worktree, forbids `^docs/development/` outright.

Now the whole directory, less one exception for this workflow's own output:

```python
FORBIDDEN_PATHS = (r"(^|/)sprints?\.md$", r"^docs/development/", r"^docs/standards/")
PERMITTED_PATHS = (r"^docs/standards/architecture/research/candidates\.md$",
                   r"^docs/development/[^/]+/roadmap\.md$")
```

A component's `research/` pool is deliberately **not** permitted. The objection that the pipeline needs it is a timing error: the research children do write there on this branch, but they run *after* this child returns, and both snapshots are taken around this workflow's own model call.

**This also gives the "Delete anything" row a real mechanism for the first time.** It previously named `ids_deleted` over the roadmap map — but 15 of 16 components have no roadmap, so razing one moved no key and nothing saw it. `boundary_crossings` now covers it via the `ABSENT` sentinel. (The `before_dirs` registry entry also named `ids_deleted`, which takes two dicts, against a `set` — it could not have run.)

**2. The inventory invited the run to charter eleven components that already exist.** `has_research` was `(d / "research").is_dir()`, and the block rendered *"**RESEARCH BUT NO CHARTER** — a shell. It has evidence and nothing saying what the component IS"*, while the prompt calls repairing a shell *the cheapest correct outcome available to you*.

Counted in the tree: **12 components hold a `research/` folder and 3 hold a paper.** So the model was told 9 empty `mkdir`s were evidence — and 7 of those components carry a `<slug>.md` saying exactly what they are, several marked ✅ COMPLETE. A compliant run charters up to 11 pre-existing components and the parent commissions a research cycle into each. **Every guard reports clean, because each creation is legitimate by construction.**

Evidence is now counted from papers, and `<slug>.md` is a first-class state. Rendered against the live tree:

```
Chartered: 1 · already defined by their own `<slug>.md`: 11 · holding research with nothing above it: 2.
```

**3. The phase/hours guard cannot tell a quoted candidate title from a plan, and two `ship` rows will trip it today.** `C-011`'s title ends `(~9 h)`; `C-013`'s contains `Phase 4`. The charter template *requires* a "Where it came from" line saying what those candidates asked for. A compliant run fails **after** the model spent its budget and opened the PR, and the exception kills the whole `plan-project` dispatch.

**The guard stays strict** — loosening it to admit a quoted phase number also admits a planned one, which is the boundary the entire child is defined by. The prompt now tells the run to paraphrase and cite by id, with the two live traps named. Both halves are pinned: the quoted titles must still trip it, the prompt's own example rewrites must still pass.

**4. The parent's research fan-out was unbounded**, and this branch is what made it live. `MAX_NEW_COMPONENTS = 3`; past it the step researches nothing and says so, rather than researching a prefix and leaving the difference looking like a decision.

**5. `COMPONENT_ROOT` existed twice with two different types** — `Path` in `plan_activities.py`, `str` in the new child — and `plan_project_activities` imported the **child's**, so narrowing it in one workflow folder would have silently broken the parent's sweep. Promoted per §10.1 rule 3, along with `NOT_A_COMPONENT`; `existing_work` now reads both instead of hardcoding them.

**6. `scaffolded_components` re-opened a quoting bug its sibling documents as retired.** Added `-z` (without it `core.quotePath` turns a non-ASCII slug into `roadmap.md"` and the component is silently skipped), derived the path depth from `COMPONENT_ROOT` instead of a literal `4`, and applied the `reviews/` exclusion.

## Fixed — claims that were false

**7. The charter template in the test was a hand copy that had already drifted** — missing `## Dependencies` and the `Serves:` line. The one test certifying *a compliant run cannot fail its own guard* was checking an approximation. It now extracts the fenced block from the prompt, and asserts there is exactly one.

**8. Four false factual claims corrected.** *"Sixteen of this repo's seventeen components have no `roadmap.md`"* → 15 of 16, by the module's own definition, which excludes `reviews/`. *"`memory-management-framework/roadmap.md` legitimately lists six phases with estimates"* → six phases, **zero** hour estimates. The `before_dirs` escape clause (above). The inventory's "it has evidence" (above).

**9. `C-078`'s Source cited `PR #87`** in both the row and the summary prose. PR #87 is *"plan-revision: plan the persistent memory protocol as its own component"* — a different, open, unrelated PR. Corrected to #88 in both places.

**10. `docs/development/temporal-integration/temporal-integration.md:5`** still enumerated the planning family as ending at `triage-candidates` in PR #85.

## Surfaced — real, out of scope, no verified home

- **A guard failure leaves a poisoned PR open with nothing on it saying so.** Every guard raises *after* the model has committed, pushed and opened the PR; the `RuntimeError` reaches the runner, which prints to stderr and returns 1. The PR stays open with the violating charter, no comment and no label. This is a **shared-family** issue — `triage-candidates` and `plan-sprint` have the same shape — that `plan-candidates` merely exposes first, because it is the first child that can fail after a sibling already opened the PR. Not scoped into this PR because the fix belongs in the parent's exception handler for all three.
- **No workflow in the fleet can correct a charter once written.** `roadmaps_edited` forbids editing a pre-existing roadmap, so on a `--pr` redispatch `plan-candidates` cannot revise its own output; the loop-back targets `plan-sprint`, which forbids `^docs/development/` entirely. A `HOLD_REDISPATCH` naming a charter defect therefore spends its full runway on an unreachable target. The routing half is small; the capability half is a design question that belongs with `plan-feature`'s scope.
- **A created charter's *content* is unchecked** — `phase_planning_in` only tests for what must not be there, so a one-line `roadmap.md` passes every guard and then becomes a research brief.
- **The charter has no reader.** Stage 1's read list does not include any existing charter, and `component_inventory` passes names and a status mark, not scope text. So the "What this is NOT" boundary — which the prompt calls the section that stops the next candidate landing in the wrong place — is written and never read back. Matters most once `plan-feature` exists.
- **`evidence_block` sits on the shared surface with one consumer** (`plan_revision`) — under §10.1 rule 3's threshold, and absent from the rule-3 audit in `plan_activities.py`'s docstring. Pre-existing; not this PR's artifact.
- **`--pr` redispatch cuts its worktree from `HEAD`** in all three sibling runners rather than the PR's branch. Pre-existing family-wide pattern.

## Rejected

- **"Permit `<component>/research/` so the pipeline's research children can write"** — rejected: those children run *after* this workflow returns, and both snapshots are taken around this workflow's own model call. Permitting it bought nothing and left the run free to rewrite evidence it did not gather. The test asserting it as permitted carried this reasoning; the reasoning was a timing error and the assertion is now inverted.
- **Extracting a `_guard(offences, message)` helper across the three workflow modules** — rejected on measurement: the mechanically identical code is ~30-40 lines against ~250 lines of per-workflow prose, and the comparators are already factored into `plan_activities`. It would save ~15 lines and put a call frame between each `raise` and the sentence that is its content.
- **Splitting `plan_candidates_activities.py`** into comparators and prompt-builders — rejected: 9 functions, one consumer each, all reading the same layer, and `component_inventory` shares `component_dirs` with the guards.

## Test results

`./testing/run-all.sh` → **1357 passed, 3 skipped**, RESULT: PASSED. Draft was 1341; branch point `1d0ab94` was 1275.

**Six mutations against the refine pass's own work, all six caught by the named test:**

| Mutation | Caught by |
|---|---|
| `has_evidence` back to `pool.is_dir()` | `test_an_EMPTY_research_folder_is_not_evidence_and_not_a_shell` |
| drop the ALREADY-DEFINED branch | `test_a_component_DEFINED_BY_ITS_OWN_SLUG_DOC_is_marked_do_not_charter` |
| drop the `roadmap.md` permit | `test_the_two_paths_it_MAY_write_are_not_crossings[the-charter]` + the authorization test |
| `FORBIDDEN_PATHS` back to the narrow phase pattern | 3 boundary cases (component-doc, cpi-log, research-pool) + the authorization test |
| remove the fan-out cap | `test_a_runaway_scaffolding_pass_RESEARCHES_NOTHING_and_says_why` |
| set the cap to 0 | `test_a_fanout_AT_the_ceiling_still_researches` |

**One of those mutations found a defect in this pass's OWN test.** The ceiling discriminator sized its fixture from `MAX_NEW_COMPONENTS`, so at a ceiling of 0 it built an empty list, researched nothing, and asserted `0 == 0` — vacuous, passing, while the step was dead. It now asserts the floor explicitly. It was caught only by mutating a constant after writing the test, which is the only way that shape shows up.

Three of the repo's own gates fired during this pass and were fixed rather than silenced: the string-split census (a new module-scope split, declared with what it splits), and the two `plan_candidates` tests that were asserting the behaviour being corrected.

## Verified

- Dry run against the real tree: `75 total · 27 ruled ship · 16 components on disk, 1 with a charter · 150 turns · 30670 bytes rendered, 0 placeholders remaining`.
- `git diff --name-only origin/main...HEAD -- docs/standards/` → `docs/standards/architecture/research/candidates.md` only. **Vendored standards untouched.**
- `sprint.md` is neither read nor written by anything here, and is now unreachable by pattern as well as by name.
- `docs/file_structure.txt` set-equal to the tree (no files added or removed by this pass).


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** **The charter's content boundary is drawn at "scope, not sequence"** — what the component is, what it is not, provenance, dependencies; nothing about phases, milestones or hours. Alternatives: (a) create only the directory plus its `research/` pool and no file at all; (b) create a roadmap with placeholder phase headings for `plan-feature` to fill. Why: (a) is the empty shell the Documentation Standard § 0 rule 1 forbids *and* is literally what the pipeline already produced at `docs/development/fleet-reliability/` — five verified papers under a component no document defines; (b) is a stub doc, forbidden by the same rule, and it pre-decides the phase count. The charter is the one thing neither downstream child can supply for itself: `plan-feature` plans phases *inside* a scope and does not choose the scope, and research is commissioned *per component pool* so it needs the component to already mean something. **This is the decision most worth a second opinion** — a reviewer who thinks a phaseless roadmap is a stub would reach a different design, and C-078 exists because the vendored standard can be read either way.

- **[Low]** **The working set is EVERY `ship` row, not the recently-triaged ones**, despite the brief's phrasing ("items recently triaged"). Alternatives: track scaffolding state in a column; infer recency from the diff. Why: there is no column recording that a candidate has been scaffolded — `status` belongs to a later process and this workflow may not write it — so "recently triaged" is not a fact available to anybody here, and inventing a column would widen the authorization the brief said not to widen. Re-examining the whole shipped set makes the workflow idempotent instead: a candidate whose component already exists needs nothing, the run says so, and nothing is written. The check is a directory listing, so the cost is negligible. **Flagged because it is a deliberate reading-against the brief's wording in service of its constraint.**

- **[Medium]** **`phase_planning_in` matches a claim SHAPE (`phase\d`, a figure with a unit), not a vocabulary.** Alternatives: forbid the word `phase` outright; make this a JUDGEMENT row with no mechanism. Why: the prompt *requires* the charter to say what it is not — "phases are not planned here; `plan-feature` writes them" — so a word-match would fail every correct run, and the obvious response to that is to delete the guard. Mutation M3 confirmed it: widening the pattern to the bare word turned 4 tests red, including the prompt's own charter template. Rejected the JUDGEMENT option because this is *the* boundary the whole child is defined by; leaving it on prose is the state the authorization registry was built to catch.

- **[Medium]** **The shell check is scoped to directories the run ADDED, not to the tree.** Alternatives: sweep every component. Why: measured — 16 of 17 components in this repo have no `roadmap.md`, so a tree-wide sweep fails every dispatch over work nobody in it did, which is the shape of guard that gets deleted rather than fixed. Mutation M4 pins the scope.

- **[Medium]** **The research step's signal moved from the sprint diff to the charter diff, deleting `new_sprint_sections` and `component_dir`.** Alternatives: keep the sprint sweep alongside the new one; invent a separate NEW-component signal. Why: with `plan-sprint` behind the research step, nothing ahead of it adds a sprint heading — the old sweep is dead by construction, not merely quiet, and the parent's own docstring said so. Keeping a dead signal is maintaining an interim past the interim. Deleting them is in scope: the brief explicitly names "making the research step's input reachable again."

- **[Medium]** **`candidate_decisions` and `rulings_this_run_had_no_right_to` promoted back to `plan_activities`.** Alternatives: duplicate them in `plan_candidates_activities`; import across sibling workflow folders. Why: §10.1 rule 3 is a consumer count and mechanically stated — *"consumer count decides, never taste"* — and it moves in both directions. Duplicating is the exact drift `normalise_cell` exists to record: two hand-written readings of one cell, disagreeing about whether a row is ruled.

- **[High]** **The loop-back still goes only to `plan-sprint`, skipping the new child.** Alternatives: loop through scaffolding too. Why: a second scaffolding pass would re-examine a structural decision the reviewer is holding the PR *on*, and a component directory is durable in a way a table row is not. The assertion was widened from "does not re-run triage" to "re-runs neither child ahead of plan-sprint", because adding a child to the front of a pipeline whose loop-back deliberately skips the front is precisely where an extra dispatch per loop gets added unnoticed.

## Deferred Work

- **A clause in the Documentation Standard § 0 stating that a component's `roadmap.md` may be chartered before its phase docs exist** — Why deferred: `docs/standards/documentation/` is a **vendored (MIRROR)** copy and must not be edited in this repo; an amendment goes upstream to MDC-Master-Planning and re-vendors. It is also a PROPOSAL (capability added to a document), which per `finding-routing.md` § 4 makes it a candidate rather than an issue whatever its done-state. Tracked at: `docs/standards/architecture/research/candidates.md`, row `C-078`, `decision` blank. Verified by: `git show origin/build-draft/plan-candidates:docs/standards/architecture/research/candidates.md | grep -oE "^\| C-078 \| [^|]{0,95}"` → printed `| C-078 | State in the Documentation Standard that a component's roadmap may be CHARTERED before its phas` — **on this branch; lands at merge.** ID allocated after re-reading `git show origin/main:…/candidates.md` immediately before writing, where the highest id is C-077.

Nothing else was deferred. The two findings against my own diff (a nonsense expression in `component_inventory`, a shadowed `slug`) were fixed at authoring time rather than recorded.

## Post-Run Reflection

**Friction:**

- **The brief's phrase "items recently triaged" describes a state nothing can observe.** There is no column recording that a candidate has been scaffolded, and the brief separately (and correctly) forbids writing `status`. The two constraints together mean "recently" is unavailable, so the design had to substitute idempotence. This resolved cleanly, but it cost a real design detour and the brief could have said "every `ship` row, idempotently" in four words.
- **The brief's own framing pointed at a bigger finding than it claimed.** It says the guide's ordering error "is what makes `plan_project`'s research step inert." True — and the same error had already produced a component with five researched papers and no planning document (`docs/development/fleet-reliability/`). That artifact turned out to be the strongest available argument for the whole design, and it was found by listing the tree rather than by anything the brief pointed at. Worth noting that *verifying the brief's asserted facts* is what surfaced it.
- **`grep` on this machine resolves to a `ugrep` shim, not GNU grep** (`type -a grep` → "grep is a function"). It rejects a bare glob that matches nothing with a *warning* rather than an error, which reads as "no results". Worth knowing before verifying anything about a command-line tool's behaviour.

**Project-level suggestions (this repo):**

- **16 of 17 components under `docs/development/` have no `roadmap.md`, and one (`temporal-integration/`) holds a standalone named doc — the exact anti-pattern Documentation Standard § 0 names.** This PR does not retrofit any of them (out of scope, and `plan-candidates` may not edit an existing roadmap), but the tree is substantially non-conformant with the standard the new workflow is built to satisfy. The `plan-candidates` prompt does treat the *shell* case — a component with research and no charter — as the cheapest available repair, so the pipeline will chip at it. The bare-directory and standalone-doc cases have no owner. Not filed anywhere: it is a state of the tree rather than a decision, and it needs an operator ruling on whether retrofitting is wanted at all before it has a done-state.

**Tooling-level suggestions (claude-dot-files):**

- **The prompt's mutation-control section now asks for a predicted failure count and a named set, and that combination is what made these controls worth running.** Predicting *which* tests go red — not just how many — caught nothing this time (all five hit), but M1's prediction required reasoning about guard *ordering* inside the workflow (the deletion check runs before the edit check, so one predicted-green test stayed green for a non-obvious reason). A count alone would have been satisfiable by accident. Worth keeping the "name them" half explicitly rather than letting it collapse back to a number.
- **The detached-HEAD warning fired again and was correct.** `git rev-parse --abbrev-ref HEAD` returned `HEAD` at Stage 5, exactly as the prompt's six-consecutive-reflections note predicted. The note is doing its job; no change needed, just confirming the seventh data point.

Pumapumapumas:
# build-refine pass — Decision Log and Post-Run Reflection

## Decision Log

Low-confidence first.

- **[Low]** **The `sprint.md` filename divergence was SURFACED for a ruling, not resolved.** Alternatives: (a) switch the charter to `<slug>/<slug>.md`, satisfying `sprint.md` and matching 11 of 16 components; (b) keep `roadmap.md` silently, as the draft did. Why: (b) is the failure `engineering-quality.md` § *Best practices vs project standards* names outright — silent selection between a documented standard and a different approach. (a) is defensible and may well be right, but it contradicts the two documents that specify the layer (`workflows.md`'s three-layer model and Documentation Standard § 0), and it would change this workflow's primary artifact, the parent's diff signal and the guide, inside a correction pass. **Either resolution requires an operator surface I may not write** — `sprint.md` is human-only. So I fixed everything downstream of the question (the inventory now marks `<slug>.md`-defined components ALREADY DEFINED, so nothing gets duplicated while it is open), wrote the divergence into the module docstring where the next reader hits it, and handed the placement up. **This is the item most worth a second opinion**, and a reviewer who reads `sprint.md:28` as binding on this workflow reaches option (a).

- **[Low]** **The phase/hours guard stays strict and the fix went into the prompt.** Alternatives: narrow the match to structural positions (headings, list items, table cells) so an inline cross-reference cannot trip it; exclude the `## Where it came from` section from the scan. Why: both weaken the one check the whole child is defined by, and neither is safe — a planned phase appears in exactly the structural positions the first option would keep watching *and* in the prose the second would stop watching. A guard that fails a compliant run is a real defect, but the compliant run has an alternative (paraphrase, cite by id) and the guard does not. The residual risk is stated rather than removed: if a run trips it, the failure is safe — nothing merges wrong, it costs a dispatch. Both halves are pinned by tests so the trap cannot silently move away from the prompt that warns about it.

- **[Medium]** **`MAX_NEW_COMPONENTS = 3`, reporting and stopping rather than researching a prefix.** Alternatives: research the first N and note the rest; no cap. Why: a partial fan-out leaves some new components with evidence and some without, and that difference reads as a decision somebody made rather than a ceiling. No cap was the draft's position and was defensible while the sweep was empty by construction — this branch is what made the loop live, so bounding it is in scope rather than adjacent. 3 is a judgement, not a measurement, and the note says what to do if the components are right.

- **[Medium]** **`FORBIDDEN_PATHS` widened to the whole directory rather than the phase-doc pattern being extended.** Alternatives: add `^docs/development/[^/]+/[^/]+\.md$` and the other missed shapes. Why: enumerating filenames is what produced the gap — the pattern was written against `phase{N}_*.md` and this repo's phase docs are named `<slug>/<slug>.md`. A rule that must be updated whenever somebody names a file differently is a rule that silently stops covering things. Forbid the layer, exempt the one output. It also matches `triage-candidates`, which runs one step earlier in the same worktree.

- **[Medium]** **Rejected the reviewer's assertion that `<component>/research/` must be permitted**, and inverted the test that carried it. The draft's test docstring argued the parent's research children write there on the same branch and in the same worktree, so a pattern catching them would break the pipeline one step later. Why rejected: both snapshots are taken around **this workflow's own** `run_claude` call, and the research children run after it returns — so nothing they write is inside the window. The reasoning was a timing error, and permitting the pool bought nothing while leaving the run free to rewrite evidence it did not gather.

- **[Medium]** **`COMPONENT_ROOT` was promoted rather than the child's copy being deleted in favour of an inline literal**, and it changed type from `Path` to `str`. Why: three of its four consumers build a repo-relative snapshot KEY from it, and `f"{Path(...)}"` makes that key format platform-dependent — a subtlety a snapshot key cannot afford. The one consumer that joins it onto a tree works unchanged, since `Path / str` is fine.

- **[High]** **Six items were surfaced without a tracker rather than filed or fixed.** Why: each is either a shared-family issue this child merely exposes first (the poisoned-PR-on-guard-failure shape is identical in `triage-candidates` and `plan-sprint`), a pre-existing pattern in files this PR does not own, or a design question that belongs with `plan-feature`'s scope. **I did not file issues for them** — several are proposals rather than defects, and `finding-routing.md` § 4 routes those to `candidates.md`, but placing six rows from one correction pass is the "four separate issues against one file" shape the clustering rule forbids. They are in the PR body with no pointer at all, which is the honest state.

## Deferred Work

Nothing was deferred with a pointer. Six items are **SURFACED** in the PR body under *Surfaced — real, out of scope, no verified home*, deliberately with no tracker: I could not verify a durable home for any of them, and an honest "no home for this" is worth more than a plausible pointer.

The draft's one deferral — the Documentation Standard § 0 clause — stands and its placement is now correct. **Verified by:** `grep -n "PR #88 (plan-candidates)" docs/standards/architecture/research/candidates.md` → 2 hits, the C-078 row and the § Where things stand paragraph, **on this branch; lands at merge.** The row's Source cited `PR #87` in both places, which is a different open PR (*"plan-revision: plan the persistent memory protocol as its own component"*, confirmed with `gh pr view 87 --json number,title`); corrected to #88. `decision` remains blank. Id re-checked against `origin/main` immediately before pushing: highest there is C-077, and two commits had landed since the branch point without taking it.

## Post-Run Reflection

**Friction:**

- **The three narrow lenses and the holistic one found disjoint defect sets, and the two that would actually have hurt came from opposite ends.** `code-reviewer` found the guard collision by reading the live candidates file against a regex; `quality-control` found the shell misclassification by counting what was in `research/` directories. Neither is visible in the diff, and neither reviewer could have found the other's. The parallel-then-sequential split earned its cost here rather than merely being followed.
- **Every single finding was at a seam between the module and the repository — not one was internal.** The draft's own discipline ("counted in code, authoritative — do not recount"; "read the artifact, never a child's account") was applied rigorously to the model and never once to the author. Four confidently-worded factual claims were false, and `ls docs/development/` would have caught all four. Worth naming because it is not a rushed-work signature: the reasoning prose is excellent and voluminous, and that is where the budget went instead of observation.
- **The `--dry-run` path computes the disproof and discards it.** It already calls `component_inventory` against the real tree and prints two integers. Printing the block would have surfaced the shell misclassification, the eleven ALREADY-DEFINED components and both count errors before the PR existed. Not changed here — it widens the runner's contract — but it is the cheapest available conversion of a habit into a mechanism.
- **Mutating a constant, rather than a code path, is what exposed my own vacuous test.** The ceiling discriminator sized its fixture from `MAX_NEW_COMPONENTS`, so at 0 it asserted `0 == 0` while the step was dead. Reading it would never have shown that; only setting the constant to 0 did.

**Project-level suggestions (this repo):**

- **`docs/development/sprint.md` and `docs/guide/workflows.md` specify the component layer differently, and 11 of 16 components follow the one no workflow implements.** This is the RULING-REQUIRED item above. It is not filed anywhere: `sprint.md` is operator-only, and the question needs a ruling on which document is authoritative before it has a done-state.
- **A guard that raises after its model has already opened the PR leaves the PR open, unlabelled and uncommented.** All three planning children share the shape. The guard messages are already written for a human and already end in `see {url}` — they just never reach the surface the operator looks at.

**Tooling-level suggestions (claude-dot-files):**

- **The prompt's instruction to mutate what *you* added, not only what you inherited, paid out directly and would not have been done otherwise.** Six of my mutations targeted my own additions; one found a real defect in a test written twenty minutes earlier. The asymmetry the instruction names is exactly right — the draft's tests had at least run against a real implementation, mine had not.
- **"Doubt your own invocation before you doubt the claim" fired early and correctly.** A chained `cd ... && cp` silently skipped the restore because the CWD had persisted from the previous call, so mutation B ran compounded on mutation A and reported 7 failures instead of 4. Catching that as a shell error rather than writing it up as a finding about guard interaction is the whole value of the instruction — and the CWD-persistence warning elsewhere in the same prompt is what pointed at it.
- **The `gh pr view --json comments` → file instruction was load-bearing and correct.** The draft's reflection is ~9.6 KB and carried the C-078 provenance claim that turned out to name the wrong PR — a finding that exists only in the comment, not in the diff.

Pumapumapumas:
# `review-pr` disposition — pass 1

**Verdict: HOLD — needs-assistance.** This is strong work and I am not re-litigating the guard machinery: the refine pass's ten fixes are real, I verified every one against the code rather than the account, and the mutation controls are the best evidence in this PR. **Six defects survive, all at the same seam the refine pass named — between this module and the repository it runs in — plus two questions only the operator can rule on.**

The headline is that **the entrypoint this PR authored cannot do the one thing its own usage block documents.** `./plan_candidates.sh --pr 42` cuts its worktree from local `HEAD`, not from the PR's branch. The correct pattern is eight characters away in `run_plan_revision.py:141` and in the parent this PR wired itself into (`plan_project_workflow.py:106`), and the run copied the two broken siblings instead. The PR body discloses it as *"pre-existing family-wide pattern"* — honest, and it does not apply: `run_plan_candidates.py` is a **new file in this diff**, so this is the run's own artifact and the label grants nothing.

**Mergeability, checked first and checked properly.** `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`; `git ls-remote origin refs/pull/88/merge` resolves to `aaa2622` (a live merge ref, so the `pull_request` merge gate could compute one); four checks green including `suite` at 2m35s. The check SET matches the previous head's — no dropped gate. **The merge gate genuinely ran on this PR.**

---

## Coverage of the mandatory sweeps — said explicitly so a later pass can tell verification from assumption

| Sweep | Result |
|---|---|
| **Is this body about THIS PR?** | Yes. Every artifact the body names — `plan_candidates.sh`, `run_plan_candidates.py`, `plan_candidates/`, the `config.yaml` entries, the guide correction, C-078 — is present in the diff or the tree. |
| **Completion-checkbox sweep** | **Zero flips.** `git diff origin/main...HEAD \| grep -E '^\+.*\[[xX]\]'` returns nothing. No planning artifact gained an `[x]`. `sprint.md` untouched — confirmed by name *and* by pattern (`(^\|/)sprints?\.md$` in `FORBIDDEN_PATHS`). |
| **Merge-commit sweep** | Neither commit is a merge (`7fa3463`, `c3c4ee8`, linear). No conflict-resolution deletion channel. |
| **Deleted-artifact sweep** | Two functions deleted. Enumerated by name against `origin/main`: `new_sprint_sections` §non-model-observable→`scaffolded_components`; §`base_ref`-required-no-default→**carried verbatim**; §added-only-match→`--diff-filter=A`; and `-z` was *added* where the original had no need of it. `component_dir` §`tree`-not-`repo_root`→carried; §pool-inside-component→carried and now cites Research Standard §1; §slug-derivation-with-`ValueError`→**legitimately dissolved**, the slug now comes off a filesystem path so no derivation and no empty-slug case exists. **Nothing landed NOWHERE.** |
| **Table rendering** | `candidates.md` rendered through `POST /markdown` in full, not sliced: **75 of 75** `C-NNN` rows render as `<td>` cells; the file's own prose declares thirty untriaged over C-047–C-078 and `test_candidates_prose_matches_the_table.py` derives that from the table. No orphaned row. |
| **Post-cut rule check** | Two commits landed on `main` after the cut (`ba677d2`, `c8c61b2`); `git merge-base --is-ancestor` says NOT ancestor for both. Neither is rule-bearing on this PR (a `plan_revision` prompt clause; `block-dangerous.sh` + `hook-scripts.md`). **This PR is not judged against either.** |
| **Reflection coverage, per-commit** | Complete. `7fa3463` (`wip:`) attested by the draft Decision Log; `c3c4ee8` by the refine Decision Log. Both substantive, neither a placeholder. |
| **Predecessor deferrals (#85)** | One trigger **fires** — see `turn-cap-authority-orphaned` below. #85 surfaced the turn-cap authority defect with no pointer *"so it is picked up downstream"*; this PR is downstream and adds the third orphaned cap. |
| **Pass number** | **DERIVED, not taken from the label.** `gh pr view 88 --json comments \| grep -c 'pr_review:'` → **0**. This is pass 1; the dispatch label also says 1. **No divergence.** |

---

## Disposition table

|---|---|---|---|
| `redispatch-worktree-cut-from-head` | correctness | **HOLD** (redispatch) | `run_plan_candidates.py:58` passes the literal `"HEAD"` to `worktree_add` regardless of `--pr`. The shim's own usage block documents `--pr 42`. `run_plan_revision.py:141`, `build_workflow.py:50`, `build_minor_workflow.py:40` and `plan_project_workflow.py:106` all do it correctly. **New file, this PR's artifact** — the "pre-existing family pattern" label does not reach it. |
| `charter-content-unchecked` | correctness | **HOLD** (redispatch) | Verified two ways: `roadmaps_created` feeds only `phase_planning_in`, which is negative-only, and `shells_without_a_charter` tests existence alone. A one-line `roadmap.md` passes **every** guard, and the parent then spends `research-write` + `research-verify` against it. That is `docs/development/fleet-reliability/` recreated through the workflow built to stop it. |
| `guard-failure-leaves-poisoned-pr` | correctness | **HOLD** (redispatch) | Every guard raises *after* the model has committed, pushed and opened the PR. `_dispose` is never reached; `plan_project_workflow.py` has no `try` around any child call. The PR stays open with the violating charter, no comment, no label. |
| `charter-boundary-never-read-back` | correctness | **HOLD** (redispatch) | The prompt calls *"What this is NOT"* the section that stops the next candidate landing in the wrong place. Verified absent two ways: Stage 1's read list (prompt L74–79) names four files and no charter, and `component_inventory` emits name + status mark only, never scope text. |
| `fanout-note-claims-a-hold-that-never-happens` | correctness | **HOLD** (redispatch) | Over the ceiling, the note reads *"NO COMPONENT RESEARCH RAN, and this is a HOLD rather than a skip"* — then the run continues to `plan-sprint` and `_dispose`, and `run_plan_project.py:97` prints that note beside whatever verdict `review-pr` returned, which may be `MERGE`. |
| `rule3-audit-omits-evidence-block` | doc-drift | **HOLD** (redispatch) | `evidence_block` (`plan_activities.py:498`) has exactly one consumer (`plan_revision_workflow.py:117`) and is absent from the rule-3 audit this PR rewrote — the same omission the docstring records having made once and claims it *"counted rather than eyeballed the second time."* |
| `charter-unreachable-by-any-loop-back` | correctness | **HOLD** (needs-assistance) | Verified end to end: `plan_candidates.roadmaps_edited` forbids revising a roadmap, so a `--pr` pass cannot fix its own charter; the loop-back calls **only** `sprint.run_plan_sprint`, whose `FORBIDDEN_PATHS` is `^docs/development/` and whose `permitted_paths()` returns the sprint plan and `candidates.md` only. The charter is unreachable from every automated path. |
| `component-filename-convention-unruled` | standards-implication | **HOLD** (needs-assistance) | The run's own RULING REQUIRED item, and it is correctly surfaced rather than settled. `sprint.md:24,28` (human-only) and Documentation Standard § 0 / `workflows.md` specify different first files. Dry run confirms the tree: **16 components, 1 with a charter**. |
| `turn-cap-authority-orphaned` | standards-implication | **ESCALATED** | `workflow-scripts.md:490` makes `docs/guide/workflows.md` authoritative for turn caps; that guide contains no turn-cap table for any temporal workflow (`grep -n 'max_turns\|turn cap'` → zero hits), while `act.max_turns()` reads `config.yaml`. Live on `main`; this PR adds the third orphaned cap. **Does not hold this PR.** |
| `c078-proposal-placed` | deferral | **DEFERRED** | Pointer VERIFIED in the diff, not in the thread: `candidates.md` gained row `C-078` with `decision` blank, Source `PR #88`, and the *Where things stand* prose moved 29→30 / C-077→C-078. Correct destination for a PROPOSAL against a vendored standard per `finding-routing.md` § 4. |
| `research-pool-permission-rejected` | scope | **REJECTED** (upheld) | The run inverted a draft test that asserted `<component>/research/` must be permitted. **Verified independently:** both snapshots are taken at `plan_candidates_workflow.py:269` and `:308`, either side of the single `act.run_claude` call at `:286`; the research children are invoked at `plan_project_workflow.py` step 3, after `run_plan_candidates` returns. Nothing they write is inside the window. The draft's reasoning was a timing error and the rejection is right. |
| `guard-helper-extraction-rejected` | scope | **REJECTED** (upheld) | I checked *how* the three guard blocks differ rather than only *that* they do: the shape `if offences: raise RuntimeError(f"…")` is shared, but each `raise` carries a distinct multi-sentence message that IS the guard's content — this is not one implementation typed three times with a parameter unset. Also out of scope per the build-review boundary: "could this be structured better" belongs to planning. |
| `activities-split-rejected` | scope | **REJECTED** (upheld) | Nine functions, one consumer each, one layer, and `component_inventory` shares `component_dirs` with the guards — splitting would separate a helper from its only caller. Structural preference, out of scope for a build review. |
| `forbidden-paths-was-a-false-enforcement-claim` | correctness | **FIXED** (verified) | `FORBIDDEN_PATHS` is now `^docs/development/` less one `roadmap.md` exemption. Verified by running `test_plan_candidates_forbids_phase_docs_while_permitting_the_roadmap_beside_them`: 11 forbidden paths caught (including `<slug>/<slug>.md`, `cpi-decisions.md`, `reviews/`, a research pool), 2 permitted paths clear. |
| `inventory-invited-chartering-defined-components` | correctness | **FIXED** (verified) | `has_evidence` now counts `raw/*.md` and `synthesis.md` rather than `is_dir()`, and `<slug>.md` is a first-class ALREADY-DEFINED state that tells the run not to charter. Verified by reading `component_inventory` and by the live dry run. |
| `phase-guard-trips-on-quoted-candidate-titles` | correctness | **FIXED** (verified) | The guard stayed strict and the prompt now names both live traps (`C-011`'s `(~9 h)`, `C-013`'s `Phase 4`) with rewrite examples. **I ran the guard against the prompt's own charter template:** exactly one fenced `markdown` block, **zero** `_PHASE_PLANNED` / `_HOURS_ESTIMATED` hits. A compliant run cannot fail its own guard. |
| `unbounded-research-fanout` | correctness | **FIXED** (verified) | `MAX_NEW_COMPONENTS = 3`, reporting and stopping rather than researching a prefix. Pinned both ways by `test_a_runaway_scaffolding_pass_RESEARCHES_NOTHING_and_says_why` and `test_a_fanout_AT_the_ceiling_still_researches`. |
| `component-root-declared-twice` | correctness | **FIXED** (verified) | One definition, `str`, at `plan_activities.py:490`, with `NOT_A_COMPONENT` beside it. `plan_candidates_activities.py:39` and `plan_project_activities.py:35` both read it; `existing_work` no longer hardcodes either. |
| `scaffolded-components-quoting-and-depth` | correctness | **FIXED** (verified) | `-z`, `_CHARTER_DEPTH` derived from `COMPONENT_ROOT` rather than a literal `4`, and the `NOT_A_COMPONENT` exclusion applied. I traced the trailing-NUL case: `"".split("/")` has length 1 ≠ 4, so it drops out cleanly. |
| `charter-template-test-was-a-drifted-hand-copy` | test-gap | **FIXED** (verified) | The test now extracts the fenced block from the prompt and asserts there is exactly one. **I re-ran that extraction independently and got exactly one block** — the assertion is not vacuous. |
| `four-false-factual-claims` | doc-drift | **FIXED** (verified) | Against the live tree: `plan_candidates.sh --dry-run` prints `16 components on disk, 1 with a charter`, matching the corrected "15 of 16". `memory-management-framework/roadmap.md`'s zero hour estimates and the `before_dirs` escape clause are both corrected in place. |
| `c078-source-cited-the-wrong-pr` | doc-drift | **FIXED** (verified) | `git diff` of `candidates.md` shows `PR #88 (plan-candidates)` in both the row's Source and the *Where things stand* prose. No `#87` remains. |
| `temporal-integration-enumeration-stale` | doc-drift | **FIXED** (verified) | `temporal-integration.md:5` now reads `…triage-candidates split out of it in PR #85, and plan-candidates added in PR #88…`. |

**Verdict rationale:** the guard machinery is sound and independently confirmed, but six live defects remain — one deterministic (the `--pr` entrypoint), one that recreates the exact failure this workflow exists to prevent (an unchecked charter becoming a research brief) — and two questions (`sprint.md`'s filename convention, and the fact that nothing in the fleet can correct a charter) are genuinely the operator's to rule on rather than mine to guess.

---

## WHAT HAPPENS NEXT — the runway

**Two operator rulings, then one dispatch closes everything else.** Nothing here re-opens a decision the refine pass made.

### 1. `component-filename-convention-unruled` — needs a ruling before more components are chartered

> **reframe:** not *"which filename is prettier"* but *"which document is authoritative for the component layer, given one of them is the operator's own file that no dispatch may edit?"*
> **bp:** where two specifications conflict, the correct move is to make one authoritative and amend the other — never to let an implementation silently pick a side, which `engineering-quality.md` § *Best practices vs project standards* names as the failure outright.
> **recommendation:** rule that `roadmap.md` is the charter's filename and amend `sprint.md`'s two sentences to say a chartered component starts with a `roadmap.md` and a single-phase component may still use `<slug>/<slug>.md`. Reasoning: `roadmap.md` is what BOTH specification documents say, it is what the parent's diff signal reads, and `sprint.md`'s *"one phase needs no roadmap"* rule assumes a phase count that does not exist at charter time. The alternative — switching the workflow to `<slug>/<slug>.md` — contradicts two documents and changes this PR's primary artifact, the parent's signal and the guide inside a correction pass. **`sprint.md` is human-only under `standards-governance.md`; I have not touched it and neither should any dispatch.** The run's downstream handling is already correct — `component_inventory` marks `<slug>.md` components ALREADY DEFINED, so nothing duplicates while this is open.

### 2. `charter-unreachable-by-any-loop-back` — needs a ruling on where the correction path lives

> **reframe:** not *"should the loop-back re-run plan-candidates"* — the Decision Log rejected that for a good reason — but *"this PR introduces the fleet's first DURABLE FILESYSTEM artifact that no workflow can subsequently correct; where does that capability live?"* Everything the loop-back previously could not fix was a table cell.
> **bp:** fail loud on unreachable state rather than degrading silently. Spending three Opus dispatch passes on a target that provably cannot touch the file is the silent-degradation shape `engineering-quality.md` forbids.
> **recommendation:** give the capability to `plan-feature` when it lands — it is the workflow that legitimately owns roadmap content, so a charter correction is naturally its scope — and, in the meantime, have the parent state in its run note that a runway naming a chartered file is not addressable by the loop-back target, so the operator edits it directly instead of watching three passes fail. I am **not** recommending relaxing `roadmaps_edited`: it is the boundary the whole child is defined by. This does not block the PR from being useful today (a human edits the file in the open PR); it blocks the automated loop from ever closing that runway, which is a design gap worth deciding deliberately rather than discovering on the first charter defect.

### 3. One `build.sh --pr 88` dispatch closes the remaining six

Sized `build.sh` rather than `build-minor.sh`: two of the six add new guard code with new tests (`charter-content-unchecked`, `guard-failure-leaves-poisoned-pr`), and new mechanism in this family has earned a fresh-context review every time.

1. **`redispatch-worktree-cut-from-head`** — make the ref depend on `--pr`, exactly as `run_plan_revision.py:141` does.
2. **`charter-content-unchecked`** — a positive content check over `roadmaps_created`: the created file must carry the charter template's four sections.
3. **`guard-failure-leaves-poisoned-pr`** — surface a guard `RuntimeError` on the PR before it propagates. The messages already end in `see {url}`.
4. **`charter-boundary-never-read-back`** — put each existing charter's *What this is NOT* into `component_inventory`.
5. **`fanout-note-claims-a-hold-that-never-happens`** — say what actually happened; do not claim a verdict the run does not produce.
6. **`rule3-audit-omits-evidence-block`** — name it in the audit.

---

## ESCALATED — live on `main`, not this PR's, does not hold it

**`turn-cap-authority-orphaned`** — `docs/standards/workflow-scripts.md:490` on the default branch says *"The authoritative current values live in `docs/guide/workflows.md` alongside the workflow inventory — keep them there so they don't drift across two locations."* That guide has **no turn-cap table for any workflow under `scripts/workflows/temporal/`** (verified: `grep -n 'max_turns\|Max turns\|turn cap' docs/guide/workflows.md` → zero hits), and `act.max_turns()` reads `config.yaml`.

- **Where it lives:** `docs/standards/workflow-scripts.md:490`, on `origin/main`.
- **Why this PR does not own it:** the PR places its cap correctly, in the file the code actually reads, with a documented basis and a revise-from-measurement note. The defect is the standard's pointer, which predates this branch — PR #85 surfaced it with no pointer explicitly so it would be *"picked up downstream."*
- **Remedy, one line:** amend § *Max Turns Per Script* to name `config.yaml` as authoritative for the Python fleet and `docs/guide/workflows.md` for the bash fleet, then delete the sentence that sends a reader to a table that does not exist. Human-gated (`standards-governance.md`), and a PR against this repo satisfies that gate.

---

```yaml
pr_review:
  pr: 88
  run_id: 5ad63536744d45fd8d2cd51fc035bef4
  pass: 1                            # DERIVED: `gh pr view 88 --json comments | grep -c 'pr_review:'` -> 0
                                     # fence-anchored blocks on this thread. Dispatch label also says 1.
                                     # NO DIVERGENCE.
  attempt: 2                         # Seeded from ATTESTING reflection comments on the PR: the draft
                                     # Decision Log (2026-08-13T22:30Z) and the build-refine Decision Log
                                     # (23:11Z). The `wip:` commit belongs to the attempt that attested it.
  verdict: HOLD
  converged: false
  findings:
    - id: redispatch-worktree-cut-from-head
      title: the documented `--pr` entrypoint pushes a tree that does not contain the PR's work
      category: correctness
      consequence: >
        `./plan_candidates.sh --pr N` cuts its worktree from local HEAD, so the model is told to push
        to PR #N's branch from a tree that lacks that PR's commits — the push is rejected
        non-fast-forward, or clobbers. Worse, `before_roadmaps` is then snapshotted against a tree
        missing the charter an earlier pass wrote, so re-creating that charter reads as a legitimate
        creation and `roadmaps_edited` — which judges only keys present on both sides — cannot see it.
        The redispatch path this workflow advertises does not work.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: charter-content-unchecked
      title: a one-line roadmap passes every guard and then becomes a research brief
      category: correctness
      consequence: >
        `roadmaps_created` feeds only `phase_planning_in`, which tests exclusively for what must NOT
        be present, and `shells_without_a_charter` tests only that the file exists. A stub charter
        therefore passes the whole guard set, and the parent immediately spends a research-write plus
        a research-verify dispatch against it. That is `docs/development/fleet-reliability/` — five
        verified papers under a component no document defines — reproduced through the workflow built
        to prevent it, and this time with a file present so nothing looks wrong.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: guard-failure-leaves-poisoned-pr
      title: a guard failure leaves an open PR carrying the violating charter with nothing on it saying so
      category: correctness
      consequence: >
        Every guard raises after the model has committed, pushed and opened the PR. `plan_project_workflow.py`
        wraps no child call in a `try`, so the RuntimeError kills the dispatch before `_dispose` runs.
        The PR stays open, unlabelled and uncommented, with the violating charter in it — and the next
        thing to read that PR (an operator, gh-monitor, a later dispatch) sees an ordinary PR awaiting
        review. This PR is the first child that can fail AFTER a sibling has already opened the PR,
        which is what makes the shared shape newly consequential here.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: charter-boundary-never-read-back
      title: the section the prompt calls load-bearing is written and never read, so the next run cannot honour it
      category: correctness
      consequence: >
        The prompt calls "What this is NOT" the section that stops the next candidate landing in the
        wrong place. Verified absent two independent ways — Stage 1's read list names four files and no
        charter, and `component_inventory` emits a name plus a status mark with no scope text. So on the
        second `plan-candidates` run the boundary does no work: a candidate a charter explicitly
        excludes can be chartered into a duplicate domain, which is a durable directory tree that
        research and planning both write into.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: fanout-note-claims-a-hold-that-never-happens
      title: the parent prints a note asserting a HOLD beside a verdict that may be MERGE
      category: correctness
      consequence: >
        Over `MAX_NEW_COMPONENTS` the note reads "NO COMPONENT RESEARCH RAN, and this is a HOLD rather
        than a skip" — then the run sets `new_components = []` and continues to `plan-sprint` and
        `_dispose`. `run_plan_project.py:97` prints that note alongside whatever verdict `review-pr`
        returned. An operator reading "HOLD" believes the pipeline stopped for their ruling when
        plan-sprint has already sequenced the unresearched components and the PR is up for merge.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: rule3-audit-omits-evidence-block
      title: the rule-3 audit is incomplete again, on the file that states the rule
      category: doc-drift
      consequence: >
        `evidence_block` (`plan_activities.py:498`) has exactly one consumer
        (`plan_revision_workflow.py:117`) and is absent from the §10.1 rule-3 audit this PR rewrote.
        That docstring presents itself as the authoritative audit and records having made this exact
        omission once before, claiming it "counted rather than eyeballed the second time" — so the
        next promotion decision reads an incomplete baseline as complete, which is how C-072's
        four-drifted-copies shape starts.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: charter-unreachable-by-any-loop-back
      title: a charter defect burns the full loop-back budget on a child that cannot reach the file
      category: correctness
      consequence: >
        `roadmaps_edited` forbids plan-candidates revising a roadmap, so a `--pr` pass cannot correct
        its own charter; the loop-back calls only `plan_sprint`, whose FORBIDDEN_PATHS is
        `^docs/development/` and whose `permitted_paths()` returns the sprint plan and `candidates.md`
        only. A HOLD_REDISPATCH naming a charter defect therefore spends up to MAX_LOOPS Opus dispatch
        passes on a provably unreachable target and then routes to a human anyway. This PR introduces
        the fleet's first durable filesystem artifact with no correction path; everything the loop-back
        previously could not fix was a table cell.
      disposition: hold
      remedy: create-missing-surface
      hold_kind: needs-assistance
    - id: component-filename-convention-unruled
      title: every component this pipeline charters entrenches a filename two specifications disagree on
      consequence: >
        `docs/development/sprint.md:24,28` (operator-only) says a component is `<slug>/<slug>.md` and
        that one phase needs no roadmap; `docs/guide/workflows.md` and the vendored Documentation
        Standard § 0 say `<slug>/roadmap.md` + phase docs. 11 of this repo's 16 components follow
        `sprint.md`; the workflow writes the other shape and `shells_without_a_charter` FAILS any run
        that writes the `sprint.md` shape instead. Every charter written before this is ruled deepens
        a split the tree already has, and only one of the two resolutions is reachable by a dispatch.
      disposition: hold
      hold_kind: needs-assistance
    - id: turn-cap-authority-orphaned
      title: the standard sends a maintainer setting a turn cap to a table that does not exist
      consequence: >
        `workflow-scripts.md:490` makes `docs/guide/workflows.md` authoritative for turn caps and tells
        the reader to keep them there "so they don't drift across two locations". That guide has no
        turn-cap table for any temporal workflow, while `act.max_turns()` reads `config.yaml`. A
        maintainer following the standard either invents a value or edits a table nothing reads, so a
        cap set in good faith has no effect on the run it was set for. This PR adds the third orphaned
        cap (`plan-candidates: 150`, after `triage-candidates: 200`).
      disposition: escalated
      escalation_location: >
        docs/standards/workflow-scripts.md:490 on origin/main. This PR does not own it: it placed its
        cap correctly in the file the code reads, with a stated basis and a revise-from-measurement
        note. The defect is the standard's pointer, which predates the branch — PR #85 surfaced it with
        no pointer explicitly so it would be picked up downstream. Merging #88 does not change its
        severity, so holding this PR on it buys nothing.
    - id: c078-proposal-placed
      title: the vendored-standard amendment this run surfaced has a verified home and is not buried
      category: deferral
      consequence: >
        A PROPOSAL against a vendored MIRROR cannot be written in this repo, so a surfaced-and-unplaced
        one dies at merge. It was placed, not merely surfaced.
      disposition: deferred
      pointer: >
        docs/standards/architecture/research/candidates.md row C-078, `decision` blank, Source
        "PR #88 (plan-candidates)". Verified by reading the committed diff of that file on this branch,
        not the run's account: the row is present and the "Where things stand" prose moved 29->30 and
        C-077->C-078 consistently. Correct destination per finding-routing.md § 4 (a proposal is a
        candidate, never an issue, whatever its done-state).
      pointer_verified: true
    - id: research-pool-permission-rejected
      title: permitting a component's research pool would have licensed rewriting evidence this run did not gather
      category: scope
      consequence: >
        Rejection upheld on independent verification, so nothing is owed here.
      disposition: rejected
      remedy: reject
    - id: guard-helper-extraction-rejected
      title: a shared _guard helper would put a call frame between each raise and the sentence that is its content
      category: scope
      consequence: >
        Rejection upheld; and a structural preference is out of scope for a build review regardless.
      disposition: rejected
      remedy: reject
    - id: activities-split-rejected
      title: splitting plan_candidates_activities would separate helpers from their only caller
      category: scope
      consequence: >
        Rejection upheld; structural preference, out of scope for a build review.
      disposition: rejected
      remedy: reject
    - id: forbidden-paths-was-a-false-enforcement-claim
      title: the prompt claimed a boundary the pattern did not enforce over most of the directory
      category: correctness
      consequence: >
        Fixed and verified against the code: FORBIDDEN_PATHS is now `^docs/development/` less one
        roadmap exemption, and the named test catches 11 boundary cases while clearing the 2 paths the
        workflow needs.
      disposition: fixed
      remedy: none
    - id: inventory-invited-chartering-defined-components
      title: an empty mkdir was reported to the model as research evidence
      category: correctness
      consequence: >
        Fixed and verified: `has_evidence` counts `raw/*.md` and `synthesis.md`, and `<slug>.md` is a
        first-class ALREADY-DEFINED state. Confirmed against the live tree by the dry run.
      disposition: fixed
      remedy: none
    - id: phase-guard-trips-on-quoted-candidate-titles
      title: two live ship rows would have failed a compliant run after its budget was spent
      category: correctness
      consequence: >
        Fixed and verified: the guard stayed strict, the prompt names both traps with rewrites, and I
        ran both regexes over the prompt's own charter template — exactly one fenced block, zero hits.
      disposition: fixed
      remedy: none
    - id: unbounded-research-fanout
      title: an unbounded fan-out would have spent two Opus dispatches per mis-chartered component
      category: correctness
      consequence: >
        Fixed and verified: MAX_NEW_COMPONENTS = 3, reporting and stopping rather than researching a
        prefix, pinned in both directions by named tests.
      disposition: fixed
      remedy: none
    - id: component-root-declared-twice
      title: two COMPONENT_ROOT declarations with two types would have split the parent from its child
      category: correctness
      consequence: >
        Fixed and verified: one `str` definition at plan_activities.py:490 with NOT_A_COMPONENT beside
        it, read by both the child and the parent's sweep.
      disposition: fixed
      remedy: none
    - id: scaffolded-components-quoting-and-depth
      title: a non-ASCII slug would have been silently skipped and its component never researched
      category: correctness
      consequence: >
        Fixed and verified: `-z`, depth derived from COMPONENT_ROOT, and the NOT_A_COMPONENT exclusion
        applied. The trailing-NUL element drops out on the length test.
      disposition: fixed
      remedy: none
    - id: charter-template-test-was-a-drifted-hand-copy
      title: the test certifying a compliant run cannot fail its own guard was checking an approximation
      category: test-gap
      consequence: >
        Fixed and verified: the test extracts the fenced block from the prompt and asserts exactly one.
        I re-ran that extraction independently and got exactly one — the assertion is not vacuous.
      disposition: fixed
      remedy: none
    - id: four-false-factual-claims
      title: four confidently-worded claims about the tree were false and one was the design's own argument
      category: doc-drift
      consequence: >
        Fixed and verified against the live tree: the dry run prints 16 components with 1 charter,
        matching the corrected "15 of 16".
      disposition: fixed
      remedy: none
    - id: c078-source-cited-the-wrong-pr
      title: the candidate row's provenance pointed at a different open PR
      category: doc-drift
      consequence: >
        Fixed and verified in the committed diff: PR #88 in both the row's Source and the prose. No
        #87 remains.
      disposition: fixed
      remedy: none
    - id: temporal-integration-enumeration-stale
      title: the component doc's planning-family enumeration ended one workflow short
      category: doc-drift
      consequence: >
        Fixed and verified in the diff at temporal-integration.md:5.
      disposition: fixed
      remedy: none
  next_steps:
    - item: component-filename-convention-unruled
      kind: needs-assistance
      note: >
        Rule which document is authoritative for a component's first filename. sprint.md is
        operator-only; no dispatch may resolve this.
      why_human: standards
      reframe: >
        Not which filename is better, but which document is authoritative for the component layer given
        one of them is the operator's own file that no dispatch may edit.
      bp: >
        Where two specifications conflict, make one authoritative and amend the other — never let an
        implementation silently pick a side (engineering-quality.md, Best practices vs project standards).
      recommendation: |
        Rule `roadmap.md` and amend sprint.md's two sentences so a CHARTERED component starts with a
        `roadmap.md`, while a single-phase component may still use `<slug>/<slug>.md`. Reasoning:
        `roadmap.md` is what BOTH specification documents say, it is what the parent's diff signal
        reads (`scaffolded_components` matches `<root>/<slug>/roadmap.md`), and sprint.md's "one phase
        needs no roadmap" assumes a phase count that does not exist at charter time — `plan-feature`
        determines it. The alternative, switching the workflow to `<slug>/<slug>.md`, contradicts two
        documents and changes this PR's primary artifact, the parent's signal and the guide inside a
        correction pass. Do NOT let a dispatch edit sprint.md to close this. The run's downstream
        handling already prevents duplication while it is open.
    - item: charter-unreachable-by-any-loop-back
      kind: needs-assistance
      note: >
        Decide where the capability to correct a written charter lives. No workflow in the fleet has it.
      why_human: architecture-gap
      reframe: >
        Not "should the loop-back re-run plan-candidates" — that was rejected for a good reason — but
        "this PR introduces the fleet's first durable filesystem artifact that no workflow can correct;
        where does that capability live?"
      bp: >
        Fail loud on unreachable state rather than degrading silently; spending three Opus passes on a
        target that provably cannot touch the file is the silent-degradation shape engineering-quality.md
        forbids.
      recommendation: |
        Give charter correction to `plan-feature` when it lands — it legitimately owns roadmap content,
        so revising a charter is naturally its scope, and that keeps `roadmaps_edited` strict, which is
        the boundary the whole child is defined by. In the interim, have the parent state in its run
        note that a runway naming a chartered file is not addressable by the loop-back target, so the
        operator edits it directly rather than watching MAX_LOOPS passes fail. Do NOT relax
        `roadmaps_edited` to close this — that trades a design gap for a guard hole.
    - item: redispatch-worktree-cut-from-head
      kind: redispatch
      dispatch_tool: build.sh
      note: Make the worktree ref depend on --pr, exactly as run_plan_revision.py:141 already does.
      dispatch_context: |
        Fix SIX findings on this branch. Touch ONLY the files named below.

        1. redispatch-worktree-cut-from-head — scripts/workflows/temporal/scripts/run_plan_candidates.py:58.
           Replace the literal "HEAD" third argument to act.worktree_add with the conditional the family
           already uses: `ref = f"origin/{act.pr_branch(a.pr_number, repo_root)}" if a.pr_number else "HEAD"`,
           computed before the worktree_add call. Copy the shape from run_plan_revision.py:141 verbatim.
           Add a unit test asserting the ref varies with --pr.
           DO NOT touch run_triage_candidates.py or run_plan_sprint.py — they carry the same defect and
           are NOT this PR's artifacts; they are recorded here so a later pass finds them, not fixed here.

        2. charter-content-unchecked — plan_candidates_activities.py + plan_candidates_workflow.py.
           Add a POSITIVE content check over own.roadmaps_created: a created roadmap must carry the four
           section headings the prompt's charter template defines (What this is / What this is NOT /
           Where it came from / Dependencies). Derive the required set by extracting the single fenced
           ```markdown block from prompts/plan_candidates.md — the existing test already does this
           extraction, so there is exactly one definition of the template. Raise with the same message
           discipline as the guards beside it. Test both directions: the prompt's own template must
           PASS, and a one-line roadmap must FAIL.

        3. guard-failure-leaves-poisoned-pr — plan_project_workflow.py, and the runner
           run_plan_candidates.py. When a child's guard RuntimeError propagates after a PR is already
           open, post the guard message as a comment on that PR before re-raising. The messages already
           end in "see {url}" so the text is written. Do NOT swallow the exception — the run must still
           fail; this only makes the failure visible on the surface an operator looks at.
           Apply it at the parent's three child call sites, not per-child.

        4. charter-boundary-never-read-back — plan_candidates_activities.component_inventory.
           For a component marked HAS A CHARTER, include its "What this is NOT" section body in the
           rendered block. Keep it bounded (that one section only, not the whole file) so the prompt does
           not balloon. Test that a charter with a boundary section has that text in the inventory output.

        5. fanout-note-claims-a-hold-that-never-happens — plan_project_workflow.py, the note emitted when
           len(new_components) > MAX_NEW_COMPONENTS. It says "this is a HOLD rather than a skip" and the
           run does not hold — it continues to plan-sprint and _dispose. Reword to state what actually
           happens: research was SKIPPED, the components reach plan-sprint unresearched, and the PR is
           still going to review. Do not change the control flow; only the sentence.

        6. rule3-audit-omits-evidence-block — plan_activities.py module docstring only.
           Add `evidence_block` to the §10.1 rule-3 audit, stating it has one consumer
           (plan_revision_workflow.py:117) and whether it stays or moves. This is a docstring line; do
           not move the function in this dispatch.

        DO NOT TOUCH: docs/development/sprint.md; any vendored standard under docs/standards/documentation,
        research, testing or temporal; docs/standards/workflow-scripts.md; the sibling runners named in (1);
        the loop-back target in plan_project_workflow.py (that is an open operator ruling, item 2 of the
        runway). Do not re-open any rejection the refine pass made.
      precheck: |
        Run ALL SIX from the repo root of a worktree checked out on the PR BRANCH
        (build-draft/plan-candidates), NOT on main. Paths are relative to that branch.
        PRECEDENCE: the dispatch_context enumeration above governs WHAT to change. These checks gate
        WHETHER each item still needs doing; they never widen or narrow the enumeration.

        (1) Property: does the ref passed to worktree_add depend on --pr?
            python3 -c "import ast;t=ast.parse(open('scripts/workflows/temporal/scripts/run_plan_candidates.py').read());print([type(c.args[2]).__name__ for c in ast.walk(t) if isinstance(c,ast.Call) and getattr(c.func,'attr','')=='worktree_add'])"
            -> ['Constant']  = still hardcoded, finding reproduces, PROCEED.
            -> anything else = ALREADY DONE, STOP this item only.
            (Different check from the grep that surfaced it: this reads the AST node type, so a
             comment mentioning "HEAD" cannot satisfy it.)

        (2) Property: does a test asserting the stub-charter behaviour exist?
            python3 -m pytest scripts/workflows/temporal/tests/unit/test_plan_candidates.py -q --collect-only -k "stub or required_section or charter_content" 2>&1 | tail -2
            -> collects 0 tests = no positive content check exists, PROCEED.
            -> collects >=1 = ALREADY DONE, STOP this item only.

        (3) Property: does the parent handle a child exception at all?
            python3 -c "import ast;t=ast.parse(open('scripts/workflows/temporal/modules/assistant/plan/plan_project/plan_project_workflow.py').read());print(sum(1 for n in ast.walk(t) if isinstance(n,ast.Try)))"
            -> 0 = no handling anywhere, PROCEED.
            -> >0 = read them; if one posts to the PR before re-raising, STOP this item; if they are
               unrelated, PROCEED.

        (4) Property: does the rendered inventory carry any charter scope text?
            cd scripts/workflows/temporal && python3 -c "import sys;sys.path.insert(0,'.');from pathlib import Path;from modules.assistant.plan.plan_candidates import plan_candidates_activities as own;o=own.component_inventory(Path('../../..'));print('What this is NOT' in o)"
            -> False = boundary text absent, PROCEED.
            -> True  = ALREADY DONE, STOP this item only.

        (5) Property: is the false HOLD claim still in an emitted string?
            python3 -c "import ast;t=ast.parse(open('scripts/workflows/temporal/modules/assistant/plan/plan_project/plan_project_workflow.py').read());print([n.value for n in ast.walk(t) if isinstance(n,ast.Constant) and isinstance(n.value,str) and 'is a HOLD rather than a skip' in n.value])"
            -> non-empty = reproduces, PROCEED.  -> empty = ALREADY DONE, STOP this item only.
            (Here the defect IS a word in an operator-facing string, so testing for that string tests
             the property rather than standing in for it.)

        (6) Property: does the rule-3 audit name evidence_block?
            python3 -c "import ast;print('evidence_block' in (ast.get_docstring(ast.parse(open('scripts/workflows/temporal/modules/assistant/plan/plan_activities.py').read())) or ''))"
            -> False = audit incomplete, PROCEED.  -> True = ALREADY DONE, STOP this item only.

        A STOP on any single item does NOT stop the dispatch — the other items proceed. "Not yet
        warranted" is not a state that arises here: all six are present defects, so the only STOP
        arm is "already done".
  laundered_deferrals:
    caught: 0
    of_total: 1
  homeless_items: 2
  redispatched: false
```

---

## Post-Run Reflection

**Notes — real, no consequence I can state, recorded only so nobody re-derives them**

- **`docs/development/reviews/roadmap.md` is an unobserved write.** `reviews` is in `NOT_A_COMPONENT`, so `component_roadmaps` and `component_dirs` both skip it — while `PERMITTED_PATHS`' `^docs/development/[^/]+/roadmap\.md$` exempts it from `boundary_crossings`. Creating, editing or deleting that one path is invisible to every guard. No such file exists, nothing downstream reads it, and I cannot state what breaks — so a note, not a finding. Worth one line in the exemption's comment if that code is touched anyway.
- **The PR body's Summary block reports `27920 bytes rendered`; its Verified section reports `30670`.** I ran the dry run: `30670` is correct. The Summary block is the draft's and was not refreshed by the refine pass. Nothing downstream reads it.
- **The `--dry-run` path still computes the disproof and discards it.** The refine reflection names this as the cheapest available conversion of a habit into a mechanism, and it is right — `component_inventory` is already called there and its block is thrown away. Printing it would have surfaced the shell misclassification and both count errors before the PR existed. Not a finding; a genuinely good suggestion the run made about itself.

**Friction with this workflow's own prompt**

- **`ESCALATED` vs `fix-it-anyway` is still ambiguous, and PR #85's reflection said so one PR ago.** That reflection recorded pass 1 escalating four one-line build-family fixes and the correction pass simply fixing them — concluding the remedy vocabulary needs `fix-out-of-scope`. I hit the identical fork on `turn-cap-authority-orphaned` and resolved it the other way, because it is a **standards** surface and therefore genuinely human-gated, which the four build-family fixes were not. That distinction is the whole answer and it is nowhere in the prompt. One clause — *escalate when the surface is human-gated; fix when it is merely not yours* — would settle it. Two consecutive PRs have now spent reasoning on it.
- **The consequence gate did real work again.** Three candidate items were demoted to the notes above rather than entering the runway. Each would have cost the operator a ruling for nothing.
- **`homeless_items: 2` is doing more work than the field can express.** Both `charter-unreachable-by-any-loop-back` and `component-filename-convention-unruled` are legitimate items with no corpus surface — but they are homeless for *opposite* reasons: one because the capability does not exist yet, one because the surface exists and is human-only. The field counts them identically, and a CPI pass mining this number cannot tell a missing capability from a correctly-gated one.
- **The deleted-artifact sweep's widening to functions and contracts earned its cost here.** This PR deleted no files. Enumerating `new_sprint_sections`'s and `component_dir`'s assertions by name and mapping each to a destination is what let me say "nothing landed NOWHERE" as a fact rather than an impression — and it is what surfaced that one contract (`component_dir`'s `ValueError` on an unmappable slug) was legitimately dissolved rather than lost, which a file-level sweep would have had no way to see.

- **The `--pr` worktree defect is now four instances with one correct fix present in the same directory.** `run_triage_candidates.py:55`, `run_plan_candidates.py:58` and `run_plan_sprint.py:58` all hardcode `"HEAD"`; `run_plan_revision.py:141` does it correctly, as do three build workflows and `plan_project_workflow.py:106`. This is not a family "pattern" — it is one correct implementation and three copies of a bug, and the next runner authored will be a fifth coin-flip. The mechanism this repo would reach for is already built: an AST census over `worktree_add` call sites asserting the ref argument is not a bare constant when the module parses a `--pr` flag. Not filed — it is one line of the runway above plus a decision about the two files this PR does not own.

Pumapumapumas:
Closing in favour of building this as an **activity**, not a child.

**The measurement that decided it: 1,605 lines and a 173-line prompt for a job the operator describes as "move the info to the right place so the next step can happen."**

Copying a ruled candidate's summary into a component folder needs no judgement, so it needs no model. It is plain code in the parent — no prompt, no turn cap, no worktree, no dispatch cost. **It runs for free.**

**All eight held findings dissolve rather than get fixed**, because the thing that had them stops existing: no charter to be unread (`charter-content-unchecked`, `charter-boundary-never-read-back`), nothing unreachable by a loop-back (`charter-unreachable-by-any-loop-back`), no guard firing after the PR opens (`guard-failure-leaves-poisoned-pr`), no worktree flag to get wrong (`redispatch-worktree-cut-from-head`), no fan-out note claiming a hold that never happens.

**The one real judgement — naming the component — moves upstream.** The run that FILES a candidate names the component it belongs to, because it has just written the proposal and knows. Downstream then has nothing to decide.

**`component-filename-convention-unruled` is settled separately and is now moot for this PR:** `sprint.md` no longer carries a one-phase exception, so every component gets a `roadmap.md` plus numbered phase docs. Ruled by the operator and committed as `e3b9f3d`.

**The brief was the cause and that is recorded rather than glossed.** It asked the run to decide *"what is the minimum legitimate structure a triaged ship row needs — state your answer and its rationale"*. That is an invitation to design, handed to a task that should have been a file copy, and 1,605 lines is what the question bought.

Nothing here reflects on the run's execution: it fixed 10 defects of its own, rejected 3 findings with reasoning, and correctly refused to loop on an item only a human could rule. It built what it was asked for.

PR #90, opened 2026-08-14
plan-revision: plan the persistent memory protocol whole — all 8 phases, write-failure answered, MMF questioned

Pumapumapumas:
Revises `docs/development/persistent-memory-protocol/` against four operator objections. The planning docs are the deliverable; this body is an index.

## Planning changes

- **`roadmap.md` rewritten.** Opens in plain words; a terms table defined once; **§ *The protocol, whole*** states the design as nine commitments *before* any ordering; **§ *The order, and what each part waits on*** records gates as scheduling facts; **§ *Questioning the Memory Management Framework*** replaces the previous version's treat-MMF-as-input framing.
- **Three new phase docs** — `phase5_snapshots_then_retention.md`, `phase7_s3_aggregation.md`, `phase8_the_poller.md`. All eight phases now have one. 22 anchor links repointed from `roadmap.md#phase-N-…` to the new files.
- **`container` is gone as a name for the bag** (5 occurrences). Every phase doc opens with a plain-language paragraph and defines its terms on first use.
- **The write-failure question is answered** in `phase3_the_emit_rule.md` § *When the journal cannot be written* — four cases, and a paired write is one unit with **two** events (intent, then completion) so replay applies only an intent that has a completion. Honoured by phase1 r8/r9 (`incomplete` state, refuse-to-start), phase4 r7, phase5 r7, phase6 r6, phase7 r6.
- **The 15 concepts and the coverage map survive**, with every cell re-verified against actual requirement numbers, plus two rows for concepts this revision adds.
- **Four MMF amendments PLACED** as items 11–14 in `memory-management-framework/roadmap.md` § *Standards-amendment candidates* — not left in this body, which dies at merge. Item 13 is an objection to MMF's own pending item 10, which sits under a fired-trigger banner.
- **Nothing written to `sprint.md`.** No vendored standard, `problem-statement.md`, code, script or config touched.

## Brief premises that did not hold — checked, and one is false

1. **The brief asserts a 2026-08-13 ruling: *"`sprint.md` is the immutable record… a closed sprint is history: it is not appended to, and it is not rewritten."*** That text **is not in `docs/development/sprint.md`.** What is there: *"a sprint plan is never a history lesson"* (`:22`) and *"a checkbox here means shipped and validated"* (`:18`). I proceeded on the operational half — do not write `sprint.md` — which `config/rules/standards-governance.md` requires independently. The rest is unsourced and is not quoted anywhere in the plan.
2. **The old roadmap quoted a `sprint.md` rule deleted the same day it was written** — *"One phase needs no roadmap; do not create one to be tidy"*, replaced by commit `e3b9f3d`. Citation corrected.
3. **Four stale cross-references the brief did not name** — phase3 cited `bagit.txt` where phase1 had already corrected it to `bag-info.txt`; two Open-inputs item numbers; one requirement number.
4. **The premise the brief predicted but did not name.** phase3's old requirement 3 promised the journal envelope would *"extend MMF Phase 3's typed exit record."* `docs/standards/exit-protocol.md` §2 forbids fields with no existing consumer (the journal event needs six) and §2.5 bounds the envelope at **4096 bytes** (a cycle's authored output measures **39,772**). Ruled impossible; requirement 3 is now *one vocabulary, two contracts*. That document is `DRAFT — NOT BINDING`, so the argument rests on the measurement and the contract shapes, with the citations as corroboration.

## Reviews

**Architect** — Critical: write-ahead ordering covered one of three ways a paired write breaks (store-write failure, retry duplication, no home for the store-assigned address) → fixed by the intent/completion split. Warnings addressed: case (d) needed a durable landing place; `incomplete` was framed both as a fourth state and as an orthogonal flag; `edge_id` "bound at ingest" has no ingest tier; phase5's gated scope; requirement 3's rationale demoted off a draft standard; property tally; challenges 2 and 3 re-costed and re-framed. **1 rejected with reasoning** (the roadmap's order table already carried phase5's split — verified by quality-control); 0 deferred.

**Planner** — Critical: phase7 r3's *"a constraint on Phase 6, stated here so it lands there"* did not land there → phase6 r7 added, plus a forward-constraint coverage table that caught a **second** unlanded constraint (egress classification slot → phase1 r7). All warnings fixed: requirement renumbering, coverage-map cell, six mis-pointed Temporal gate citations (`temporal-integration.md` contains neither string; both are `sprint.md` checkboxes), transcript retention period, phase7 seams, phase5 gated labels, phase8 escalation-as-emit. 0 deferred.

**Security** — 2 Critical: phase8 bounded neither which child a cue starts nor what row content becomes input, with provenance stripped by the rebuild → r6, r7, ingress in its gate list, and phase4 r9 carries provenance through; phase4's containment covered only the scratch root while the roadmap sells a live restore → contract binds every replay target, restore built as r8 with a path allowlist. All High/Medium fixed: transcript-loss escalated to a stop, gap events closed to a typed field set, capture filter moved to append time, redaction separated from `pruned`, retention floors and event exemptions, local threat model stated (append-only is a convention the fleet keeps, not one the filesystem enforces). 0 deferred.

**Standards** — 3 Critical: C-074's Note credited two arguments the roadmap retracts → corrected (Note only; `decision` untouched, `status` untouched); the property count contradicted itself → five-row table, fails property 4 only; the MMF proposals belonged in MMF's amendment list → placed as items 11–14. Warnings fixed: `problem-statement.md` misquote (the phrase is the synthesis's), Temporal §7.5 citation, content-store reclamation placed rather than noted. **Ruled for me:** per-phase-doc term repetition is *not* a Documentation Standard violation (a glossary is not a codified field set; no drift found across eight docs), and § *Questioning the MMF* is in the right bucket.

**Quality-control** — consolidated the four Criticals into **two design decisions the plan had not made** (does an emit record an intent or a fact; is a rebuilt store trusted like a hand-written one), **one discipline the plan invented and did not apply to itself** (forward constraints), and **one argument stated in three disagreeing places** (component-vs-phase). All four clusters resolved. Its own addition — the roadmap sells *"if a table gets corrupted, you regenerate it"* and no requirement delivered it — is fixed as phase4 r8.

## Cross-reference consistency: PASS

All `.md` links and every `#anchor` in the changed docs resolve (scripted check, 0 broken). The 4 unresolvable `exit-protocol.md` paths in MMF's roadmap are inside a fenced code block on `main`, correct relative to where that block is destined to be pasted — pre-existing, not touched.

`./testing/run-all.sh` from the worktree: **RESULT: PASSED** (2956 passed, 3 skipped).

## Sprint-item candidate — operator only

This component still has **no row in `sprint.md`** (`grep -i "persistent.memory" docs/development/sprint.md` → zero matches). It is planned but unscheduled against everything else. Phases 1–4 are ungated and deliver the whole thesis; phase 6 needs *Port `review-runs`*; phases 5 and 8 need the Temporal server; phase 7 needs a second machine plus two operator rulings. Surfaced here for operator-reviewed insertion, per `standards-governance.md`.


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Ruled that the journal event is a **separate contract** from MMF's typed exit record rather than an extension of it, reversing the previous plan's requirement 3. Alternatives: keep the extension and propose raising `exit-protocol.md`'s 4096-byte bound; or split but call it an extension anyway. Why: the size gap is ~10× on a measured figure (39,772 bytes of authored output per cycle against a 4 KB fixed part) and the field gap is six fields no parent branches on, so the extension was not achievable rather than merely awkward. **Confidence is low on the framing, not the conclusion** — `exit-protocol.md` is `DRAFT — NOT BINDING`, so I rebuilt the argument on three intrinsic differences (reader, lifetime, admission rule) and demoted the citations to corroboration.

- **[Low]** Added a **restore** requirement to phase4 (r8) rather than deleting the roadmap's *"if a table gets corrupted, you regenerate it"* claim. Alternatives: state that replay-to-scratch is all this component builds and soften the roadmap sentence. Why: the operator's whole objection was that buildability drove the design, and "delete the capability from the sales pitch because no requirement delivers it" is that failure in miniature. The containment cost is one allowlist, and the alternative leaves the capability to whoever needs it first — writing into a tree where a `destination` of `config/hooks/` is executable.

- **[Low]** **Rejected** the architect's warning that the roadmap's order table commits the symmetric-constraint error a third time by listing phase5 flatly under the Temporal gate. Why: it does not — the third column already carried the split, which quality-control verified independently. The substance underneath (nothing labelled *which* requirements are gated) was real and is fixed. Recorded because rejecting a review finding on a specialist's home turf deserves the reasoning in writing.

- **[Medium]** Placed the four MMF amendments in **MMF's own § *Standards-amendment candidates*** rather than only this PR body, as the brief's wording suggested. Alternatives: PR body only, per the brief. Why: a PR body dies at merge — `memory-model.md` §2.6 calls it *"the most effective burial available"* — and challenge 3 is an objection to MMF's pending item 10, which sits four entries above under a **⚑ THE TRIGGER HAS FIRED** banner awaiting an operator paste. An objection recorded where the person pasting will not look is not an objection. That section is a planning artifact a dispatch may edit; the standards files themselves are untouched.

- **[Medium]** Reversed challenge 3 during review. The first draft argued MMF's Kind 3 amendment *"will be stale on arrival"* and should be held. Read in full, MMF's own re-open condition says property 1's failure is a binding fact while *"the other two legs are properties of the content and do not move"* — so the ruling is true when it lands and only one leg is affected. Reframed to **land it and add the cross-link**. Recorded because the first version would have parked a measured finding behind an unstarted phase — the exact failure this plan argues against everywhere else.

- **[Medium]** Wrote full phase docs for three gated phases, **diverging from `sprint.md`:34** (*"Phase docs are written when a sprint is picked up, not in advance"*). The operator's instruction overrides, but the divergence is stated in the roadmap rather than passed over silently. It is also now **self-evidencing**: writing the gated docs surfaced three forward constraints on earlier phases, and **two had landed nowhere** — one of them against phase 6, which is fourth in rollout order. The sizing risk `sprint.md` names is real and unmitigated; those docs will need re-reading when their gates open.

- **[Medium]** Edited **C-074's Note** in `candidates.md` rather than only flagging it. Alternatives: leave the row and note the discrepancy here. Why: the Note credited the roadmap with two arguments the revised roadmap explicitly retracts, and C-074 is the row the operator rules component-vs-phase from. `standards-governance.md` makes `research/` a working surface and reserves only `status` to the human; `decision` stays blank and untouched.

- **[Medium]** Escalated a failed **transcript** write from an ordinary gap to a run-stopping failure, unlike the other two unpairable writes. Why: it is the fleet's only record of what commands ran, the fleet runs with permissions bypassed, and a run can itself create the disk-full condition that drops it — so "continue with a gap" is evidence loss that reads as a routine defect. Cost: a run dies on a full disk that would previously have completed.

## Deferred Work

Nothing deferred. Everything quality-control ranked deferrable was cheaper to fix than to record, per `engineering-quality.md` § *Resolve live* — including the content-store reclamation gap, which the previous version stated twice as *"until something owns this"* and which is now written into the decision that creates it (`phase2` r7(a)) with a conditional checkbox in `phase5`.

The **sprint entry** is surfaced in the PR body rather than deferred: it is operator-only by `standards-governance.md`, so there is no tracker I could point at without writing the file this run may not write. Stating it with no pointer is the honest shape.

## Post-Run Reflection

**Friction**

- **A brief premise was false and it was the one describing a ruling.** The brief quoted a 2026-08-13 sprint-file ruling verbatim; `docs/development/sprint.md` contains no such text. The operational half was independently required by a rule, so nothing was blocked — but had the plan needed the immutability half, it would have been built on a sentence with no source. The workflow's *verify the task's own asserted facts* instruction caught it; it is worth noting that this is now **four** false brief premises in two days, all rulings or counts.

- **`exit-protocol.md`'s `DRAFT — NOT BINDING` banner is easy to miss and expensive to miss.** I built the single largest design change of this run on §2 and §2.5 before two reviewers pointed at the banner. The conclusion survived on intrinsic grounds, but a run that had leaned on the authority alone would have shipped a requirement anyone can reopen by reading line 3.

- **The forward-constraint problem generalises past this component.** Writing gated phase docs surfaced three constraints on earlier phases and two had landed nowhere — invisible while the gated phases were roadmap rows. The roadmap now tracks them as a table, and that device probably belongs in any component whose later phases constrain its earlier ones. It is the same shape as the concept-coverage map, applied to the plan's own internals.

- **`candidates.md` C-074's Note was a claim about another document's state that nothing re-derived when that document changed** — which is the mechanism MMF's roadmap already names in its own words and which `test_candidates_prose_matches_the_table.py` gates for counts but not for prose claims about other files. Recording it because this is the second instance in that file.

- **The Stage 4a dispatch instruction to name rejected brief premises to reviewers paid for itself, measurably.** All four reviewers independently re-verified the false sprint-ruling premise, and the standards reviewer found the *related* stale citation I had fixed but not flagged (the deleted one-phase exception). Worth keeping.

- **Stage 4b's explicit CONSOLIDATE instruction is the highest-value line in the review stage.** The narrow lenses returned 4 Criticals and ~30 warnings, which reads as a plan in trouble; consolidated they were two unmade design decisions, one discipline unapplied to itself, and one argument stated three disagreeing ways. Remediating the 34 separately would have produced fixes that did not cohere — the intent/completion split alone dissolved six findings across three reviewers. *"Look for meta-patterns"* would not have produced that.

- **A gap in the workflow's own guidance:** it tells the run to verify the task's asserted facts, but not to check whether a cited standard is marked non-binding. That is a distinct failure mode from a wrong line number and it bit this run.

PR #92, opened 2026-08-14
build-refine: plan-feature — the write half of planning, with guards that cover what their rows claim

Pumapumapumas:
## Summary

Builds **`plan-feature`** — the WRITE half of planning. It takes ONE component directory and produces that component's `roadmap.md` plus its numbered `phaseN_<name>.md` docs, from that component's research. It does not review them, does not size them, and does not touch `sprint.md`.

Mirrors `triage_candidates`' standalone-and-composable shape exactly: entry script (`scripts/plan_feature.sh`), runner (`scripts/run_plan_feature.py`), and a workflow function `plan_project` calls directly.

**`plan-verify` is not built here and is referenced only as the handoff it is.** `plan_revision` is untouched apart from a one-line defect fix described below.

### Where it sits

```
triage-candidates → plan-candidates → research-write → research-verify → plan-feature → plan-sprint → review-pr
```

`plan-feature` runs **inside** the per-component research loop, immediately after the `research-verify` that gated its evidence — not in a second loop afterwards, which would separate each component's plan from its own evidence by every other component's research cycle.

### The four prohibitions are OBSERVED, not requested

| Prohibition | What observes it |
|---|---|
| **No hours** — sizing is `plan-verify`'s | `hour_estimates` over every top-level doc, keyed on estimate SHAPE |
| **No rename/renumber/delete of an existing phase doc** — the number is IDENTITY | `phase_docs` snapshot + `ids_deleted` — a rename and a renumber are both a filename that vanished |
| **New plan docs are `roadmap.md` or `phaseN_<name>.md`, on an unused number** | `malformed_phase_docs` over `plan_docs`, `reused_phase_numbers` |
| **ONE component; `sprint.md`, siblings and its own `research/` are off-limits** | `FORBIDDEN_PATHS`/`permitted_paths` via `worktree_state`/`boundary_crossings` |

Plus the shared column guards on `candidates.md` (`decision`/`status`/`component`), checkbox counting in both directions, and `grants_that_vanished` on the write grants. The `You MAY NOT` table and the before/after snapshots are registered in `MAY_NOT_OBSERVERS` / `DISAPPEARANCE_OBSERVERS`, which the two existing registry tests **discover by AST** — so this workflow inherited every assertion in them automatically.

---

## What the refine pass changed, and why

**Four defects, and three of them are the same defect: a mechanism that did not cover what its own registry row claimed.** Every one was reproduced by execution before it was fixed, and every fix was mutation-tested afterwards.

### 1. The hour guard fired on ordinary prose (Critical, code-reviewer)

`_HOURS` gained a `\.?` after the label group to catch `Est. 2.5 hours`. That period also matched a genuine **sentence-ending** full stop, so `[^.\n]` stopped bounding the match to one sentence:

| Input | Shipped pattern | Now |
|---|---|---|
| `the estimate. It took 3 hours` | **MATCHED** | no match |
| `That is the whole effort. Then 5 hours later` | **MATCHED** | no match |
| `We settled the sizing. We waited 2 hours` | **MATCHED** | no match |
| `Est. 2.5 hours` | matched | matched |

**Consequence:** `run_plan_feature` raises and fails a plan that violated nothing — the exact "fail three correct runs" failure the module's docstring says the shape-keyed design avoids. **Aggravating: the comment above `_HOURS` cited the first of those strings as the false positive it prevents**, so the file readers trust most asserted the opposite of the truth. The period is now permitted after the abbreviation `est` and nowhere else; the comment says what is actually true, including the residual limit.

The three prose fixtures were all lifted verbatim from the tree, and none of them happens to contain a marker word — which is why a fixture set that looked exhaustive was blind to the whole class. All three failing strings are now negative controls.

### 2. The component name reached a regex unescaped (Critical, code-reviewer)

`permitted_paths` interpolated `component_rel.as_posix()` raw. The sibling `plan_sprint_workflow.py:87` does `re.escape(sprint_rel)` for the identical construction.

**Consequence:** a component named `v2.1-migration` produces a grant whose `.` matches any character, so it also grants `v2x1-migration/` — the write boundary silently widening to a SIBLING, which is the one failure the whole module exists to prevent. Inside `plan-project` the name has been through `component_slug` and cannot carry a metacharacter; **the standalone dispatch, which is a documented supported mode, has no such filter.** All three boundary tests used `alpha`, which has nothing to escape.

### 3. The guards were scoped narrower than the grant they enforce (Warning → the widest one)

The write grant is `<component>/[^/]+\.md$`. `malformed_phase_docs`, `hour_estimates` and the checkbox counter were all sourced from a `^phase` sweep instead.

**Consequence, and the reason this is the most serious of the four:** a phase doc written as **`the_run_bag.md`** — the number dropped entirely, which is the Documentation Standard's first-named failure mode and what a model writes when it forgets the convention rather than mistyping it — does not begin with `phase`, so it was invisible to `malformed_phase_docs` while the `You MAY NOT` row said that function observes it. It is permitted by the grant, so `boundary_crossings` exempts it. **It passed every guard.** The same gap left an hour estimate or a pre-ticked checkbox in any other top-level file read by nothing.

Closed by a new reader, `own.plan_docs` — every top-level markdown file, i.e. **the grant's own set** — so guard scope and grant scope are now equal by construction rather than by remembering. `roadmap.md` is the one exempt name and the code says so explicitly.

### 4. The phase-identity message was unreachable (Warning, code-reviewer)

Every phase doc is also a granted path, so a renumber satisfies both `grants_that_vanished` and the identity check. With the generic guard first, the operator got *"a file you may write ceased to exist"* for the one failure this workflow was built to name — the specific message, and the only place the identity rule is explained, was dead in practice.

The identity check now runs ahead of it. Nothing is weakened: that guard parses no file, so the reason `grants_that_vanished` runs early — every READER below assumes the file it parses is still there — still holds.

### 5. The escape test passed for the wrong reason (my own Stage 1 finding)

`test_the_runner_REFUSES_a_component_outside_the_repo` built a plain `tmp_path` directory, so `preflight` refused it with *not inside a git repository* and returned 1 **before the escape check ever ran**. Deleting the `is_relative_to` guard outright would not have failed it. Confirmed by execution (stderr read `not inside a git repository`). The tmp repo is now `git init`-ed and the message is asserted, because an exit code alone cannot tell two refusals apart.

### 6. Three structural items (refactoring-evaluator, Medium)

- The two registry preambles **restated** the argument `triage_candidates_workflow` makes, where `plan_sprint_workflow` **cites** it. Now cited, with only what is new to this workflow kept.
- `_plan_boxes` hand-rewrote the phase-file predicate in a second file. Moved to `plan_feature_activities.plan_boxes`, sourced from `plan_docs` — one spelling, and it picked up the scope fix.
- The runner's `--dry-run` branch hand-assembled its own copy of the nine-key `values` dict. This family has already shipped that bug once (`plan_sprint`'s `correction_note` docstring records a dry run previewing a *different* prompt from the one dispatched). Both paths now call `wf.prompt_values`.

### Negative controls — six mutations, each predicted before running

| Mutation | Predicted | Observed |
|---|---|---|
| drop `re.escape` from the grant | 1 | **1** |
| `malformed_phase_docs` back onto `phase_docs` maps | 1 | **1** |
| swap the identity check back behind `grants_that_vanished` | 1 | **1** |
| restore the blanket `\.?` in `_HOURS` | 3 | **3** |
| narrow the hour scan back to roadmap-or-phase | 1 | **1** |
| drop `git init` from the escape test | 1 | **1** |

Tree restored from copies before each (never `git checkout --`), control green after each.

**One miss during authoring, recorded because it is the same class this pass was fixing.** The first version of the end-to-end harness stubbed `worktree_state` to `{}`, which makes `grants_that_vanished` empty on every input — the *exact* guard the ordering test claims to be ordered against. It passed while asserting nothing. The stub now mimics git: a path once reported stays reported, and a file that is gone reads as `ABSENT`. **A second miss was diagnostic, not real:** a control run went red after a mutation cycle against files byte-identical to their backups — stale `__pycache__`, not a defect. Every mutation above was re-run with the cache cleared.

---

---

## What the SECOND refine pass changed, and why

**Three confirmed defects, and all three are ONE class: a guard whose scope is not the question it answers.** That is the same class the first refine pass closed four instances of, which is why every fix below ships a **class-keyed check** rather than a per-instance one. Each defect was reproduced by execution before it was fixed, and each check was mutation-tested with the failure count predicted first.

### 1. A stray `phase_notes.txt` counted as a phase doc (Critical, my Stage 1)

`phase_docs` matched `^phase` case-insensitively with **no markdown-suffix test**, while its sibling `plan_docs` requires `.md`. Two consequences, both reproduced:

- **The deliverable guard was bypassed.** `if not after_phase: raise` is what stops a run shipping a roadmap with no phases. A component holding one `phase_notes.txt` satisfies it, so a run that wrote **zero** phase docs passes clean.
- **`planning_state` lied to the model** — its output is handed over labelled *"Counted in code, authoritative — do not recount"*, and it reported the `.txt` as a phase doc.

Closed by requiring a markdown suffix. The stem stays deliberately wide (`Phase3.md` is still a phase doc whose deletion is an offence); only the suffix tightened, and the asymmetry with `plan_docs` — which must equal the case-sensitive write grant — is now stated in the code.

### 2. Two phase docs written in ONE run could both be phase 5 (Critical, code-reviewer)

`reused_phase_numbers` built `taken` from `before` and **never grew it while walking `after`**, so a collision between two brand-new files was invisible. `reused_phase_numbers({}, {'phase5_a.md':…, 'phase5_b.md':…})` returned `[]`. Nothing else in the chain catches it: both names are grammatically valid, nothing was deleted, nothing left the grant. **A component could ship two documents both called Phase 5, past the one guard that makes a number an identity.**

The fix compares new files against each other by **full identity** (`phase_identity` → `(number, sub-letter)`), not by bare number, because rule 6's carve-out must survive. All three behaviours are now pinned:

| Case | Rule | Verdict |
|---|---|---|
| `phase5_a.md` + `phase5_b.md` in one run | two phases, one name | **caught** |
| `phase5a_poc.md` + `phase5b_rollout.md` in one run | rule 6 carve-out, planned together | legal |
| `phase5b_late.md` added after `phase5a_poc.md` shipped | rule 6 failure mode (c), retroactive sub-letter | **caught** |
| `phase5_a.md` + `phase5a_b.md` in one run | a phase cannot be both chunked and not | **caught** |

### 3. The hour guard failed runs for estimates they did not write (Warning → the one that was LIVE)

`hour_estimates` was **the only prohibition guard keyed on post-run STATE** rather than on a delta. Pre-existing estimate-shaped text failed the run with a message asserting *"plan-feature wrote N hour estimate(s)"* — false, and the component stays unplannable until a human edits text this workflow has no business touching.

**Live today, not theoretical:** `docs/development/reviews/` carries `~7h` and `~12.8 hours`. And `plan-revision` — the unsplit planner this is the write half of a replacement for — **does** size work, so any component it planned is one `plan-feature` could never extend.

Closed with `hour_hits`, a before/after `Counter` keyed on `(file, matched text)` — the `plan_boxes` idiom, and keyed on text rather than line so inserting a phase above an untouched estimate is not reported as a new one. The `before_hours` snapshot is registered in `DISAPPEARANCE_OBSERVERS`, and the registry test proved it discriminates: dropping the row fails the suite.

### 4. The dry-run/live duplication the last pass SURFACED and did not close

Pass 1 fixed `run_plan_feature.py` and surfaced its two siblings with no tracker. **A surfaced item that reaches only the next reviewer costs another dispatch cycle**, so it is closed here instead: `plan_sprint_workflow` and `triage_candidates_workflow` grew `prompt_values`, both runners call it, and the orphaned `own` import in `run_triage_candidates.py` is gone. All three dry-run paths were exercised and render 0 placeholders remaining.

The class check is `test_dry_run_previews_the_dispatched_prompt.py` — an AST sweep over **every** `scripts/run_*.py`: the second argument to `render()` may be neither a dict literal nor a runner-local call one indirection down. A new runner copying the pattern fails on the day it is written.

### 5. A `plan-feature` failure orphaned its component silently (Warning, quality-control)

`scaffold_candidate_components` decides *does this need work?* from the `<!-- plan-candidates: seeded, no research yet -->` marker, and `research-write` strips it as its first act. So a component whose research **succeeded** and whose planning then failed reads as `extends` on every later `--pr` redispatch — it never re-enters `to_research`, and nothing anywhere checks for a missing `roadmap.md`. Verified by reading the three files that form the chain; **no `try`/`except` exists anywhere in `run_plan_project`, on this branch or on `origin/main`.**

The parent now catches a mid-loop failure and re-raises it (nothing swallowed, exit code unchanged) carrying two things the operator cannot reconstruct: that automatic recovery will not happen, and the exact standalone recovery command. It also attaches the accumulated `notes` — which were local and returned only on the success path, so the accounting this function's own docstring calls *"the ones worth reading"* was guaranteed absent at exactly the moment it was being read for a diagnosis.

The remaining **capability** gap — resume keyed on the planning deliverable — is placed as **C-087**, not fixed here.

### 6. Two Info fixes

- `run_plan_feature.py`'s dry run built `rel_candidates` from the raw CLI string while the live path relativised it, so an absolute `--candidates` previewed one prompt and dispatched another. Both operator paths are now resolved and escape-checked once, identically.
- The `grants_that_vanished` message explained every failure in terms of `candidates.md` even when the vanished file was the component's `roadmap.md`.

---

## Negative controls — six mutations, each predicted before running

| Mutation | Predicted | Observed |
|---|---|---|
| hour guard back to post-run STATE | 1 | **1** |
| `phase_docs` back to bare `^phase` | 5 | **5** |
| collision check back to before-only | 2 | **2** |
| drop the `before_hours` registry row | 1 | **1** |
| hand-built values dict in `run_triage_candidates` | 2 | **1** |
| values routed through a runner-LOCAL helper | 1 | **1** |

**The one miss was my model of the fixture, not a weaker guard**, and saying which is the point of predicting first. The two runner checks are **disjoint by design** — one rejects a dict literal, the other rejects a local call — so a literal fires only the first. The sixth mutation exists to prove the second is not vacuous, and it failed on exactly that test.

Restored from copies before each (never `git checkout --`); `__pycache__` cleared between mutation and control, per the last pass's own reflection; all three files verified byte-identical to their backups afterwards.

---

## Findings I did NOT act on, with the reasoning checked

- **`refactoring-evaluator`, Medium — hoist the duplicated authorization regex literals** (`^docs/standards/architecture/research/candidates\.md$` and three others appear verbatim in three workflows). **REJECTED, and the reasoning was verified by execution rather than asserted:** the stated consequence is a *silent* authorization gap, and it is not silent. A typo'd copy of the candidates grant makes `boundary_crossings` report the run's own legitimate append as a crossing — the run fails loudly on its first act. Measured: correct grant → `[]`, typo'd grant → `['docs/standards/architecture/research/candidates.md']`. Extracting them would widen this diff into a workflow it does not own to buy a guard that already exists. The Low item (`DEFAULT_CANDIDATES` as a fourth copy) is rejected on the same evidence.
- **`standards-auditor`, Info — a `MAY_NOT_OBSERVERS` row proves a mechanism resolves, not that it fires.** Already filed by this PR as **C-086**; nothing further to do here.

## Brief premises checked, and one thing the gates caught

- *"a test enforces `docs/file_structure.txt` set-equality"* — TRUE, and the new test file is listed.
- *"every relative markdown link must resolve"* — TRUE, green.
- *"`./testing/run-all.sh` green from the worktree"* — **3,141 passed, 3 skipped, 203.20s**, taken after merging `origin/main` twice.
- **The merge landed a triage ruling, and the tally gate caught me.** `origin/main` ruled a previously-blank candidate while this branch was open, so appending C-087 left the untriaged count at thirty-seven rather than thirty-eight. `test_candidates_prose_matches_the_table` failed with *"check whether a row went missing rather than assuming it did not"* — which is exactly what had happened. Corrected against a re-measurement, not against my arithmetic.
- **`origin/main` moved three times during this pass**, the last landing a **new prompt-budget gate**. Merging it here rather than after the fact is what caught `plan_feature.md` (17,821 bytes) having no budget line — a failure that would otherwise have been discovered on `main`. The budget is set at the file's current size, per that test's own rule.

## Surfaced — real, out of scope, no tracker

Nothing new. **The item pass 1 surfaced here — the dry-run duplication in the two sibling runners — is FIXED in this pass**, so it is no longer surfaced.

## Test results

**`./testing/run-all.sh` from the worktree: PASSED — 3091 passed, 3 skipped, 204.01s**, run after merging `origin/main` (which had moved on by one commit, #90). integration/e2e skip (no such directories in the tree).

`tests/unit/test_plan_feature.py`: **44 tests** (was 33). New: the unnumbered-name case, the roadmap exemption, the metacharacter grant, the grant-vs-guard scope equality, `plan_boxes` over the whole plan, three sentence-crossing prose fixtures, and **two tests that drive `run_plan_feature` end to end** — the layer where the ordering defect lived and which nothing previously exercised.

---

## Deviations from plan (from the draft pass, unchanged)

1. **`plan_sprint_activities.py` is deleted.** Both of its helpers earned a second consumer — `candidate_decisions` because `plan-feature` holds the same `candidates.md` write grant, and `checked_boxes` because `plan-feature` writes phase docs. §10.1 rule 3 is mechanical — *"consumer count decides, never taste"* — so both move to `plan_activities` and the module is left empty and removed. Notably, that module's docstring argued `checked_boxes` was *"genuinely single-consumer, not merely single-consumer-today"* because triage's path boundary forbids every checkbox-bearing file. True of the workflow it reasoned about, and one workflow too general. **A claim that no second consumer CAN exist is a claim about every workflow not yet written.**

2. **`evidence_block`'s parameter renamed `repo_root` → `tree`, and both call sites re-anchored.** In `plan-project` the component is scaffolded and researched **in the worktree**, so a repo-anchored enumeration lists every pool except the one the run is planning from — and reports having seen them all. `plan_revision`'s identical call site is fixed in the same one-line way; it is the same defect, not a refactor of that workflow.

3. **`plan-roadmap` and `plan-phase` are removed from the guide's target chain.** They were two prospective workflows for the two document layers `plan-feature` now writes together. Splitting them would put a dispatch boundary in the middle of one decision — the phase boundaries *are* both documents' subject — so this is a resolution, stated as one.

## A pre-existing defect found and fixed: the link checker has never run in a dispatch

`test_relative_links_resolve.py` matched `SKIP_PARTS` against `set(p.parts)` of an **absolute** path. Every autonomous dispatch works in `<repo>/.claude/worktrees/<name>/`, which puts both `.claude` and `worktrees` in the prefix of every single file — so the intersection matched everything and `_files()` returned nothing.

**Re-verified independently this pass, same commit on both sides:** the old filter yields **0** files from the worktree, the new one **354** (the checker's own `_files()`, after its ownership filters, now returns 241 → 244). The check passed from the main checkout and could not run from a worktree, which is where every autonomous run works.

## Brief premises checked — one was false

| Premise | Verdict |
|---|---|
| *"eleven of thirteen components under `docs/development/` have no roadmap and no phase docs"* | **FALSE — it is 14 of 16**, re-measured post-merge: `find docs/development -mindepth 1 -maxdepth 1 -type d ! -name reviews \| wc -l` → 16; `find docs/development -maxdepth 2 -name roadmap.md \| wc -l` → 2. The "11 of 13" figure appears verbatim in `docs/standards/documentation/README.md:50` as an *upstream* anecdote. Direction unchanged. |
| *"`plan_project`'s docstring records the gap"* | TRUE |
| *"`triage_candidates` is the standalone+composable precedent"* | TRUE |
| *"`plan_revision` stages 1–3 are ASSESS → PLAN → REVISE"* | TRUE |
| *"correct the `workflows.md` chain if it still disagrees"* | It did; corrected |
| *"a test enforces `docs/file_structure.txt` set-equality"* | TRUE — `test_file_structure_map_covers_the_tree.py`, 6 tests, green post-merge |

## Standards audit

`standards-auditor` returned a **clean pass** — no Critical, no Warning — on folder-per-workflow layout, rule-3 promotion mechanics and the deletion of the emptied module, vendored-standard immutability, `config.yaml` conventions, the `candidates.md` row and its routing, relative-link depth, test placement, and parent/child composition.

Worth stating plainly, since it sat beside two Criticals: **the standards measure structure, and the defects were semantic.** Nothing in the vendored standards requires a boundary regex to escape an interpolated segment or a filename classifier to be exercised against a non-matching case. That is `code-reviewer`'s lens by design, and the split worked.

## Surfaced — real, out of scope, no tracker

- **`run_plan_sprint.py` and `run_triage_candidates.py` carry the same dry-run/live `values` duplication that was fixed here.** Neither is in this diff and neither is currently wrong; both are one added placeholder away from previewing a prompt that is not the one dispatched, which is a bug this family has already shipped once. Fixing them means editing two workflows this PR does not own.

## Not in scope, and not built

`plan-verify`; hour estimation; `--project`/`--feature` flags on `plan-project`; retiring or refactoring `plan-revision`; renaming or renumbering any existing phase doc; writing `sprint.md`.


Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Removed `plan-roadmap` and `plan-phase` from `workflows.md`'s target chain rather than leaving them as future workflows. Alternatives: leave both drawn and have `plan-feature` write only one layer; leave them drawn as duplicates of work now done. Why: the phase boundaries ARE the roadmap's entries and the phase docs' subjects, so splitting them puts a dispatch boundary inside one decision and gives two contexts a chance to disagree about the same thing. **This is the decision most worth a second opinion** — it forecloses a shape the guide had been carrying for a while, on an argument rather than on measurement, and reversing it later means splitting a working workflow.

- **[Low]** `max_turns: plan-feature: 250`. Alternatives: 200 (matching `triage-candidates`), 300 (matching `plan-revision`). Why: the stage walk lands near 70 and the variance is in the phase COUNT rather than in any stage, so a component with eight phases is a materially different run from one with three. 250 is ~3x the walk, matching `plan-sprint`'s posture for comparable read-then-write planning. It is a ceiling, not a spend — but nothing has measured it and one run will not settle it.

- **[Low]** Gave `plan-feature` a `candidates.md` write grant. Alternatives: bar it from the file and have it surface proposals in the PR body under the "your write boundary wins" clause. Why: that clause exists for runs whose *dispatch* forbids the file (the three research children); nothing forbids this one, and a proposal left in a PR body dies at merge — measured, three times. The cost is that this workflow now needs the full set of column guards for a file it writes one row of.

- **[Medium]** Deleted `plan_sprint_activities.py` instead of leaving a one-consumer copy of either helper behind. Alternatives: duplicate `_CHECKED` into `plan_feature_activities`; leave `candidate_decisions` where it was and have `plan-feature` import across workflow folders. Why: rule 3 is mechanical and both helpers genuinely gained a second consumer; a duplicated parse is the exact drift `normalise_cell` exists because of. The module being left empty is a consequence, not a goal.

- **[Medium]** Fixed `plan_revision`'s `evidence_block(repo_root)` call despite the brief saying not to refactor `plan-revision`. Alternatives: leave it and surface it. Why: it is the same one-argument defect I had just fixed one caller over, not a refactor — and leaving a known-wrong call site beside a corrected one invites the next reader to conclude the corrected one is the anomaly. Flagged here because it is a scope judgement a reviewer could reasonably reverse.

- **[Medium]** Made the hour detector key on estimate SHAPE rather than on the word `hours`. Alternatives: `JUDGEMENT` in the registry with no mechanism, on the grounds that prose has no artifact; a word-keyed pattern. Why: a word-keyed pattern fails three correct runs against this tree today (measured), and `JUDGEMENT` gives up a property that IS mechanically checkable. The residual cost is stated in the code: an estimate written in a spelling none of the three alternatives covers passes.

- **[High]** Ran `plan-feature` inside the per-component research loop rather than in a second loop after it. Alternatives: a second `for` over the same `origin` map. Why: the child plans from the evidence `research-verify` just gated, and a second loop puts every other component's full research cycle between the two.

- **[High]** Scoped the write grant as `<component>/[^/]+\.md$` — a shape — rather than adding a deny rule for `research/`. Why: there is no rule to forget to update when a component grows a second subdirectory. Asserted against a directory that does not exist today, which is the whole claim.

## Deferred Work

- **`plan-verify` — the fresh-context judge for a plan, and the owner of hour estimation** — Why deferred: explicitly out of scope for this dispatch; it is the sibling half of the split this PR lands the write side of. Tracked at: `docs/standards/architecture/research/candidates.md`, row **`C-085`**, `decision` blank, `status: open`, `component: workflow-decomposition`. Verified by: `git show HEAD:docs/standards/architecture/research/candidates.md | grep -c '^| C-085'` → `1`, **on this branch; lands at merge.** The row body records the consequence rather than the absence — with this PR merged, no pass anywhere in the pipeline sizes a plan (`plan-feature` fails the run on an hour figure, and `plan-sprint` totals hours it is handed and derives none), and `review-pr --type planning` judges a diff rather than reading a roadmap cold. It is a PROPOSAL — capability that would be added — so `candidates.md` is its home and an issue is not.

Nothing else is deferred. The three other findings this run produced were all resolved live: the link-checker blindness (fixed, with the fix demonstrating itself by going red on a real break), `evidence_block`'s anchor at both call sites (fixed), and the `before_boxes` registry entry that asserted coverage without naming code (fixed).

## Post-Run Reflection

**Friction**

- **The brief's "eleven of thirteen" was false (it is 14 of 16), and the string is a near-verbatim quote of `docs/standards/documentation/README.md:50` — where it is an *upstream* anecdote about conformance reports flagging 11 of 13 folders, not a claim about this tree.** That is a specific and repeatable failure mode: a memorable figure from a nearby document migrating into a brief as though it had been measured here. One `find` caught it. Worth noting that the direction was unchanged, which is exactly why it would have survived — a false premise that supports the right conclusion gets no scrutiny.

- **The brief said `./testing/run-all.sh` green from the worktree, and that was not achievable at the branch point** — `test_relative_links_resolve.py` fails from any worktree at `4cf4c55` and passes from the main checkout. Diagnosing whether that was mine cost a stash cycle and a cross-checkout comparison. A brief asserting a green baseline is asserting a measurement; this one had not been taken from a worktree, and no dispatch's ever had.

- **Three hand-maintained control sets and one census had to be widened by hand** (`test_authorization_is_observed`, `test_disappearance_is_observed` ×2, `test_pr_url_address`, the repo-root census). Each is deliberate and each says so in its own message — *"a THIRD workflow arriving is expected to fail here once, on purpose"* — and that design is right. Worth recording that the cost is real and rises linearly with fleet size: five edits to add one workflow, and each is a place a future author can widen the set without actually checking coverage.

- **The link-checker defect is the interesting one, and it generalises.** A test whose filters run against absolute paths is invisibly scoped to where it is invoked from. `test_relative_links_resolve.py` had a vacuity guard and it is the only reason this surfaced at all — the assertion it protects had never once run in a dispatch. **Worth a sweep for other absolute-path filters in the testing tree**, and worth noting as evidence that the vacuity-guard convention pays for itself: without it, the suite would have reported green over a scan of nothing indefinitely.

- **The Documentation Standard's prose and its own binding regex disagree about version suffixes** (`phase3_the_thing_v2.md`): the prose forbids them, the regex under *Filename pattern (binding)* admits digits anywhere in the descriptor. This PR resolved it toward the regex and dropped the test case. That is a vendored MIRROR standard, so the amendment is an upstream handoff rather than a local edit — surfacing it here rather than filing it, since the README's own framing is that a divergence is a question to raise upstream.

- **The "predict the failure count before mutating" instruction worked, and the "say which of two things a miss was" follow-up is what made it productive.** Three misses here: two were my model of the fixture, one was a genuinely weaker guard (`Est. 2.5 hours`). Without the second instruction I would have adjusted the prediction to match on all three and reported four clean controls. Recording that it earned its place.

- **The registry pattern (`MAY_NOT_OBSERVERS` / `DISAPPEARANCE_OBSERVERS` + AST discovery) is the single highest-leverage thing in this tree and a new workflow inherits it for free.** Writing the authorization table forced me to answer "what observes this?" eleven times, and one answer was honestly `JUDGEMENT`. Two of the eleven mechanisms did not exist until the registry demanded them. Worth considering whether the build and research families should declare tables too — they have prohibitions in their prompts and no registries, so the class this apparatus closes is still open on two thirds of the fleet.

Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Widened three guards onto a new `plan_docs` reader — every top-level markdown file — rather than NARROWING the write grant to `roadmap.md` plus the phase pattern. Both close the same hole. Alternatives: narrow the grant, so an unauthorised `notes.md` fails as a boundary crossing; do both. Why: narrowing introduces a new failure mode for a run that writes a legitimate extra file, and it reports a naming error as an authorization error, which sends the operator to the wrong rule. Widening the guards adds no new way to fail a correct run and makes guard scope equal grant scope by construction. **This is the decision most worth a second opinion** — it leaves the grant deliberately wider than the two file kinds the prompt names, on the argument that the guards, not the grant, are where the specificity belongs.

- **[Low]** Merged `origin/main` into the branch rather than leaving the PR one commit behind. Alternatives: leave it and let GitHub merge; rebase. Why: `#90` and this branch both edited `docs/file_structure.txt`, and a *textually* clean merge can still land a tree the set-equality test rejects — which would go red on `main`, not here. `git merge-tree` said clean, and the merged state is what the 3091-test run and the green `suite` check were actually taken against. Rebase was not an option: the branch is pushed, so it would need a force-push.

- **[Medium]** Kept `_HOURS`' `{0,24}` window rather than widening it alongside the sentence-boundary fix. Alternatives: widen to ~40 to catch *"Estimated cost, based on prior sprints and team velocity: 40 hours"*, which code-reviewer raised as Info. Why: nothing has measured either value, the six negative controls are calibrated against this one, and a wider window buys recall by spending exactly the precision the fix above just recovered. The limit is now stated in the code rather than left to be discovered. Reversible in one character if a real run misses an estimate.

- **[Medium]** Rejected the "the standards-auditor's clean pass is in tension with two Criticals" framing, and said so in the PR body instead of treating it as a finding. Alternatives: file it as a gap in the standards corpus. Why: the standards measure structure and these defects were semantic; asking `standards-auditor` to catch an unescaped regex is asking it to be `code-reviewer`. The lens split worked. What IS a real gap — that a registry entry is never checked against the behaviour it claims — is filed as C-086, which is a different statement.

- **[Medium]** Placed ONE candidate (C-086) covering findings that arrived as four separate defects. Alternatives: a row per defect; none at all, on the grounds that all four are fixed here. Why: they share one remedy — *a registry proves a mechanism exists, never that it fires on the case it names* — and one remedy is one item. Filing four would be four issues against one file from one pass, which this repo has measured as how a queue stops being read. Filing none would lose the generalisation, and the same property holds on `plan-sprint` and `triage-candidates` today.

- **[High]** Moved the phase-identity check ahead of `grants_that_vanished` rather than making the generic message phase-aware. Alternatives: leave the order and append the identity sentence to the vanished message. Why: the ordering comment's stated reason — every READER below assumes its file is still there — is satisfied either way, because the identity check parses nothing. Reordering keeps one message per prohibition; the alternative grows one message to cover two.

- **[High]** Reported the vacuous-test finding against my own additions as prominently as the ones I inherited. The first end-to-end harness stubbed `worktree_state` to `{}`, which zeroes the exact guard the ordering test claims to be ordered against — it passed while asserting nothing. Why it is in the log rather than quietly fixed: it is the same class as the four defects this pass was correcting, found in work written minutes earlier by the reviewer who had just written that class up.

## Deferred Work

- **A gate requiring every `MAY_NOT_OBSERVERS` row to be exercised, not merely resolvable** — Why deferred: it is a repo-wide capability spanning three workflows' registries and the shared `observer_registry` probe, not a fix to anything this PR built; every defect it would have caught is FIXED here. Tracked at: `docs/standards/architecture/research/candidates.md`, row **`C-086`**, `decision` blank, `status: open`, `component: workflow-decomposition`. Verified by: `git show HEAD:docs/standards/architecture/research/candidates.md | grep -c '^| C-086'` → `1`, and the row reads *"Require every `MAY_NOT_OBSERVERS` row to be EXERCISED by a test that makes the offence and…"* — **on this branch; it lands at merge and not before.** It is a PROPOSAL — capability that would be added — so `candidates.md` is its home and an issue is not.

**One item is SURFACED with no pointer, deliberately:** `run_plan_sprint.py` and `run_triage_candidates.py` carry the same dry-run/live `values` duplication fixed here. Neither is in this diff, neither is wrong today, and both are one added placeholder away from a bug this family has already shipped once. There is no verified home for it and I did not invent one — it is stated plainly in the PR body for `review-pr` and the operator to dispose of.

Nothing else is deferred. Every other finding from the four reviewers is fixed, or rejected with its reasoning recorded above.

## Post-Run Reflection

**Friction**

- **Stale `__pycache__` produced a false control during mutation testing, and it read exactly like a real regression.** After restoring a mutated file from a byte-identical copy, one test failed with an assertion showing the *mutated* behaviour, against source that `diff` proved unchanged. I spent a debugging cycle instrumenting the workflow before clearing `__pycache__` fixed it. **The failure mode is specific to the mutate-restore loop this prompt mandates**: rapid `cp` restores can land inside the mtime resolution the bytecode cache keys on. Every mutation above was re-run with the cache cleared and all six then hit their predictions exactly — including one that had reported `2 failed` against a prediction of `1`. **Worth a line in the mutation instructions: clear the bytecode cache between mutation and control, or a restored tree can lie to you in both directions.**

- **The `sleep` block fired on a legitimate CI wait and the suggested remedy was in the error text, which was good.** Recording only that `until ! gh pr checks N | grep -q pending; do sleep 20; done` is the shape that works, since the first thing reached for — `sleep 45; gh pr checks` — is blocked and the block message's example is abstract.

- **The negative-control methodology has a structural blind spot worth naming, because it is not obvious and it produced four defects here.** Every control mutates a guard and counts failures — which proves the tests discriminate against the code AS SCOPED. It cannot see a guard whose scope is wrong, because mutating a narrow guard still fails the narrow tests written for it. All four defects this pass fixed were scope defects: a regex crossing a sentence, a grant crossing a directory, a sweep missing a filename class, an ordering hiding a message. **The technique that found every one of them was asking what the guard does NOT look at** — which is C-086, filed. The draft's four controls all hit, and that was true and insufficient at the same time.

- **`test_relative_links_resolve.py`'s vacuity guard is the reason a two-week-old blindness surfaced at all**, and the convention deserves restating: an assertion over an empty set is indistinguishable from full coverage. Re-verified this pass — the old filter returns 0 files from a worktree and 354 from the operator's checkout, same commit. Worth the sweep the draft suggested for other absolute-path filters in the testing tree.

- **The instruction to mutate what *I* added, not only what I inherited, paid for itself on the first try.** My own end-to-end harness stubbed the guard it was written to test into oblivion and passed. Nothing else in this prompt would have caught it: it is a new test, it went green, and the defect it hides is invisible in a diff. Recording that it earned its place, in the same words the last pass used about a different instruction.

- **"Doubt your own invocation before you doubt the claim" needs a sibling: doubt your own TOOLING before you doubt the code.** The stale-cache episode had every signature of a real regression — a specific assertion, a reproducible failure, source that provably matched a known-good copy — and the correct diagnosis was that the interpreter was not running the file I was reading. The existing instruction covers shell quoting and `type -a`; the bytecode cache is the same class one layer down and is not mentioned.

- **The three narrow-lens agents disagreed usefully and the `standards-auditor` clean pass was informative rather than empty.** It arrived beside two Criticals from `code-reviewer` on the same files, and reading them together is what made the point that this PR's conformance was structural while its defects were semantic. Worth keeping the lenses separate rather than nudging the auditor toward correctness; the pairing is the signal.

Pumapumapumas:
## Decision Log

Low-confidence first.

- **[Low]** Fixed `phase_docs`' suffix by tightening the READER rather than by narrowing the deliverable guard to ask `_PHASE_FILE.match` directly. Alternatives: leave `phase_docs` permissive and make each of its two consumers re-filter; make the deliverable guard count conformant names only. Why: the two consumers want the same answer, and re-filtering at each call site is how the first refine pass's scope defects happened — a guard remembering to narrow is a guard that eventually forgets. **This is the decision most worth a second opinion**, because it leaves `phase_docs` and `plan_docs` deliberately UNEQUAL (wide stem, same suffix) and the reason is stated in a comment rather than enforced by anything. The class check pins the suffix agreement; nothing pins the stem disagreement.

- **[Low]** Kept the `(number, sub-letter)` collision rule at four cases rather than flagging every same-number pair among new files. Alternatives: flag any repeated bare number and treat rule 6 as an operator override; flag nothing lettered. Why: `phase5a_` + `phase5b_` planned in one pass is the carve-out the standard explicitly permits, and a guard stricter than the binding pattern fails work the standard allows — which this workflow cannot fix, because renaming is out of scope. The four cases are each pinned by their own test so the next reader can see where the line is.

- **[Medium]** Rejected `refactoring-evaluator`'s Medium (hoist the duplicated authorization regex literals) **after verifying its premise by execution rather than arguing about it**. Its stated consequence was a *silent* authorization gap; a typo'd copy of the candidates grant makes `boundary_crossings` report the run's own legitimate append as a crossing, so the run fails loudly on its first act. Alternatives: hoist them anyway as cheap insurance. Why not: it widens the diff into `triage_candidates_workflow`, which this PR does not otherwise own, to buy a guard that already fires.

- **[Medium]** Closed the dry-run duplication in the two SIBLING runners rather than re-surfacing it. Alternatives: leave it surfaced, as pass 1 did — neither file is in this PR's original scope and neither is wrong today. Why: pass 1 surfaced it with no tracker, and a surfaced item that reaches only the next reviewer becomes a HOLD and another dispatch cycle. I am the actor that can both find and fix it in one pass. The scope cost is two workflows edited; the extraction is mechanical and all three dry-run paths were exercised.

- **[Medium]** Added the `try`/`except` in `plan_project`'s loop over pre-existing behaviour I verified is not this PR's. Alternatives: file the whole thing as a candidate and change nothing, on the grounds that `run_plan_project` has never had a handler on `main` either. Why: this PR ADDS a third raise site inside that loop, and the failure it adds is the one whose recovery story is worst — research succeeded, so nothing downstream will ever re-offer the component. Splitting it into "the defect is mine, the capability is not" is what produced the fix here and C-087 there.

- **[High]** Made the hour guard a delta rather than widening `_HOURS` to exclude the shapes that were false-positiving. Alternatives: exempt files the run did not modify; keep the state scan and accept that a component carrying a legacy estimate must be hand-edited once. Why: the prohibition is *this run may not size the work*, and a guard reading the tree answers a different question. Every other prohibition in the chain was already a delta; this was the odd one out, and making it match removed a class rather than an instance.

- **[High]** Reported the M5 prediction miss as a fixture-model error rather than adjusting the prediction to match. Alternatives: write "6 mutations, 6 hits" — the two runner checks are disjoint, so 1 was always the right answer and I could have said so afterwards. Why it is in the log: the instruction to say WHICH of two things a miss was is the only thing that distinguishes a calibration failure from a weak guard, and M6 exists solely because the miss made me ask whether the second check was vacuous. It was not, and I would not have checked.

## Deferred Work

- **Resume `plan-project` on the PLANNING deliverable, so a component with research complete and no `roadmap.md` re-enters the loop** — Why deferred: the DEFECT half is fixed here (the parent now tells the operator the component is orphaned and gives the standalone recovery command); what remains is a capability the resume classifier has never had, spanning `plan_project_activities`' scaffolder and its marker contract. Tracked at: `docs/standards/architecture/research/candidates.md`, row **`C-087`**. Verified by: `git show HEAD:docs/standards/architecture/research/candidates.md | grep -c '^| C-087 '` → `1`, and the row's cells read `component: workflow-decomposition`, `decision: ` (blank), `status: \`open\`` — **on this branch; it lands at merge and not before.** It is a PROPOSAL — capability that would be added — so `candidates.md` is its home and an issue is not. **Explicitly NOT an expansion of C-083**, and the row says why: C-083 has `research-verify` record its own success, and on this scenario that marker would correctly read *verified* and still skip the component, because the step that failed was the one after it.

Nothing else is deferred. **The one item pass 1 surfaced with no tracker — the dry-run/live `values` duplication in `run_plan_sprint.py` and `run_triage_candidates.py` — is FIXED in this pass**, with a class-keyed AST check over every runner. Every other finding from the four reviewers is fixed, or rejected with its reasoning recorded above and verified by execution.

## Post-Run Reflection

**Friction**

- **`origin/main` moved three times during this pass, and the last move landed a gate that this PR failed.** A new `test_prompt_budgets.py` requires every prompt over 8 KB to carry a budget line; `plan_feature.md` is 17,821 bytes and had none. Merging `main` a second time immediately before pushing is the only reason it was caught here rather than turning `main` red. The prompt's own instruction to re-check `origin/main` at Stage 5 is written as an id-collision guard, and it earned its place for a completely different reason — **a gate that lands while a branch is open is a gate the branch has never run.** Worth widening that instruction from "re-check for collisions" to "merge and re-run, because main may have acquired a check you have not met."

- **The candidates tally gate caught me being right about arithmetic and wrong about the world.** I appended C-087, incremented thirty-seven to thirty-eight, and the gate failed with *"check whether a row went missing rather than assuming it did not."* A row had not gone missing — the merge had landed a triage RULING, so a previously-blank `decision` was now filled. The count was 37 either way. **The failure message told me to re-measure rather than re-derive, and re-measuring is what found it.** That phrasing is doing real work and is worth copying into other prose-versus-artifact gates.

- **`grep` on this machine is a shell function, not the binary.** An early sweep returned `ugrep: warning: … No such file or directory` for a path that existed, which read like a missing file rather than a CWD problem. `type -a grep` (which the prompt already mandates) showed the shim; `command grep` was used for everything load-bearing afterwards. Recording that the existing instruction fired correctly and cheaply.

- **A `cd` into a subdirectory persisted across Bash calls and silently broke the next sweep**, which returned "No such file or directory" for four paths that all exist. The prompt warns about exactly this. What made it survivable was that the failure was loud; what made it cost a turn was that the error names the paths rather than the CWD.

- **Every defect found on this PR across three passes has been a SCOPE defect, and the negative-control methodology structurally cannot find them.** A control mutates a guard and counts failures, which proves the tests discriminate against the code AS SCOPED — mutating a narrow guard still fails the narrow tests written for it. Pass 1 named this and filed C-086. **This pass is the evidence that naming it was not enough:** the technique that found all three defects here was asking *what does this guard NOT look at*, and it found them in code pass 1 had just finished hardening. The three class checks added here (`test_a_PRE_EXISTING_violation_of_a_PROHIBITION_does_not_fail_a_clean_run`, the lookalike parametrize, and the runner AST sweep) are the first ones in this tree keyed on the class rather than the instance, and each one is written to fail on a *future* member. Whether that shape generalises to the build and research families is the interesting question and I have not tested it.

- **A guard that fires when it should not is as much a defect as one that stays silent, and this suite could only express the second.** `_drive` asserted `pytest.raises` unconditionally, so no test in the module could show a run SUCCEEDING. Splitting `_harness` out is three lines and it is what made the pre-existing-violation class check expressible at all. Worth checking whether the sibling workflow test modules have the same one-directional shape — `plan_sprint` and `triage_candidates` both have long guard chains and no equivalent.

- **"Mutate what YOU added, not only what you inherited" earned its place again, in a different way than last time.** Pass 1 reported it catching a vacuous test it had just written. Here it caught something subtler: my prediction for M5 was wrong, and chasing WHY produced M6, which is the only evidence that the second of my two class checks is not vacuous. **The instruction's value was not the mutation — it was that predicting first made a miss legible.** A run that mutated without predicting would have seen "1 failed" and moved on.

- **The instruction to clear `__pycache__` between mutation and control, added from pass 1's reflection, was followed and nothing went wrong.** Recording the negative result explicitly, because a instruction that prevents a failure leaves no trace that it worked, and the next reflection has no way to say so unless somebody does.

- **The three-narrow-lenses-then-quality-control shape produced findings that did not overlap at all**, which is worth recording as a positive control on the split: `code-reviewer` found a comparator bug, `refactoring-evaluator` found no High and rejected a rewrite with reasoning, `standards-auditor` returned clean with one self-labelled note, and `quality-control` found the only cross-module finding — an integration seam three files wide that no single-module lens would have crossed. **The instruction to tell each agent it has no shell and that I would run commands for it worked**: none of the four opened with a paragraph about the missing shell, and three handed back specific commands, two of which confirmed findings I would otherwise have had to re-derive.

PR #124, opened 2026-08-19
build-draft: Phase 2 family alignment — a ruling procedure, validated blind at kappa 0.000, and a frozen baseline emptied

Pumapumapumas:
# Phase 2 — Family alignment

Delivers all five requirements of [`phase2_family_alignment.md`](docs/development/workflow-decomposition/phase2_family_alignment.md). **Requirement 3 was allowed to fail and it did** — the blind trial measured κ = 0.000, so ruling moved from per-pair to per-family, which is the fallback the phase named in advance.

## Eleventh pass — `build-refine`: the guard shipped to close false safety prose was carrying three claims of its own that nobody drove

The tenth pass closed the fail-open **correctly**. I re-derived that rather than taking it: five mutations, each predicted before running, none missed — restoring the `= None` default alone (1 failure), default plus skip branch (2), the required parameter kept with a defensive `None` branch re-added (1, the property test alone), the same default on `wait_for_ci` (1), and the brief's `(a)(ii)` implemented faithfully (2). `routing.py` is absent from the diff, so `ci_gate`'s six states, notes and HOLD kinds are untouched by construction. Every claim-shape grep for a surviving copy of the old false sentence came back clean.

**What it also shipped was three coverage claims in the new guards' own prose that nobody had executed.** All three are the same shape as the defect this PR exists to close — a sentence telling the next reader that something is checked, in a file readers are trained to trust — and all three are now closed. They are not three coincidences: the tenth pass wrote the guards' prose ahead of their assertions.

### 1. The property test's own docstring promised something the brief had explicitly forbidden it to promise

Its docstring said the guard would outlive the mechanism — *"swap the required parameter for a stopping verdict and this still passes."* **Driven: it does not.** Implementing `(a)(ii)` exactly as the brief describes it (`repo_root: Path | None = None` plus `if repo_root is None: return UNREADABLE_POLICY`) turned `test_a_NON_HOLDING_gate_is_unreachable_without_a_policy_READ` **red**.

The cause is one line of scope. The brief's property is *"no path returns a **NON-HOLDING** verdict without having read a check policy"*, but the assertion sat outside the `hold is None` branch, so it fired on **every** return — pinning the mechanism while its docstring disclaimed doing so. The brief's `(b)` had named this in advance: *"Do not replace one unverified safety claim with another — whatever the new sentence asserts must be the thing `(c)` guards."*

**Now scoped to the property, and both directions measured rather than asserted:**

| Mutation | Predicted | Observed |
|---|---|---|
[...]

PR #123, opened 2026-08-19
plan-feature: PMP — one new phase for run identity, and four things the surrounding replans invalidated

Pumapumapumas:
# PR #123 — PMP Phase 9, and four things the surrounding replans invalidated

**What this PR is:** a new PMP phase doc — `persistent-memory-protocol/phase9_one_run_one_identity.md`, **160 lines, the largest artifact in the diff** — plus the corrections its arrival forces on two components' planning surfaces. Three review findings, one defect found by verifying the task's own premises, the propagation gap peer review caught, the ordinal-citation *class* a later pass closed, `main`'s move to random candidate ids, and — from the final correction pass — **four false claims about shipped artifacts** that seven consistency sweeps could not see.

## Planning changes

**`ti-residual-cas-two-carriers` — a false premise, corrected against the code**
- `temporal-integration/roadmap.md` § *What is deliberately not built* claimed compare-and-swap "does not survive, because the bag's create is already atomic". **That premise is false.** `open_bag`'s docstring (`scripts/workflows/temporal/modules/journal/bag.py`) states bag creation is *"three syscalls, not one"* — the `mkdir` is atomic, creation is not; sequential retry is safe, *"a true simultaneous race is not"*.
- Restated as a relationship rather than a second ownership claim: **PMP Phase 9 r7 is the MECHANISM**, **TI Phase 2 requirement 4 is the CONSUMING requirement**. `bag.py` cited so the next reader does not re-derive it.
- `temporal-integration.md`'s supersession section now names the consumer half, so the two documents point at each other.
- One clause on Phase 2 requirement 4 naming r7 as what it depends on.

**`ti-phase6-cites-renumbered-migration-step` — the citation, and then the whole class**
- `phase6_the_rest_of_the_fleet.md:46` cited migration "step 3"; the layout step moved to 4 when dispatch identity was inserted as 2. Rewritten to cite by content, with the reason stated inline.
- **The cross-file sweep found exactly one sibling** — that one — and the class stayed open, because `temporal-integration.md` cited its OWN migration path by ordinal from **six** places: `:11` (×1), `:54` (×4), `:77` (×1). All six were correct on the day and all six carried the exact fragility `phase6:46` was rewritten to remove. A prior pass counted four and declined to fix a fragment; the count is six — **line 54 carries four, not two**.
[...]

PR #122, opened 2026-08-19
research-minor: Temporal activity retry boundary (gh() vs RetryPolicy)

Pumapumapumas:
## Summary

**This is a MINOR cycle: one paper plus a rewritten synthesis.**

The Temporal Integration research pool (`docs/development/temporal-integration/research/`) inherited five papers on 2026-08-19 when the "Fleet Reliability" sprint section dissolved. This run:

1. Read the Temporal Integration sprint entry (`docs/development/sprint.md:175-193`) and cross-checked all five inherited papers' `Feeds:` lines and the section's own open milestones against current coverage.
2. Wrote **one new paper**: `raw/activity_retry_boundary.md`.
3. Corrected one stale `Feeds:` line (`durable_dispatch_identity.md` — its milestone survived and moved into Temporal Integration unchanged; only the pointer was stale).
4. Rewrote `synthesis.md`, scoped to what this pool cycle actually feeds.

## Why this topic

**Question:** How does Temporal's Python-SDK activity retry mechanics (`RetryPolicy`, `non_retryable_error_types`, and specifically what actually triggers a retry — a raised exception vs. a returned `ActivityResult`) compose with this fleet's existing in-process bounded retry for `gh` CLI calls, once those calls move inside a Temporal activity?

**Destination:** `sprint.md:189-190` — "Rule the retry boundary before wrapping anything" and "Reduce `gh()`'s own retry when wrapping" — both currently unchecked, both explicitly named in the sprint's own text as needing a decision.

**Why this and not another gap in the same section:** I checked every open Temporal Integration milestone against existing coverage first:
- "A `claude_cli` activity domain" (heartbeating + payload limits) — **already fully closed** by the upstream product paper `python_sdk_long_activities.md` (definitive, first-party sourced, its own `Feeds:` line names this exact milestone). Nothing to add.
- "A restart-recovery contract" — already covered by `durable_dispatch_identity.md`, already in this pool.
- "Prove an invocation is indistinguishable from an operator at a terminal" — the PERMISSION half is answered upstream by `anthropic_tos_and_enterprise.md` §1.5 (headless mode is ToS-sanctioned). **The DESIGN half is open, and the sprint line says the design half is the question** — *"Not a permission question; a design one."* A separate billing-classification question also remains, in that paper's §9 (which its own header marks NOT critic-verified). *(Corrected by the verify pass: the original wording called this milestone "substantially answered," which overstated the evidence, and pointed at a §7.2 that does not exist in that paper.)*
[...]

PR #126, opened 2026-08-19
triage-candidates: judge a candidate against the TRAJECTORY, then size it

Pumapumapumas:
TWO CHANGES, AND THE FIRST IS THE IMPORTANT ONE.

1. THE WORTHINESS TEST. Triage's whole test was "can this be scheduled?" — a
READINESS question. It asks whether a candidate is actionable and says nothing
about whether it is worth doing, so a perfectly schedulable candidate serving
nothing this platform is building passed it cleanly.

A `ship` promotes a concept into committed work: a component directory, a
research cycle, a planning chain, a sprint entry, and a claim on the operator's
weeks. The prompt now asks the hard question FIRST — does this serve the
trajectory — and names where the trajectory is written: problem-statement.md,
the project research pool and synthesis, sprint.md, and the roadmaps it touches.
Failing that is a `reject`; passing it and failing schedulability is `requires
review`. The Note must say WHICH of those was weighed, because "serves the
thesis" is an assertion and "serves differentiator 2, and no sprint covers it"
is a ruling somebody can check.

2. THE `size` COLUMN — triage's second ruling, asked ONLY of a `ship`.
`feature` / `phase` / `checkboxes`, best-guess at concept stage.

WHY IT NEEDED A COLUMN RATHER THAN AN INFERENCE. `plan-candidates` derived size
from a proxy: if the named component's directory did not exist, the candidate
must be a new component. Right for a `feature`, wrong for the other two — a
phase-sized candidate naming a new component got a whole component scaffolded
around one phase of work, and a checkbox-sized one had no expressible form at
all. A proxy correct in one case of three reads as a decision and is an accident.

Now: `feature` scaffolds and runs the full chain; `phase` and `checkboxes`
decline with a note saying that is the CORRECT outcome, not a skip.

BLANK IS SKIPPED, NEVER GUESSED, and that is what makes the backfill
self-healing. The 29 rows already ruled `ship` carry no size; each heals when
triage next reaches it, and until then `plan-candidates` reports them rather
than scaffolding components nobody asked for. No migration.

THE PAIRING IS GUARDED, not the columns separately. `sized_without_shipping`
reads both cells from ONE row: checking each column alone would pass a table
where every row is sized and none is shipped — both legal in isolation, the
combination nonsense. The prompt asks two questions in order, and a size beside
a rejection means the run stopped reading its own first answer.

`size` is triage's alone. `plan-feature` and `plan-verify` carry it in their MAY
[...]

PR #127, opened 2026-08-19
research: a paper's `Feeds:` target must resolve, and `git status` is not `git ls-files`

Pumapumapumas:
MINED FROM PRs #122, #123 AND #124. **Correction pass 2026-08-22 closed all five `review-pr` pass-1 findings, and a `build-refine-minor` peer-review pass the same day closed four more against the correction itself. What this PR now contains is described below, section by section.**

## 1. The `Feeds:`-liveness guard

Requested by FIVE runs on one PR and by the reviewer on a second. PR #122's two producing runs, its correction pass and both review passes each independently proposed this check; #123's run hit the same class from the other side. When five actors reach for the same instrument, the instrument is the finding.

`Feeds:` is what makes a paper's destination checkable — the Research Standard's framing is that a paper with no destination is not in scope. **A DEAD `Feeds:` IS WORSE THAN A MISSING ONE**: it reads as a destination, so a planner follows it, finds nothing, and cannot tell whether the paper was misfiled or the target moved.

MEASURED: all 36 papers in the pool carried a header `Feeds:` and **15 pointed at a file that does not exist — a 42% rot rate.** None is a typo; every one is a target that moved. (This first read *35 of 36*, which was the guard's own blind spot talking — see §4.)

### Six were repointed, nine stay frozen *(finding 1)*

The baseline froze fifteen on the grounds that *"each needs somebody who knows where that evidence was actually meant to land"*. That was true for nine and **false for six, whose correct target the baseline's own comment already named**:

- `problem-statement.md` ×4 — three repo-root-relative, one missing the `docs/` prefix. The file is at `docs/standards/architecture/problem-statement.md`.
- MMF ×2 — the narrative doc that was never written became `docs/development/memory-management-framework/roadmap.md`.

**The `§` anchors were repointed in the same edit**, because a correct path with a dead anchor is the invisibly-wrong pointer the freeze exists to avoid, and this guard does not check anchors. `problem-statement.md` was substantially reworded on 2026-08-20/21, **after** the review that named this finding — so the anchors were matched against the file's headings as they are now, not as the review quoted them:

| The paper said | The heading is now |
|---|---|
| `Where this repo sits` | `Where this sits: read this first` |
| `Why coding is the first edge, and not merely the earliest` | gone; the argument is claim #3 of `Where we actually differ` |
| `Affordability is not a footnote — it is the enabler` | `Affordability is the enabler` |
| `Where we actually differ` #2, *"Edges are dedicated and non-fungible"* | reworded to *"Work is bound to a machine by identity"* |
| `Where we actually differ` #1, the three-tier trust model | unchanged — verified, not assumed |

MMF's anchors were resolved against `roadmap.md`'s own § *Where the research contradicted the sprint milestones*, which maps the sprint milestone names those two papers quote onto the phases that absorbed them.

The remaining nine (8 × `docs/development/roadmap.md`, 1 × an ambiguous bare `synthesis.md`) genuinely need an operator ruling and are deferred to `C-9yi1yv2h`.

### The ratchet tells a repair from a deletion *(finding 3)*

Deleting a paper's `Feeds:` line drops its entry out of `_dead()` **exactly as a real repair does**, so the ratchet reported *"these baseline entries now resolve — the pointer was repaired or the target came back. Delete their lines."* That statement is false and the instruction launders the rot: the baseline can be shrunk by destroying the evidence instead of fixing it.

`_declared()` now keeps three answers apart — a live pointer, a paper with no `Feeds:` line, and a paper that is gone — and `_ratchet()` splits the entries that left the baseline into REPAIRED and GONE, with its own test and its own message for the second. `_ratchet()` is pure over already-read inputs so every arm is driven on synthetic data in `test_the_ratchet_DISCRIMINATES`; a classifier only ever exercised against today's answer has never been tested.

**Driven red, all three arms, then restored.** Deleting `hermes_assessment.md`'s `Feeds:` line now reports the deleted-pointer message (it reported the repair message before this change); repointing it at a real file reports the repair message; breaking a repointed target reports new rot.

### Scope, stated narrowly and now PLACED *(finding 4)*

It proves a reader ARRIVES somewhere, never that arriving helps. Section anchors, wrapped continuation lines and `candidates.md` Notes carry the same class — and they are now a `candidates.md` row, **`C-gkuvhctk`**, `component: pointer-liveness`, rather than a docstring paragraph that dies at merge.

Measured while doing the repoint: **of the ten `§`/milestone anchors those six papers carry, four were live.** The file-level guard was green on all ten. Two further dead targets are invisible for a second reason — `_FEEDS` matches one line, so a target on a wrapped continuation line is never read: `backbone_edge_generality.md` names a `system-overview.md` that has never existed anywhere in this tree, and `dedicated_edge_routing.md` names the same `docs/development/roadmap.md` that eight baselined papers are frozen for. Neither is in the baseline, because the guard cannot see them to count them.

Not built here, deliberately: fixing continuation-line blindness would GROW the frozen baseline, which its own header forbids.

## 2. `git status` does not prove a file is in the commit *(finding 2)*

The ignore check has every run confirm new files show as untracked — correct for the ignore question and silent on the one that follows. Several guards here read TRACKED files rather than the working tree, so a new file passes every local check while untracked and turns CI red the moment it is committed. The original run did this to itself: full suite green on an untracked new test file, committed, CI failed on the map gate.

**The `git ls-files` half now lives in the shared fragment, not in `plan_feature.md` alone.** The first version of this PR left it inline and deferred the promotion on the grounds that *"PR #124 is promoting that exact block to the shared pool… it lands there once, after that merges"*. **#124 merged as `a078c41` and is in this branch**, so the deferral's own condition was already met — the premise was reversed by this branch's own rebase, and only the PR body recorded it.

- the paragraph moves into `assistant/prompts/gitignore_collision_check.md` (555 B → 1309 B, and 1590 B after the refine pass reworded the staging step — still under the 8,000 B budget `FLOOR`, so no budget line is owed)
- `plan_feature_workflow` supplies `GITIGNORE_COLLISION_CHECK`; `plan_feature.md` renders the placeholder and its **divergent** inline first paragraph is deleted rather than left beside the shared one
- `_FRAGMENT_FLOOR` 1 → 2, measured — the fragment is two paragraphs, so the floor is again the whole of it
- `plan_feature.md`'s budget **re-measured downward**, 24,482 → 23,437. Keeping the old ceiling would silently re-grant the 1,045 B the promotion freed

`build_draft`, `build_draft_minor` and `plan_revision` all render this fragment, so all three now get the complete check. This is what `fork_vs_parameterize.FAMILY_RULINGS`'s operational-safety ruling already required: *"a cheaper or differently-jobbed run is not a run permitted to be less careful"*.

**Driven red, then restored:** dropping the placeholder from `plan_feature.md` fails the render check; cutting the fragment back to one paragraph fails the floor at *"1 lines, floor 2"*.

## 3. `origin/main` was merged in

`review-pr` pass 1 correctly recorded the branch as fully rebased. Main had moved **10 commits** by the time this pass started, the PR had gone `CONFLICTING` on `docs/file_structure.txt`, and two of the five findings are unbuildable against the stale tree: `problem-statement.md`'s headings were reworded on 2026-08-20/21, and candidate ids were migrated from `C-NNN` to random base36 on 2026-08-21 (`2af48df`). **A merge commit rather than a rebase** — no pushed history is rewritten and no force-push is needed, which matters because nothing in the task authorised one.

Two merges were needed: `#136` and `#123` landed mid-pass and added two `candidates.md` rows. Both conflicts are resolved by keeping **both** sides in full — the map entries, the candidate rows, and the derived untriaged count (68 → 70 on main, 71 with `C-gkuvhctk`).

## 4. What the peer-review pass changed *(`build-refine-minor`, same day)*

One `code-reviewer` lens against the correction above. Two of its findings were live defects in the guard this PR ships, and both are fixed here rather than filed.

- **The guard was silently skipping a paper.** `_declared()` read `splitlines()[:20]`, and `cross_node_memory_protocol.md` carries a 34-line SUPERSEDED banner that pushes its header `Feeds:` to **line 45**. The guard recorded it as *"declares no destination"* — **the identical value a DELETED pointer produces**, which is the one state §1's ratchet split exists to treat as evidence-destruction. A long enough banner bought a paper a silent exemption from `test_no_NEW_dead_feeds_target`. The whole file is searched now; **measured, it changes no verdict** — nine dead targets either way — so the frozen baseline cannot grow on this. Negative control: a dead target placed on that line-45 `Feeds:` now fails the guard and was invisible to it before.
- **Every verdict in that module is keyed by BASENAME**, `ACCEPTED` included, across five separate research pools. No collision exists today and nothing enforced that; an assertion now fails loud on the day one is introduced, rather than letting the wrong paper be judged repaired. Negative control: a duplicated basename fails it, naming the file.
- **The promoted `git ls-files` half had nowhere to stand.** It said to check *"after you stage"*, but all four consumers stage and commit on a single `git add -A && git commit` line — so a run following the literal instructions either checks before staging, where `git ls-files` correctly reports nothing about a file that is fine, or splits the command on its own initiative. It now says to stage first, which is answerable in every tier that renders it.
- **`C-gkuvhctk`'s Note said two things that would not survive a triager checking them.** Its anchor arithmetic read as self-contradictory — the reviewer computed *three live* from the same sentence that states *four* — and it described `dedicated_edge_routing.md`'s continuation-line target as `docs/development/roadmap.md` when the paper names a bare `roadmap.md`. Both corrected in place; the row's count and `status` are untouched.

Rejected, with reasoning: **`backbone_edge_generality.md` keeps its pointer to `system-overview.md`**, a file that has never existed in this tree. Deleting it is the laundering §1 exists to stop and guessing its target is the invisibly-wrong pointer the freeze exists to avoid — it is the same class as the nine frozen entries the task reserved for an operator, and it is recorded as evidence on `C-gkuvhctk`.

## Files

- `testing/scripts/tests/unit/test_a_research_papers_FEEDS_target_resolves.py` — the guard, its baseline (15 → 9) and the split ratchet
- `docs/standards/architecture/research/raw/{backbone_edge_generality,dedicated_edge_routing,subscription_economics,edge_identity_trust}.md` — repointed
- `docs/development/memory-management-framework/research/raw/{dual_channel_outcome_records,non_model_observables}.md` — repointed
- `docs/standards/architecture/research/candidates.md` — `C-gkuvhctk`, and the derived untriaged count, which lands at **seventy-one** after the two mid-pass merges below
- `scripts/workflows/temporal/modules/assistant/prompts/gitignore_collision_check.md` — the complete check
- `scripts/workflows/temporal/modules/assistant/plan/plan_feature/{plan_feature_workflow.py,prompts/plan_feature.md}` — consumes it
- `scripts/workflows/temporal/tests/unit/test_promoted_fragments_render_for_every_consumer.py` — floor 1 → 2
- `testing/scripts/tests/unit/test_prompt_budgets.py` — budget re-measured downward
- `docs/file_structure.txt` — the guard's annotation, and the merge resolution

The refine pass touched three of those files — the guard, the shared fragment and `candidates.md` — and added none.

No files were added, removed or renamed by the correction pass, so `docs/file_structure.txt` needed only its annotation updated.

`./testing/run-all.sh unit` → **3214 passed, 3 skipped**, exit 0, re-run after the refine pass. All four CI checks on the pushed head are green.


Pumapumapumas:
## `review-pr` — disposition pass 1 (PR #127)

**Verdict rationale:** the two artifacts this PR ships are individually sound — the `Feeds:` guard's baseline is *exactly* correct (I measured 37 papers, 36 with `Feeds:`, 15 dead, 15 baselined; both ratchet directions drive as documented) and the full suite is green (`129 passed`, exit 0, plus the map gates). It holds on five items, of which two are consequential: **the baseline freezes six repairs whose answers it already wrote down**, and **the `git ls-files` half — the half whose absence the author personally turned CI red with — reaches exactly one of the four prompts that need it**, because this branch's own rebase reversed the premise of the deferral that parked it.

Mergeability read first: `MERGEABLE` / `CLEAN`, `closingIssuesReferences` empty (the body claims no closures — the field agrees). Branch is fully rebased onto current `origin/main` (`git merge-base --is-ancestor origin/main HEAD` passes), so nothing in this PR is being judged against a rule that landed after the cut.

**Sweeps run.** Asymmetric-presence (the headline finding came from it — F2). Deleted-artifact: no files/functions/guards removed; the one deletion is the `# +${FILING_A_CANDIDATE_ROW}` annotation on the budget line, which no guard reads — demoted to a note, no consequence statable. Completion-checkbox: **N/A**, this PR flips no boxes in any planning artifact. Merge-commit sweep: N/A, one non-merge commit. Boundary lens on the new chokepoint: produced F3 and F4. Current-tree check ran in **both** directions.

---

### Attestation

**This PR has zero comments and zero reviews** — no Decision Log, no Deferred Work, no Post-Run Reflection. All four sibling PRs carry them (#132: 14 comments, #133: 7, #126: 10, #124: 17, each with all three sections). The PR body is unusually thorough and does much of that work, but it is the body, not the attestation surface. **Everything below is reconstructed from the diff and the body — it is not attestation, and I am saying so rather than presenting it as such.** This is not a formality: the PR's one deferral lived only in body prose, and that is precisely how F2 went stale unnoticed.

---

### Disposition table

|---|---|---|---|
| `feeds-baseline-freezes-known-repairs` | correctness | **HOLD** (redispatch) | Six of the fifteen frozen entries have their correct target **named in the baseline's own comment**. `problem-statement.md` ×4 → the note says *"the file is at docs/standards/architecture/"*, and **that file exists**; MMF ×2 → the note says *"MMF's narrative doc became roadmap.md"*, and **`docs/development/memory-management-framework/roadmap.md` exists**. The stated reason for freezing — *"each needs somebody who knows where that evidence was actually meant to land"* — is true for the other nine and **false for these six**: the run already worked out where they land and wrote it down, then froze the pointer anyway. |
| `ls-files-half-reaches-one-prompt` | correctness | **HOLD** (redispatch) | The `git ls-files` half is in `plan_feature.md`'s **inline** copy only. `build_draft`, `build_draft_minor` and `plan_revision` render the shared fragment `assistant/prompts/gitignore_collision_check.md`, which does **not** carry it. The body defers this — *"PR #124 is promoting that exact block to the shared pool… it lands there once, after that merges"* — but **#124 merged as `a078c41` and is already in this branch**. Premise reversed by the PR's own rebase. |
| `ratchet-cannot-tell-repair-from-deletion` | correctness | **HOLD** (redispatch) | **Driven, not reasoned:** I deleted the `Feeds:` line from `hermes_assessment.md` and the ratchet failed with *"these baseline entries now resolve — the pointer was repaired or the target came back. Delete their lines"* — a false statement, and an instruction that launders the rot. Tree restored clean. |
| `anchor-liveness-proposal-not-placed` | scope | **HOLD** (redispatch) | The docstring surfaces anchor rot and `candidates.md` Note rot as the next axis and places **no `candidates.md` row**. A surfaced-not-placed proposal dies at merge. **Not hypothetical:** of the four anchors I checked, **two are already dead**. |
| `no-attestation-comment` | standards-implication | **HOLD** (redispatch) | Zero attesting comments against a four-for-four sibling norm. Demonstrated consequence: F2. |
| `feeds-baseline-dissolved-destinations` | deferral | **DEFERRED** | The remaining nine (8 × `docs/development/roadmap.md`, 1 × ambiguous bare `synthesis.md`) genuinely need an operator ruling. **Pointer: `docs/standards/architecture/research/candidates.md`, row C-116, status `open`** — verified by reading the row, which covers exactly this question and explicitly distinguishes itself from this PR's guard. |

**On the producing run's one deferral:** classified **MIS-DISPOSITIONED**, not laundered. The run disclosed openly, gave real reasoning, and was *correct when it wrote it*. A home existed (the shared fragment, editable now) and the rebase made it reachable. It is a deferral that should have become a fix — counted in neither `laundered_deferrals` nor `homeless_items`.

---

### WHAT HAPPENS NEXT

One dispatch closes all five. Fire `build_minor.sh --pr 127` once; the five scoped contexts below are all known fixes to known lines, and the guards (`test_prompt_budgets`, the fragment floor, the map gates, the `Feeds:` guard itself) verify every one.

1. **Repoint the six knowable entries** and delete their `ACCEPTED` rows. Fixing the path without the anchor would create the *"invisibly wrong pointer"* the run rightly warned about — so match each `§` heading against the target's current headings (`Where this repo sits` → `Where this sits — read this first`; `Affordability is not a footnote — it is the enabler` → `Affordability is the enabler`; the two `Where we actually differ` anchors are still live).
2. **Move the complete check into the shared fragment** and have `plan_feature` consume it, deleting its divergent inline copy. The fragment is 555 B against an 8,000 B floor, so no budget line is needed; `plan_feature.md`'s budget then needs **re-measuring downward**.
3. **Split the ratchet's two states** so a deleted `Feeds:` line fails loud instead of printing "repaired".
4. **Place the pointer-liveness proposal** as a `candidates.md` row with a named `component`.
5. **Post the Decision Log / Deferred Work / Post-Run Reflection.**

---

```yaml
pr_review:
  pr: 127
  run_id: 4f9716b381c247159b9af48f875e8c89
  pass: 1                            # verified: zero fence-anchored `pr_review:` blocks on the
                                     # thread (the PR has zero comments). Matches the dispatch label.
  attempt: 0                         # seeded per the pass-1 rule from the count of ATTESTING
                                     # reflection comments, which is ZERO. Flagged rather than
                                     # silently corrected: one commit did land, so this seed
                                     # under-reports — and that under-report IS `no-attestation-comment`.
  verdict: HOLD
  converged: false
  findings:
    - id: feeds-baseline-freezes-known-repairs
      title: Six papers ship with dead Feeds targets whose correct destination the baseline already records
      category: correctness
      consequence: >
        A planner following any of these six arrives nowhere — the exact harm the guard exists to
        prevent — while the repair sits written down three lines away in ACCEPTED. Because the
        ratchet only shrinks when a target starts resolving, the list will not move until somebody
        re-derives what this PR already determined. The reported 43% rot rate is inflated by six
        repairs the run had in hand.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: ls-files-half-reaches-one-prompt
      title: Three file-creating tiers still get the incomplete ignore check, so a new file passes locally and turns CI red on commit
      category: correctness
      consequence: >
        build_draft, build_draft_minor and plan_revision render the shared fragment, which carries
        only the `git status` half. Each creates new files and runs the suite before staging, so each
        can go green on an untracked file and fail the map gate once committed — the precise failure
        this PR documents its own author hitting. The repo's own merged operational-safety family
        ruling (fork_vs_parameterize.FAMILY_RULINGS) says this block must NOT fork per consumer:
        "a cheaper or differently-jobbed run is not a run permitted to be less careful".
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
      reviewed_sha: a078c41
    - id: ratchet-cannot-tell-repair-from-deletion
      title: Deleting a failing Feeds line satisfies the ratchet and prints that the pointer was repaired
      category: correctness
      consequence: >
        The cheapest way to make a failing Feeds entry pass is to delete the field the guard exists
        to protect. The ratchet then asserts "the pointer was repaired or the target came back" —
        false — and instructs the run to drop the baseline row, erasing the record of the rot.
        A gate passable by damaging the artifact it protects is worse than no gate.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: anchor-liveness-proposal-not-placed
      title: The next axis of pointer rot is surfaced only in a docstring, so it dies at merge while already live
      category: scope
      consequence: >
        The docstring names section anchors and candidates.md Notes as carrying the same class with
        "no watcher", and no candidates.md row was added — a surfaced proposal nobody places is
        indistinguishable at merge from one nobody had (C-116's own row records this exact outcome).
        It is not hypothetical: two of the four anchors I checked are already dead, so repointing
        paths alone leaves papers resolving to a file and not to the section they cite.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: no-attestation-comment
      title: The PR carries no Decision Log, Deferred Work or Reflection, so its one deferral had no surface on which to be re-checked
      consequence: >
        Zero comments against a norm every sibling PR meets (#132/#133/#126/#124 all carry the three
        sections). The deferral parking the ls-files half lived only in body prose, was invalidated
        by this branch's own rebase, and nothing re-read it — that is finding
        `ls-files-half-reaches-one-prompt`, so the cost is demonstrated rather than asserted. It also
        forces this review to reconstruct intent from the diff instead of reading the run's own words.
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: feeds-baseline-dissolved-destinations
      title: Nine baselined papers point at destinations that dissolved, and no rule says what state such a paper is in
      category: deferral
      consequence: >
        Eight papers cite `docs/development/roadmap.md`, which never existed, and one cites a bare
        ambiguous `synthesis.md`. Unlike the six above, no note names where they land — each needs an
        operator who knows which component roadmap the evidence was for. Left unruled, they read as
        current evidence for work that may no longer be planned.
      disposition: deferred
      pointer: "docs/standards/architecture/research/candidates.md — row C-116, status `open`, component `temporal-integration`"
      pointer_verified: true
  next_steps:
    - item: feeds-baseline-freezes-known-repairs
      kind: redispatch
      note: Repoint the six entries whose target the baseline already names, and delete their ACCEPTED rows.
      dispatch_tool: build_minor.sh
      dispatch_context: |
        Repoint EXACTLY these six papers' header `Feeds:` lines and delete their rows from
        ACCEPTED in testing/scripts/tests/unit/test_a_research_papers_FEEDS_target_resolves.py:

          1. docs/standards/architecture/research/raw/backbone_edge_generality.md
             `problem-statement.md` -> `docs/standards/architecture/problem-statement.md`
             Anchor `§ "Where this repo sits"` has drifted; the live heading is
             `## Where this sits — read this first`. Update the anchor to the live heading.
          2. docs/standards/architecture/research/raw/dedicated_edge_routing.md
             `problem-statement.md` -> `docs/standards/architecture/problem-statement.md`
             Anchor `§ "Where we actually differ"` is LIVE — leave it unchanged.
          3. docs/standards/architecture/research/raw/subscription_economics.md
             `problem-statement.md` -> `docs/standards/architecture/problem-statement.md`
             Anchor `§ "Affordability is not a footnote — it is the enabler"` has drifted; the live
             heading is `## Affordability is the enabler`. Update the anchor.
          4. docs/standards/architecture/research/raw/edge_identity_trust.md
             `standards/architecture/problem-statement.md` -> `docs/standards/architecture/problem-statement.md`
             Anchor `§ "Where we actually differ"` is LIVE — leave it unchanged.
          5. docs/development/memory-management-framework/research/raw/dual_channel_outcome_records.md
             `docs/development/memory-management-framework/memory-management-framework.md`
             -> `docs/development/memory-management-framework/roadmap.md`
          6. docs/development/memory-management-framework/research/raw/non_model_observables.md
             same rename as 5 (note: this paper's Feeds target is wrapped in backticks — keep them).

        For 5 and 6 the cited milestones are phase-level; point at roadmap.md and keep the milestone
        names in the prose. If a cited milestone cannot be located in roadmap.md, still repoint the
        PATH and say in the paper that the milestone name is unconfirmed — do NOT invent one.

        DO NOT TOUCH: the other nine ACCEPTED rows (they are correctly frozen and are deferred to
        candidates.md C-116); the guard's resolution logic; the test's docstring measurements.
      precheck: |
        Run on the PR branch `feeds-liveness` (NOT main — these papers' repairs live in this
        unmerged PR's baseline), from the repo root:

          test -f docs/standards/architecture/problem-statement.md \
            && test -f docs/development/memory-management-framework/roadmap.md \
            && echo BOTH_TARGETS_EXIST

        This tests the PROPERTY the finding rests on (the named destinations resolve), which is a
        DIFFERENT check from the one that surfaced it (reading the ACCEPTED comment text).

        STOP PREDICATE, split:
          · BOTH_TARGETS_EXIST does not print -> the repair is NO LONGER KNOWABLE. This is
            "not yet warranted", NOT "already done": do not guess a destination. Report and stop.
          · All six keys are already absent from ACCEPTED and the six papers' Feeds lines already
            carry the paths above -> ALREADY DONE. Stop.
          · Anything else -> proceed.

        PRECEDENCE: the dispatch_context enumeration of SIX papers governs. The precheck gates
        whether to act; it never widens the set. Do not repoint a seventh paper because it looks
        similar — the other nine are deferred to C-116 by this review.
    - item: ls-files-half-reaches-one-prompt
      kind: redispatch
      note: Move the complete ignore check into the shared fragment and have plan_feature consume it.
      dispatch_tool: build_minor.sh
      dispatch_context: |
        Add the `git ls-files` half to the SHARED fragment
        scripts/workflows/temporal/modules/assistant/prompts/gitignore_collision_check.md,
        so build_draft, build_draft_minor and plan_revision all receive it.

        Then remove plan_feature's divergent inline copy: replace the two paragraphs in
        scripts/workflows/temporal/modules/assistant/plan/plan_feature/prompts/plan_feature.md
        (the `.gitignore-collision check` paragraph and the `git ls-files` paragraph this PR added)
        with the placeholder `${GITIGNORE_COLLISION_CHECK}`, and add
        "GITIGNORE_COLLISION_CHECK": act.shared_prompt("gitignore_collision_check")
        to the render map in plan_feature/plan_feature_workflow.py (alongside the existing
        shared_prompt entries around line 195).

        Then RE-MEASURE plan_feature.md and update its BUDGETS entry in
        testing/scripts/tests/unit/test_prompt_budgets.py — it will SHRINK from 24_482. Record the
        new number as measured, and say in the comment that the shrink is the promotion seam, not
        a content deletion.

        The fragment is 555 bytes against an 8,000-byte FLOOR, so it still needs no BUDGETS entry.
        _FRAGMENT_FLOOR["gitignore_collision_check"] = 1 is a MINIMUM, so growth passes it.

        DO NOT TOUCH: scripts/workflows/*.sh or scripts/workflows/children/*.sh — the bash fleet is
        frozen reference. Do not change the other three consumers' prompts; they pick this up
        through the placeholder they already carry.
      precheck: |
        Run on the PR branch `feeds-liveness`, from the repo root. This tests the PROPERTY
        (does the instruction reach every consumer of the shared block?), not the diff that
        surfaced it:

          python - <<'PY'
          from pathlib import Path
          A = Path("scripts/workflows/temporal/modules/assistant")
          frag = (A/"prompts/gitignore_collision_check.md").read_text()
          pf   = (A/"plan/plan_feature/prompts/plan_feature.md").read_text()
          print("FRAGMENT_INSTRUCTS_LS_FILES", "ls-files" in frag)
          print("PLAN_FEATURE_STILL_INLINE", "${GITIGNORE_COLLISION_CHECK}" not in pf)
          PY

        STOP PREDICATE, split:
          · FRAGMENT_INSTRUCTS_LS_FILES True and PLAN_FEATURE_STILL_INLINE False -> ALREADY DONE. Stop.
          · Any other combination -> proceed. In particular, "the shared fragment is short" or
            "only one consumer needs it" is NOT a reason to stop — the operational-safety family
            ruling already decided this block does not fork per consumer.

        PRECEDENCE: the dispatch_context governs which files change. The precheck decides only
        whether to act.
    - item: ratchet-cannot-tell-repair-from-deletion
      kind: redispatch
      note: Fail loud when a baselined paper's Feeds line disappears, instead of reporting it repaired.
      dispatch_tool: build_minor.sh
      dispatch_context: |
        In testing/scripts/tests/unit/test_a_research_papers_FEEDS_target_resolves.py, make
        test_a_REPOINTED_target_leaves_the_baseline distinguish its two states before asserting:

          · the target now RESOLVES  -> current message is correct; keep it.
          · the paper no longer declares that `Feeds:` target at all (line deleted, or the paper
            deleted) -> a DIFFERENT failure with its own message, saying the pointer was REMOVED
            rather than repaired, and that removing the field is not a way to satisfy this guard.

        Implement by reading each baselined paper's header block and checking whether the baselined
        target token is still present in its `Feeds:` line, rather than inferring from _dead() alone.
        A paper legitimately retired (like cross_node_memory_protocol.md, which carries a SUPERSEDED
        banner and no `Feeds:` line) must be able to pass — treat an explicit SUPERSEDED/retired
        marker as an accepted removal and say so in the message.

        Also drop the unused `tmp_path` parameter from test_the_detector_can_actually_FAIL.

        DO NOT TOUCH: the ACCEPTED baseline contents (item 1 owns those), or _dead()'s resolution
        logic.
      precheck: |
        Run on the PR branch `feeds-liveness`, from the repo root. This DRIVES the defect rather
        than reading the code that contains it:

          P=docs/standards/architecture/research/raw/hermes_assessment.md
          cp "$P" /tmp/feeds-precheck.bak
          sed -i '/^Feeds:/d' "$P"
          python -m pytest testing/scripts/tests/unit/test_a_research_papers_FEEDS_target_resolves.py \
            -q 2>&1 | grep -c 'the pointer was repaired or the target came back'
          cp /tmp/feeds-precheck.bak "$P"; git status --short

        STOP PREDICATE, split:
          · The "pointer was repaired" string is printed (count >= 1) -> the defect is live, proceed.
          · Count is 0 AND the run still FAILS with a message naming a REMOVED pointer
            -> ALREADY DONE. Stop.
          · Count is 0 AND the run PASSES -> do NOT stop; that is a worse state than the finding
            described (deletion is now silent). Proceed and report it.

        The restore line is mandatory — verify `git status --short` is empty before continuing.
    - item: anchor-liveness-proposal-not-placed
      kind: redispatch
      note: Place the pointer-liveness-beyond-paths proposal as a candidates.md row with a named component.
      dispatch_tool: build_minor.sh
      dispatch_context: |
        Append ONE new row to docs/standards/architecture/research/candidates.md for the proposal
        this PR's guard docstring surfaces and does not place: pointer liveness is checked on FILE
        PATHS only, while `§` section anchors in `Feeds:` lines and `.md` pointers inside
        candidates.md Notes carry the same rot class with no watcher.

        Carry this evidence in the row: of four `Feeds:` anchors checked on 2026-08-20,
        TWO were already dead — backbone_edge_generality.md cites
        `§ "Where this repo sits"` and subscription_economics.md cites
        `§ "Affordability is not a footnote — it is the enabler"`, while
        docs/standards/architecture/problem-statement.md now heads those sections
        `## Where this sits — read this first` and `## Affordability is the enabler`.
        So repointing a path can leave a paper resolving to the file and not to the section it cites.

        Fill the `component` cell — do not leave it blank, or plan-candidates scaffolds nothing.
        State in the Note that this is DISTINCT from C-116 (which asks what STATE a paper is in when
        its destination dissolved; this asks what is CHECKED) and distinct from the PR #127 guard
        (which checks path resolution only).

        Follow the file's existing column order and id sequence exactly; take the next free C-number.
        DO NOT set the `decision` or `status` columns — leave both empty for triage.
        DO NOT extend the guard itself in this PR; placing the proposal is the whole task.
      precheck: |
        Run on the PR branch `feeds-liveness`, from the repo root:

          grep -c 'Where this repo sits' docs/standards/architecture/problem-statement.md
          grep -c 'Where this sits — read this first' docs/standards/architecture/problem-statement.md

        Expected: 0 then 1 — the cited anchor is absent and the live heading is present, which is
        the evidence the row asserts. This is a different check from the docstring reading that
        surfaced the proposal.

        STOP PREDICATE, split:
          · First count is 1 (the cited anchor IS live) -> the evidence sentence is wrong; file the
            row WITHOUT that example rather than stopping. The proposal stands on the docstring's
            own statement that these surfaces have no watcher.
          · candidates.md already carries a row for anchor/Note pointer liveness -> ALREADY DONE.
            Stop, and do not file a duplicate.
          · Otherwise -> proceed.

        PRECEDENCE: ONE row, for the pointer-liveness-beyond-paths proposal. The precheck does not
        license filing a second row for anything else noticed nearby.
    - item: no-attestation-comment
      kind: redispatch
      note: Post the Decision Log, Deferred Work and Post-Run Reflection this PR never got.
      dispatch_tool: build_minor.sh
      dispatch_context: |
        Post the missing attestation as PR comments on #127, in the same shape every sibling PR
        uses (#132, #133, #126, #124 each carry all three sections):

          · Decision Log — the real choices: freezing rather than clean-failing the 15; making the
            ratchet two-way; scoping anchors and candidates.md Notes out; putting the `git ls-files`
            half in a prompt at all.
          · Deferred Work — with VERIFIED pointers. The nine dissolved-destination entries point at
            candidates.md C-116. Do NOT re-defer the six knowable repairs or the shared-fragment
            promotion; both are being fixed in this same correction pass.
          · Post-Run Reflection — friction and tooling-level suggestions.

        Write it for the work AS IT STANDS AFTER this correction pass, not as the original run would
        have written it, and say plainly that it was authored by a correction pass rather than by the
        run that made the choices — reconstruction presented as attestation is the failure mode here.

        No line in any comment body may START with `@claude`.
      precheck: |
        Run from anywhere with gh authenticated against helloskyy-io/Claude-Dot-Files:

          gh api repos/helloskyy-io/Claude-Dot-Files/issues/127/comments \
            --jq '[.[].body | scan("^## (Decision Log|Deferred Work|Post-Run Reflection)")] | unique'

        This checks the PROPERTY (the three attesting sections exist on the thread), not the comment
        count — a review comment or an unrelated note must not satisfy it.

        STOP PREDICATE, split:
          · All three section headings present -> ALREADY DONE. Stop.
          · Some present, some missing -> proceed, and add ONLY the missing ones.
          · None present -> proceed with all three.
  laundered_deferrals:
    caught: 0
    of_total: 1
  homeless_items: 0
  redispatched: false
```

---

## Post-Run Reflection

**Friction.**

- **`attempt:`'s pass-1 seeding rule breaks on exactly the PR that most needs it.** The rule seeds from the count of attesting reflection comments; this PR has zero, so the seed is `0` even though a build attempt plainly landed a commit. The seed is least accurate precisely when the attestation gap it is measuring is total. I recorded `0` per the literal rule and flagged the divergence in-block, but a later pass reading `attempt: 0` alongside one landed commit will have to re-derive this. Worth considering: seed from attesting comments, floored at 1 when the PR has commits.

- **The `Feeds:` glob reads the working tree, not `git ls-files`** — which is the opposite of the lesson the same PR teaches in `plan_feature.md`. I checked and it is *correct here* (a guard over a working tree gives a true local answer for this question, and the false-green direction the PR describes does not apply), so it is not a finding. Noting it because the next reader will very likely trip on the apparent contradiction and re-derive the same conclusion.

**Tooling-level suggestions (claude-dot-files).**

- **The "one dispatch, five next_steps" shape has no expression in the schema.** All five runway entries here name `build_minor.sh` and are meant to be fired **once**, but the schema gives one `dispatch_tool` and one `dispatch_context` per entry, so nothing distinguishes "five entries, one dispatch" from "five entries, five dispatches". I stated it in prose above the yaml, which a machine consumer will not read. A `dispatch_group:` key — entries sharing a value ride one invocation — would make it checkable.

- **The disposition enum cannot express "correctly reasoned when written, invalidated by the branch's own rebase".** `ls-files-half-reaches-one-prompt` is not a producing-run failure in any ordinary sense: the run's reasoning was sound at authoring time and its own rebase reversed the premise. MIS-DISPOSITIONED is the closest available classification and it still reads as a charge against the run. This is the *reverse* of the case that motivated MIS-DISPOSITIONED — there, honesty was being scored as dishonesty; here, a mechanical consequence of rebasing is. A `premise-reversed` classification, counted against neither the run nor the corpus, would describe it accurately.

Pumapumapumas:
## Decision Log

**Rigour tier: safety control — 6 mutations.** This PR ships a CI gate whose whole job is to make a class of rot visible, and pass 1's third finding was that one of its two assertions told a *false* story. A gate that cannot go red, or that goes red with the wrong message, is worse than no gate — so every arm was driven, not reasoned. The six: (1) delete `hermes_assessment.md`'s `Feeds:` line → must report *deleted*, not *repaired*; (2) repoint it at a real file → must report *repaired*; (3) break a repointed target → must report *new rot*; (4) drop `${GITIGNORE_COLLISION_CHECK}` from `plan_feature.md` while still supplying it → must fail the render check; (5) cut the fragment back to one paragraph → must fail the floor; (6) `test_the_ratchet_DISCRIMINATES` drives all four classifier arms on synthetic data permanently. All six fired; tree restored and re-verified green after each.

- **[Low]** `component: pointer-liveness` is a **new** component name for `C-gkuvhctk`, not an existing one. Alternatives: `continuous-process-improvement`, whose own doc scopes it to *"mining run logs and reflections"* — this is documentation-surface integrity, not run-artifact mining; `safety-and-guardrails`, which is Claude Code hooks and permissions and is `✅ COMPLETE`; and leaving it **blank**, which the file's own rules explicitly permit as *"an unanswered question, not an error"*. Why: the task said *named* `component`, and a name commits `plan-candidates` to scaffolding `docs/development/pointer-liveness/`. **This is the decision on this run most worth a second opinion** — if the operator reads pointer-liveness as too small to be its own component, the cell should go blank rather than be re-pointed at a component that does not want it.

- **[Low]** Anchors for the two MMF papers were resolved to **phase headings in `roadmap.md`**, using that file's own § *Where the research contradicted the sprint milestones* table, which quotes the exact sprint milestone names those papers cite (*"Design it … with a total fail-safe default"*) and says which phase absorbed each. Alternatives: point at the phase docs directly (`phase3_typed_exit_record.md`), or leave the milestone names bare. Why: the baseline's own note says the destination is `roadmap.md`, and the roadmap is where the mapping is written down — pointing past it at a phase doc would be my inference rather than the tree's. **A reader who disagrees should say so**: this is the one anchor set I derived rather than found stated.

- **[Medium]** `origin/main` was **merged in twice**, and that is not one of the five findings. Alternatives: rebase (rewrites pushed history — the safety rule requires explicit approval and this task gave none); or build against the stale tree. Why: the task's own central instruction — *"read the file's current headings yourself and match against those… as they are on disk right now"* — was **unbuildable as written**. See Friction. A merge is non-destructive, needs no force-push, and made the instruction true. Cost: two merge commits in the history, and the PR's diff now includes a `candidates.md` conflict resolution.

- **[Medium]** Continuation-line blindness in `_FEEDS` was **surfaced, not fixed**, and folded into `C-gkuvhctk` rather than filed separately. Alternatives: fix it (it is ~2 lines); file it as its own row. Why: fixing it would surface at least two more dead targets, which would **grow** the frozen baseline — forbidden by its own header (*"THIS LIST MAY SHRINK. IT MAY NEVER GROW"*) — and turn a five-finding correction pass into a baseline re-ruling. It shares one remedy with anchor rot and Note rot (*a pointer that resolved once and does not now*), so under the cluster rule it is one item, not three. The task's explicit warning about a sibling PR that grew a one-line fix into a 604-line guard applied directly.

- **[Medium]** `backbone_edge_generality.md` keeps its pointer to `system-overview.md`, **a file that has never existed anywhere in this tree**, even though I rewrote the block it sits in. Alternatives: delete the sentence; guess where that content went. Why: deleting a pointer is precisely the laundering that finding 3 exists to stop — I would have been committing the defect I was fixing in the same commit. Guessing is the *"invisibly wrong pointer"* the freeze exists to avoid, and it is the nine-entry class the task reserved for an operator. It is recorded as evidence on `C-gkuvhctk` instead.

- **[Low]** Anchor text was matched against **`origin/main`**, which at merge time became "on disk". The review's own suggested replacement (`Where this sits — read this first`) was already stale when it was written; the task flagged this and was right. Two further headings changed that the task did not list — `Jarvis — the assistant edge (this repo)` → `The assistant edge (this repo)` and `Building & industrial automation — the next edge` → `Future edges` — confirming its *"do not assume this list is complete"*. Neither is referenced by the six papers.

## Deferred Work

- **The nine remaining frozen `ACCEPTED` entries** (8 × `docs/development/roadmap.md`, 1 × a bare ambiguous `synthesis.md`) — Why deferred: each needs somebody who knows where that evidence was meant to land; the destinations dissolved rather than moved, so there is no correct target to derive. Explicitly out of scope per the task (*"Leave the other nine frozen"*). Tracked at: `docs/standards/architecture/research/candidates.md` row `C-9yi1yv2h`, `status: open`, `decision` blank. Verified by: `git show origin/main:docs/standards/architecture/research/candidates.md | grep C-9yi1yv2h` on the **default branch** → present, `status` cell is `` `open` ``, and the row reads *"A research paper whose `Feeds:` milestone was DELETED has no defined state, so it stays `current` while pointing at nothing"* and closes by distinguishing itself from *"the `Feeds:`-liveness guard in PR #127, which checks that a target RESOLVES and says nothing about what to do when it legitimately cannot"* — i.e. it covers exactly this item and says so by name.

- **Pointer liveness below the file level** — anchors, wrapped `Feeds:` continuation lines, and `candidates.md` Notes. Why deferred: capability that does not exist, so a PROPOSAL rather than a defect; and building it here would grow the frozen baseline and blow the stated scope. Tracked at: `docs/standards/architecture/research/candidates.md` row `C-gkuvhctk`, `component: pointer-liveness`, `status: open`, `decision` blank. Verified by: `grep -n 'C-gkuvhctk' docs/standards/architecture/research/candidates.md` **on this branch** → line 284, and `test_candidates_prose_matches_the_table.py` passes with the derived untriaged count at seventy-one. **This is a placement in this PR's own diff, so the pointer only becomes true at merge** — `finding-routing.md` §4 requires the producing run to place its own proposal, which by construction is not on the default branch yet. If this PR is closed unmerged the row goes with it.

## Post-Run Reflection

- **Friction — the task's central instruction was unbuildable as written, and its own warning is what caught it.** The brief says *"Read the file's current headings yourself and match against those… as they are on disk right now"*, in bold, as the guard against following the stale review. But the worktree was **10 commits behind `origin/main`**, and on disk `problem-statement.md` still read `Where this sits — read this first` and `The nearest neighbor` — i.e. **exactly the stale text the brief told me not to trust**. Following "on disk right now" literally would have shipped the defect the warning exists to prevent. The same reversal hit the candidate ids: the brief says ids are random base36 *"minted by `research_activities.candidate_ceiling` and handed to you"*, but **no batch was handed over**, and in the worktree that function still computed `max+1` over `C-NNN`. The migration was on `main` (`2af48df`), not in the branch. The general shape: **a brief written from `main`'s vantage describing a worktree that predates it.** The stage-1 instruction to check `origin/main` and not only my tree is what made both catchable — it is doing real work and should stay.

- **Friction — `candidates.md`'s derived untriaged count is a serialisation point that no id scheme can fix.** The id migration to random base36 removed the *id* race, but § *Where things stand* still carries a hand-written total that `test_candidates_prose_matches_the_table` derives, so **every concurrent PR that appends a row conflicts on that one sentence**. It bit twice in this run: 68 → 69, then a mid-run merge of #136 and #123 moved main to 70 and I resolved to 71. The conflict is trivially resolvable and correctly caught — but two branches can also merge cleanly and both leave the sentence wrong, which is the same shape the id race had.

- **Tooling-level suggestions (claude-dot-files):**
  - **Have the dispatcher state the branch's position against `origin/main`, or refuse to dispatch a `--pr` correction pass onto a stale worktree.** Both false premises here were the same defect — the brief described `main` and the worktree was 10 commits behind it — and the PR was already `CONFLICTING` before I touched it. A single line in the task context (*"HEAD is N commits behind origin/main; last common commit <sha>"*) would have turned two acts of detective work into a read.
  - **Say what a correction pass may do about a stale base.** The safety rule makes rebase conditional on explicit approval, this task gave none, and the finding-level work genuinely could not be built against the stale tree. I chose a merge because it needs no force-push, but I chose it unaided. `build_draft_minor`'s prompt should name merge-from-base as the sanctioned move (or forbid it) rather than leaving each run to re-derive it.
  - **`review-pr` should timestamp or SHA-pin the anchors it quotes.** Pass 1's remedy text quoted `Where this sits — read this first` as the *fix*; the heading had already been reworded. The review was written against a tree that was correct at the time and stale by the time it was actioned — a `(as of <sha>)` on quoted target text would let the next pass tell "the review is wrong" from "the review is old".
  - **The count sentence in `candidates.md` § *Where things stand* is a hand-maintained restatement of something a test already derives.** The file's own history section says every number in it *"was wrong"* at least once. Rendering it (or letting the test rewrite it) would remove the last per-append serialisation point in a file explicitly re-engineered for concurrent appends.

Pumapumapumas:
## Decision Log

**Rigour tier: safety control — 2 negative controls, plus 5 discrimination runs against the inherited gate.** This PR ships a CI gate, and the tier table asks for discrimination analysis on a guard rather than a bare red-step. I did not read the draft pass's account of its own six mutations — I re-ran the arms myself, because a harness reports a verdict and a wrong verdict is invisible in a diff. What that found is the first item below: the gate had a live blind spot that three passes had read past.

- **[High]** **The 20-line header window was fixed, and this is the decision I am least sure of, because widening a frozen-baseline gate can only ever make it stricter.** `_declared()` read `splitlines()[:20]`; `cross_node_memory_protocol.md` carries a 34-line SUPERSEDED banner and its header `Feeds:` sits at **line 45**, so the guard recorded it as *"declares no destination"* — **the identical value a DELETED pointer produces**, which is the exact state the ratchet split (finding 3) was built to treat as evidence-destruction. Alternatives: leave it and file a row (the reviewer graded it latent); comment the magic number without changing behaviour; parse the fenced metadata block instead of the whole file. Why fixed: it is not latent, it is live — the paper was invisible to `test_no_NEW_dead_feeds_target` today, and only luck (its target resolves) kept that from being a false PASS. **The risk that stopped me from being casual about it is baseline growth**, which the baseline's own header forbids, so I measured before changing: nine dead targets with the window, nine without. Zero delta, so the frozen list cannot grow on this and the change is strictly a coverage gain.

- **[Medium]** **`backbone_edge_generality.md` keeps its dead pointer to `system-overview.md`** — a file that has never existed anywhere in this tree — and I am upholding the draft pass's call rather than reversing it. Alternatives: delete the sentence; guess the target; annotate it in the paper as known-dead, which the guard's own failure message actually prescribes (*"say so in the paper rather than leaving a pointer to nothing"*). Why upheld: it is the **same class as the nine frozen entries** the task explicitly reserved for an operator — a destination that dissolved, where the run cannot know what was meant — and it is already recorded as evidence on `C-gkuvhctk`. The third option is the tempting one and I rejected it because annotating a pointer as dead in the paper is a content edit to research evidence, which is a different authority from repointing a target the baseline already named.

- **[Medium]** **The basename collision got an assertion rather than a re-key.** Every verdict in the module, `ACCEPTED` included, is keyed by `p.name` across five research pools; no collision exists today. Alternatives: re-key everything on the path relative to `ROOT`, which is the structurally correct fix and rewrites all nine frozen baseline keys. Why the assertion: the frozen baseline is the artifact this PR spent three findings stabilising, and rewriting its keys to close a defect that has never fired trades a live risk for a latent one. The assertion makes the invariant explicit and fails loud on the day it breaks, which is what the re-key would buy, without touching the baseline.

- **[Low]** **The `C-gkuvhctk` Note was corrected in place rather than left for triage.** The reviewer read *"four were live"* and computed *three* from the same sentence — the prose was parseable two ways, and it is the evidence a triage pass will rule on. It also described `dedicated_edge_routing.md`'s continuation-line target as `docs/development/roadmap.md` when the paper names a bare `roadmap.md`. Alternatives: leave both, since the row is untriaged anyway. Why corrected: this PR filed the row, so it is this PR's own artifact and not another filer's — the column-ownership rule that forbids amending someone else's row does not reach it. `decision`, `status` and the derived count are untouched.

- **[Low]** **`origin/main` was merged in a third time, at push.** Two commits landed while this pass ran, one of them editing `candidates.md` (row `C-yi65bngf` → `closed`). Alternatives: push without merging and let the PR go stale or conflicting. Why: the untriaged count is a hand-written sentence a test derives, so a `candidates.md` change on main is exactly the shape that silently invalidates it. Merged and re-derived: still seventy-one, because that row's `decision` stays blank.

## Deferred Work

Nothing new was deferred by this pass. The two items carried by the correction pass stand, and I re-verified both rather than inheriting the attestations:

- **The nine remaining frozen `ACCEPTED` entries** — Why deferred: explicitly out of scope per the task (*"Leave the other nine frozen"*); each needs an operator who knows where that evidence was meant to land. Tracked at: `docs/standards/architecture/research/candidates.md` row `C-9yi1yv2h`. Verified by: `git show origin/main:docs/standards/architecture/research/candidates.md | grep -c C-9yi1yv2h` → `1` on the **default branch**, `status` cell reads `open`, and the row's Note distinguishes itself by name from *"the `Feeds:`-liveness guard in PR #127, which checks that a target RESOLVES and says nothing about what to do when it legitimately cannot"*.

- **Pointer liveness below the file level** — anchors, wrapped `Feeds:` continuation lines, `candidates.md` Notes. Why deferred: a PROPOSAL, not a defect — capability that does not exist — and building it here would grow the frozen baseline. Tracked at: `candidates.md` row `C-gkuvhctk`, `component: pointer-liveness`, `status: open`, `decision` blank. Verified by: `grep -n 'C-gkuvhctk' docs/standards/architecture/research/candidates.md` **on this branch** → line 284, and `test_candidates_prose_matches_the_table.py` passes with the derived untriaged count at seventy-one after the merge of main. **This is a placement in this PR's own diff, so the pointer only becomes true at merge** — `finding-routing.md` §4 requires the producing run to place its own proposal, which by construction is not yet on the default branch. If this PR is closed unmerged, the row goes with it.

## Post-Run Reflection

- **Friction — the mutation instruction points almost entirely at inherited work, and the live defect this pass found was in inherited work that had already been mutated six times and reviewed twice.** The draft pass ran six mutations, all sound, and its account of them was accurate. What none of them asked was *"is there a paper this gate never sees at all?"* — mutation proves the arms you know about discriminate, and says nothing about the population being fed to them. The `[:20]` window is not visible from any arm; it is visible only from asking what `_declared()` returns for **every** paper and comparing that against the tree. A line in the stage-1 block along the lines of *"for a gate, also check the POPULATION it reads, not only the verdicts it returns"* would have made this a first-look check rather than a lucky one.

- **Friction — `review-pr`'s finding text quoted anchors that were stale when it was written, and the brief's correction was itself incomplete.** The brief warned about this and listed known changes, and the list was short by two headings — it says *"do not assume this list is complete"*, which is what saved it. The general fix is upstream of both documents: **a review that quotes target text should SHA-pin it.** The draft pass's reflection asked for the same thing independently, which is two runs on one PR reaching for one instrument.

- **Project-level suggestions (this repo):** the research pool has no rule about where a header block may start. One paper's 34-line SUPERSEDED banner sits above its metadata fence, which is what pushed `Feeds:` to line 45 and out of the guard's view. The guard is now banner-proof, but the underlying question — is the metadata fence the first thing in a paper, or may prose precede it? — is unstated, and the next tool that parses these headers will make the same assumption from scratch.

- **Tooling-level suggestions (claude-dot-files):**
  - **A gate-shaped PR should get a coverage check, not only a mutation check.** Same evidence as the friction item above: six correct mutations and two review passes did not surface a paper the gate was skipping, because every instrument in the workflow's vocabulary asks *does the check fire when it should* and none asks *what is the check reading*.
  - **The `build-refine-minor` prompt tells the run to defer to an existing home before filing, but gives no verb for "this defect is in an artifact my own PR filed".** The column-ownership rules say do not edit another filer's row; that reads, on a first pass, as covering rows this PR itself placed. It took a paragraph of reasoning to conclude the row was mine to correct. One clause — *a row your own PR filed is your artifact* — removes that.

Pumapumapumas:
## `review-pr` — disposition pass 2 (PR #127)

**Verdict rationale: MERGE — converged.** All five pass-1 HOLD findings are genuinely fixed, and I verified each against the artifact rather than against the run's account. The sixth (the deferral) re-verifies on the **default branch**. My own sweeps — run from four angles pass 1 did not use — produced nothing that passes the consequence gate. Per the convergence rule, no remaining item would have blocked this PR on pass 1's own bar.

**Mergeability read first:** `MERGEABLE` / `CLEAN`, all four checks green (`suite`, `CodeQL`, `Analyze (actions)`, `Analyze (python)`). `closingIssuesReferences` is empty and the body claims no closures — the field agrees. `git merge-base --is-ancestor origin/main HEAD` passes, so nothing here is judged against a rule that landed after the cut.

**Pass count:** exactly **one** fence-anchored `pr_review:` block on the thread → this is pass 2, matching the dispatch label. No divergence.

---

### What I measured myself, rather than read

The producing run's claims are unusually load-bearing on this PR (a frozen baseline, a rot rate, an anchor set), so I re-derived them:

| Claim | My independent measurement | |
|---|---|---|
| 9 dead `Feeds:` targets remain | Ran the resolution logic standalone: **9 dead**, and the set is **exactly** `ACCEPTED` — no drift, no growth, nothing laundered out | ✅ |
| Baseline shrank 15 → 9 | `ACCEPTED` holds 9 entries; six removed | ✅ |
| `plan_feature.md` re-measured to 23,437 | `wc -c` → **23437**, exact | ✅ |
| Suite green at 3214 passed / 3 skipped | `./testing/run-all.sh unit` → **3214 passed, 3 skipped**, exit 0 | ✅ |
| Fragment reaches every tier | **6 render sites** across `build_draft`, `build_draft_minor` (×2), `stages_1_to_4_from_plan`, `plan_feature`, `plan_revision` | ✅ |

**The anchor check is the one nobody had run.** Pass 1 checked four anchors against the *pre-repoint* text; the refine pass did not re-check them. A correct path with a dead anchor is precisely the "invisibly-wrong pointer" this PR argues against, so I resolved all ten against the live files:

- `§ "Where this sits: read this first"` → line 3 ✅ · `§ "Where we actually differ"` → line 58 ✅ · `§ "Affordability is the enabler"` → line 173 ✅ · `§ "What this means for anything built here"` → "Nothing may assume the coding edge", line 231 ✅
- Claim **#2** quoted as *"Work is bound to a machine by identity, and no scheduling label can move it"* — **exact match**, line 92 ✅ · Claim **#3** quoted as *"The first edge builds the others, then operates inside them"* — **exact match**, line 102 ✅
- MMF `§ "What this component is"` → line 11, and it genuinely holds the Kind 1 / Kind 2 relationship decision, which is what the paper cites it for ✅ · `§ "Phase 2 — Document Kind 1 as a framework"` → line 51 ✅ · `§ "Phase 3 — The typed exit record…"` → line 67 ✅ · `§ "Phase 4 — Migrate the fleet"` → line 82 ✅

**Ten of ten resolve.** Notably, **pass 1's own prescribed replacement was stale** — it quoted `Where this sits — read this first` (em dash) and the live heading is `Where this sits: read this first` (colon). The run read the file instead of following my prior pass, which is the correct behaviour and the right way round. Recording it under invariant 3: the defect was in my predecessor's prescription, not in the work.

---

### Sweeps run, and why each angle is new

Pass 1 ran asymmetric-presence, boundary, deleted-artifact, completion-checkbox and current-tree. I did not re-issue those searches. Four angles were unavailable to it and are live now:

1. **Merge-commit sweep — N/A for pass 1 (one non-merge commit), live now (three merges).** A conflict resolution is a deletion channel with no red lines. I read all three `--cc`. Both `candidates.md` conflicts and the `file_structure.txt` conflict resolve by keeping **both** sides; the derived untriaged count lands at seventy-one and `test_candidates_prose_matches_the_table` passes. **Nothing was discarded.**
2. **Regression check on the refine pass's own fix.** Widening `_declared()` from `splitlines()[:20]` to the whole file trades a blind spot for a false-positive risk: `.search()` takes the *first* match, so a `Feeds:` in body prose or a fenced example would now win. Checked: **zero** papers carry more than one line-start `Feeds:`, and **every** first match sits inside the header fence. The widening is clean.
3. **Asymmetric budget sweep on the promoted fragment.** The fragment grew 555 → 1590 B, and only `plan_feature`'s budget was adjusted — the obvious shape of a missed sibling. It is not one: `BUDGETS` is keyed on **file** size, and the other four consumers already carried the placeholder, so their files did not change. The fragment itself sits far under `FLOOR = 8_000`, and `test_every_prompt_over_the_floor_HAS_a_budget` only requires an entry above it. **No gap.**
4. **Map-gate set-equality after the merge conflict.** My first naive check reported 219 files "missing" from `docs/file_structure.txt`. That is my predicate being wrong, not the map: `test_planning_directories_are_ROLLED_UP_in_the_map` deliberately requires planning directories to be **summarised, never enumerated**. Reporting it would have been a false absence claim, so I am recording the check and its refutation rather than the finding.

---

### Disposition table

|---|---|---|---|
| `feeds-baseline-freezes-known-repairs` | correctness | **FIXED** (verified) | Six repointed, baseline 15 → 9. I re-derived the dead set independently: **9, exactly matching `ACCEPTED`** — the freeze did not absorb anything it should have repaired, and did not grow. Paths *and* anchors were repointed together, and all ten anchors resolve (above). The two quoted claim-wordings are byte-exact against the reworded source. |
| `ls-files-half-reaches-one-prompt` | correctness | **FIXED** (verified) | The `git ls-files` half is in the shared fragment; six render sites cover all four tiers. `plan_feature`'s divergent inline copy is deleted, not left beside it. `_FRAGMENT_FLOOR` 1 → 2 with the shrink-reason recorded. Budget re-measured **downward** to 23,437 — verified exact — so the freed 1,045 B is not silently re-granted. |
| `ratchet-cannot-tell-repair-from-deletion` | correctness | **FIXED** (verified) | `_ratchet()` now returns `(repaired, gone)` and keeps three states apart — live pointer, no `Feeds:` line, paper absent. It is **pure over injected inputs**, so `test_the_ratchet_DISCRIMINATES` drives all four arms on synthetic data this tree does not contain, plus a fifth assertion that a still-dead entry stays silent. `test_a_DELETED_pointer_is_NOT_a_repaired_one` carries its own distinct message. The gate is no longer passable by deleting the field it protects. |
| `anchor-liveness-proposal-not-placed` | scope | **FIXED** (verified) | `C-gkuvhctk` at `candidates.md:284`, `status: open`, **`component: pointer-liveness` is populated** (I checked the cell per the blank-component rule). The Note carries the measurement, the consequence, an explicit done-state, and distinguishes itself by name from `C-9yi1yv2h`. Its arithmetic is now internally consistent — six dead + four resolved = the ten it claims. This is the prescribed placement for a proposal: a committed row in the diff, not a docstring. |
| `no-attestation-comment` | standards-implication | **FIXED** (verified) | Two attesting comments now on the thread, each carrying **Decision Log + Deferred Work + Post-Run Reflection**. Coverage is per-commit and holds: the correction pass attests its three commits, the refine pass attests its one. Both state plainly that they were authored by correction passes rather than by the original run — the reconstruction-as-attestation failure mode is avoided explicitly. |
| `feeds-baseline-dissolved-destinations` | deferral | **DEFERRED** (pointer re-verified) | The nine genuinely need an operator who knows where that evidence was meant to land. **Pointer id changed since pass 1** — `C-116` was migrated to random base36 as `C-9yi1yv2h` (`2af48df` on main). I did not inherit the run's attestation: `git show origin/main:…/candidates.md` → present, `status: open`, and its Note names PR #127's guard explicitly as the thing it is *distinct* from. **Verified on the default branch**, so it survives this PR being closed. |

**Laundering check.** Three deferrals have been raised across both passes; **zero laundered**. `C-gkuvhctk` lives in this PR's own diff and the run said so in the same breath — under the proposal-placement rule that is the *prescribed* surface for a proposal, not a burial, and the disclosure is the opposite of laundering. `C-9yi1yv2h` resolves on `main`. `homeless_items: 0`.

**Invariant 1 (absence is non-terminal):** nothing vanished between passes. All six pass-1 ids are carried forward and terminally dispositioned above.

---

### Recommendation, carried without holding the PR

One item surfaced and **did not pass the consequence gate**, so it is recorded here rather than in the runway — visible without costing a ruling:

- **The guard's headline docstring says `36 papers carry a header Feeds:`; the live population is 37.** `36` is the pre-fix blind-spot count — the number the `[:20]` window saw before the refine pass widened it. I confirmed no paper arrived by merge (37 at `2a69b2a`, `931a9a0`, `cac3677` and `origin/main` alike). Nothing breaks: no guard reads the number, no decision turns on it, and the `_declared()` docstring forty lines below states the recovery explicitly, so a reader of the module gets the correct picture. If the file is touched again, `36 → 37` (and `42% → 41%`) is the tidy-up.

---

## Post-Run Reflection

**Friction.**

- **The convergence rule and gate 0 pull in opposite directions on a preventive item in the PR's own artifact, and nothing says which wins.** Gate 0 makes `noted` illegal for an artifact this PR created — legal values are `fixed`, `rejected`, `hold`. The docstring count above is none of those: it is not fixed, it is genuinely wrong so it cannot be rejected, and `hold` would contradict the convergence rule's explicit *"do not HOLD a PR whose remaining findings would never have blocked it on pass 1."* I resolved it upstream, at the consequence gate — it is not a finding, so no disposition is owed — which I believe is right and is also the only path the schema leaves open. Worth stating because the next pass will hit the same corner and re-derive the same reasoning.

- **Pass 1's `attempt: 0` seed propagates a known-wrong number forward.** Pass 1 recorded `0` per the literal rule and flagged in-block that it under-reported by one. Carrying forward and incrementing per landed dispatch gives `2`; the true count is **3** (original build + correction + refine). I have recorded `2` per the stated rule and am naming the divergence here rather than silently correcting it, since a durable field should not be quietly re-based. Pass 1's own suggestion — *seed from attesting comments, floored at 1 when the PR has commits* — would have produced the right answer.

**Tooling-level suggestions (claude-dot-files).**

- **A "verify the prior pass's prescriptions" step should be explicit in stage 1, not only in the invariants.** Invariant 3 asks whether my prior pass's fixes landed and regressed anything; it does not ask whether the prior pass's *prescribed text* was itself correct. Here it was not — pass 1 quoted an anchor that was already stale, and the run was right to override it. The catch came from reading the artifact, but a run under turn pressure would plausibly have "verified" the fix by checking it matched the prescription, which would have shipped the stale anchor. One clause — *check the prescription against the artifact, not the fix against the prescription* — closes it.

- **Two independent passes on this PR have now asked for SHA-pinned quotes in review findings** (pass 1's own reflection, and both producing-run reflections). That is three requests across one PR for one instrument. A `(as of <sha>)` on any target text a review quotes would let the next pass tell *"the review is wrong"* from *"the review is old"* — which is exactly the distinction that cost this PR a false premise.

- **The map-gate predicate trap is worth a line in the absence-claim rule.** My naive `git ls-files` set-equality check reported 219 false absences, because this repo deliberately rolls up planning directories. The existing guidance says verify absence claims twice with two different checks; what actually saved it was reading *the repo's own gate* to learn the intended predicate. "When the artifact is an enumeration, read the guard that enforces it before deriving your own predicate" is the sharper form.

```yaml
pr_review:
  pr: 127
  run_id: e39d35e913c4441898e68f9847708236
  pass: 2                            # verified: exactly ONE fence-anchored `pr_review:` block on
                                     # the thread (pass 1's). Matches the dispatch label; no divergence.
  attempt: 2                         # carried from pass 1's `0`, +1 for the build-draft-minor
                                     # correction pass and +1 for the build-refine-minor pass, both
                                     # of which landed work. FLAGGED: the true count is 3 — pass 1's
                                     # seed under-reported by one and said so in its own block. Not
                                     # silently re-based; see Post-Run Reflection.
  verdict: MERGE
  converged: true
  findings:
    - id: feeds-baseline-freezes-known-repairs
      title: Six papers ship with dead Feeds targets whose correct destination the baseline already records
      category: correctness
      consequence: >
        A planner following any of these six would arrive nowhere while the repair sat written down
        three lines away in ACCEPTED, and the baseline could not shrink until somebody re-derived
        what the run already knew.
      disposition: fixed
      remedy: fix-in-place
    - id: ls-files-half-reaches-one-prompt
      title: Three file-creating tiers still get the incomplete ignore check, so a new file passes locally and turns CI red on commit
      category: correctness
      consequence: >
        build_draft, build_draft_minor and plan_revision would have kept only the `git status` half,
        so each could go green on an untracked new file and fail the map gate once committed — the
        precise failure this PR documents its own author hitting.
      disposition: fixed
      remedy: fix-in-place
    - id: ratchet-cannot-tell-repair-from-deletion
      title: Deleting a failing Feeds line satisfies the ratchet and prints that the pointer was repaired
      category: correctness
      consequence: >
        The cheapest way to clear a failing entry would have been to delete the field the guard
        exists to protect, while the ratchet asserted it had been repaired — a gate passable by
        damaging the artifact it protects.
      disposition: fixed
      remedy: fix-in-place
    - id: anchor-liveness-proposal-not-placed
      title: The next axis of pointer rot is surfaced only in a docstring, so it dies at merge while already live
      category: scope
      consequence: >
        Six of the ten anchors those six papers carry were dead while the file-level guard read
        green on all ten; unplaced, that measurement would have been lost at merge.
      disposition: fixed
      remedy: fix-in-place
    - id: no-attestation-comment
      title: The PR carries no Decision Log, Deferred Work or Reflection, so its one deferral had no surface on which to be re-checked
      consequence: >
        The deferral parking the ls-files half lived only in body prose and was invalidated by this
        branch's own rebase with nothing to re-read it, which is how that finding went stale.
      disposition: fixed
      remedy: fix-in-place
    - id: feeds-baseline-dissolved-destinations
      title: Nine baselined papers point at destinations that dissolved, and no rule says what state such a paper is in
      category: deferral
      consequence: >
        Eight papers cite a `docs/development/roadmap.md` that never existed and one a bare ambiguous
        `synthesis.md`. No note names where they land, so each needs an operator who knows which
        component roadmap the evidence was for. Left unruled they read as current evidence for work
        that may no longer be planned.
      disposition: deferred
      pointer: "docs/standards/architecture/research/candidates.md — row C-9yi1yv2h, status `open` (pass 1 recorded this as C-116; the id was migrated to random base36 in 2af48df)"
      pointer_verified: true
      reviewed_sha: 3db1acec170c594bdfdeaa41e8a383fa9fadf7a5
  laundered_deferrals:
    caught: 0
    of_total: 3
  homeless_items: 0
  redispatched: false
```

PR #138, opened 2026-08-24
build-draft: PMP Phase 9 — one run, one identity, one bag

Pumapumapumas:
Builds [PMP Phase 9 — One run, one identity, one bag](docs/development/persistent-memory-protocol/phase9_one_run_one_identity.md).

Phase 1 built the folder a run writes into and keyed it by a run id. It never said **who names a run**, because there was only one shape an invocation could take. Two clocks ended that — the Temporal port makes an entrypoint a client, and Workflow Decomposition Phase 3 gives nine children runners of their own — and this phase settles both before either lands.

> **This PR has been through two correction passes** — `build-refine` (`50c68e2`) and a scoped `build-refine-minor` (`a133e5b`) answering `review-pr`'s HOLD. What each changed is in *Refine pass* and *Correction pass* below; the sections above them describe the PR as it now stands, re-verified against the tree rather than carried forward.

## Summary of what was built

**One naming authority (r1).** `mint_run_id` is the fleet's only run-naming function, with exactly one caller — `scripts/workflows/temporal/scripts/dispatch_identity.py`, the client side of a dispatch. The second, unrelated `run_id` (the per-model-invocation nonce minted inside `run_claude`) is **renamed `invocation_id`** across every Python identifier — 160 occurrences in 15 files, 8 of them outside `tests/`.

**The name arrives from outside (r2).** Zero of eleven entrypoints mint. Two new flags, declared once in `dispatch_identity.add_identity_arguments` and wired into all eleven parsers:

| invocation | meaning |
|---|---|
| `--run-id X` | this invocation **IS** run X (a parent, or a retry of one) |
| `--run-id X --writer w` | this invocation is a **member** of run X — adopts its bag, takes subfolder `w` |
| `--writer w` alone | **refused** — a writer with no run to join is the failure the flag exists to prevent |
| neither | this invocation is the run; the boundary mints and **prints** the name so a retry can supply it back |

**Sequential-retry idempotency, demonstrated (r3).** Opening a bag twice under one id yields one bag, `bag-info.txt` byte-identical, a first attempt's gap record intact.

**The child discriminator is passed, never inferred (r4).** Explicitly not from an env var: it is inherited by every descendant, so a parent exporting its own name hands it to a grandchild that should have had its own. **And routing through the boundary that enforces this is itself swept** — see the refine pass.

[...]

PR #139, opened 2026-08-25
plan-verify: size and judge persistent-memory-protocol

Pumapumapumas:
# plan-verify — Persistent Memory Protocol, sized and judged cold

Nine phase docs and a roadmap, read cold against the synthesis, the problem statement, the Architecture Standard, the Documentation Standard, the stack reference and `sprint.md`. Estimates land in `roadmap.md` and nowhere else.

## Sizing

| Phase | Delivers | Estimate | Basis | Gated on |
|---|---|---|---|---|
| **1** The journal root and the run bag | root resolution, the BagIt bag, its validator, bag-open as an activity with an enumerating sweep | **~0 hrs remaining — COMPLETE** | shipped; a forecast for finished work would describe nothing. Excluded from the total below | — |
| **9** One run, one identity, one bag | one naming authority, the id supplied by the caller, the standalone-child rule, a validated permitted set | **~10 hrs (RESIDUAL)** | built 2026-08-24; r1/r3/r4/r5/r6 met. What is left is r2's reconciliation and r7's mechanism. The spread is r2's own fork: a parameter change if the orchestrator's workflow id can BE the run id, a migration if a second name must be joined | the port naming a dispatch; `C-zhdm5gh1`'s carrier |
| **2** The content store | bytes by checksum, offline `verify` with three exit codes, `evidence_set_hash` | **~14 hrs** | one new package, two activities, unit+integration. Spread is r7(c): teeing an existing tool vs. a new fetcher carrying its own https/redirect/SSRF/size policy | none |
| **3** The emit rule | every write path emits verbatim; a failed journal write is never silent | **~45 hrs** | thirteen requirements; an inventory whose halves need different mechanisms; intent+completion per paired write; four failure cases against a real journal; six event fields; capture-time filter; tag-namespace ruling; run-log cut-over; then wiring every path | Phase 1, and Phase 9 r1+r4 only |
| **4** Rebuildability is a test | replay, diff, restore, gap accounting | **~28 hrs** | pure event→tree replay with a containment contract binding every target; versioned two-section snapshot; stated normalisation set; separately-contained restore; committed synthetic fixture; two CI arms, neither skippable | Phase 3 |
| **5** Snapshots, then retention | the 1 GB budget, whole-folder oldest-first deletion, the snapshot barrier | **~22 hrs** | five-step pass with a terminal state; carry-with-compaction preserving class and run id; concurrency demo against a live run; conditional reachability pass | Temporal server — **r1's recurrence only**; r2–r9 ungated |
| **6** CPI reads the journal | the sweep sources the journal; every producer gets a named consumer | **~18 hrs** (excludes the port) | storage interface, repoint, producer/consumer table. Expensive part is r2's dual run with every disagreement explained — bounded by how much they disagree, which nobody has measured | the `review-runs` port |
| **7** Cross-machine aggregation | bags ship to object storage, local-first | **~34 hrs** | async shipper with bounded observable backlog; post-transfer validation plus a demonstrated corruption; prefix-scoped credentials and origin derivation; bucket posture; a real bucket to test against | a second machine that produces runs |
| **8** The poller | a schedule reads a to-do bit and starts a child | **~16 hrs** | scheduled workflow, fixed cue→dispatch table, origin filter tested against a synthetic non-local origin, failure path with a named escalation surface. Cost concentrates in r3's identity derivation and the durable already-fired marker | Temporal schedules |

**Total: ~187 hrs** across the eight phases with work left. Sprint-1 set (9 residual, 2, 3, 4) is **~97 hrs**; Sprint-2 set (5, 6, 7, 8) is **~90 hrs**.

**Calibrated against** `sprint.md`'s only sized component — Temporal Integration at 16/22/16/20/28/26/31 — and against what this component has actually shipped: Phase 1 plus Phase 9 landed ~2,180 lines of module code and ~5,075 lines of test across `modules/journal/` and eleven test modules. That is the anchor for what "one PMP phase" costs here, and it sits at the top of the sprint's range rather than the middle.

**This total is an input to a decision I did not make.** I have not scheduled anything, and I did not touch `sprint.md`.

---

## Where is this plan WEAKEST?

**The emit rule's model-issued half, and it is weakest precisely where the component's own design test lives.**

Phase 3's write-path inventory has two halves. The fleet-code half is a call site you wrap, held by an enumerating sweep in the shape Phase 1 r11 established and proved. The model-issued half — the child itself running `gh pr comment` — has no call site, so its mechanism is a post-exit harvest with a stated failure mode of its own. **What guards the two is asymmetric, and the plan says so in its own words:**

> *"A build that enumerates only the first half populates the table, wires every row, closes requirement 1, and never emits a single PR comment — and nothing goes red."* — Phase 3 § *The write-path inventory*

The only thing standing behind the harvest is one checklist item: demonstrate one `research_minor` cycle whose PR body and comments land verbatim. **That is a one-time demonstration, not a check.** Phase 3 r12's activity boundary explicitly *"does not reach model-issued writes at all"*, and Phase 4 does not close it on Phase 4's own reading either — r3's negative test is scoped to the stores in the test set, r1 names those as `candidates.md` and `direction.md`, and § *Which stores are in the test set* says GitHub-hosted surfaces are *"the expected hard case"* that may land in r5's not-covered table instead.

So the likely end state is: the harvest half has no rebuild test **and** no standing test, while the two stores that DO have a rebuild test are the two whose highest-value content is written outside any run. The guarantee then decays exactly the way this component's thesis says an unenforced rule always does — on the PR body, the decision log and the reflection comment, which are the artifacts the operator's sentence is about.

**What would have to be true for it to hold:** a standing check for the harvest half, in the family of `test_every_parent_opens_a_run_bag` — either a sweep asserting every workflow that posts to a GitHub surface produces a harvested completion in its own bag, or a per-run reconciliation counting store-side comments against harvested completions and failing on a shortfall. **Placed as `C-6i1u3b1f`** rather than left in this body.

**Second-weakest, for contrast:** Phases 1–5 ship producers whose only planned consumer is Phase 6, which is gated on the `review-runs` port — and the roadmap itself establishes that the port sits below a milestone that *"gates everything below it"*, behind a whole sibling component with four open phases. The plan names this collision honestly and splits Phase 6 out of Phase 8 specifically to avoid it. It is still the case that the 262 MB failure being cited is being reproduced at larger scale with a merge-path gate on top. I did not make this the headline because the plan sees it; the harvest hole it does not see as a *guard* gap.

---

## One verifiable outcome per phase

| Phase | The single thing demonstrable when it is done |
|---|---|
| 1 | a folder a validator says is valid, and a red test when a parent does not open one |
| 9 | one run produces one bag under one name, across a parent, three children, a retry and a standalone child |
| 2 | `verify` resolves a real prior run's citations with the network off and distinguishes verified / missing / tampered |
| 3 | a full cycle's authored output appears in the journal verbatim, and each of four write-failure cases behaves as stated |
| 4 | replay reproduces two live stores, and deleting one emit turns the test red |
| 5 | a real over-budget journal reaches budget by deletion, refuses to cross the last snapshot, and still passes Phase 4's test |
| 6 | the journal-sourced sweep produces the incumbent's findings over one overlapping window |
| 7 | a run completes with the bucket unreachable and the backlog drains, with the bag validating on arrival |
| 8 | one real cue goes from `status: open` to a child that ran, and a repeated tick starts nothing new |

**Phase 3 is the one I would split, and the plan already contains the argument.** Its own sibling doc says adding a fourth concern to Phase 3 *"makes the phase a project"* — and it carries considerably more than four. Two verifiable outcomes are separable today:

1. **The emit and its failure contract** — the event, write-ahead ordering, intent+completion, the four failure cases, the gap event, the six admission fields. Demonstrable on the fleet-code half alone.
2. **The capture half** — the model-issued inventory, the post-exit harvest, capture-time secret filtering, and the run-log cut-over. Demonstrable independently, and it is the half that carries the weakness above.

That is a boundary judgement, not an edit. **The phase list is unchanged and I renamed nothing** — a phase number is identity.

Phase 7 is the second split, and it needs no argument from me: the synthesis sized it *"a couple of phases or its own sprint"*, and the phase doc already names the three seams (ship-and-validate / read / attribute) so that splitting it is mechanical.

---

## Producer / consumer

Walked in order. Every producer has a reader **somewhere**, and the exceptions are named rather than found:

| Producer | Read by | Note |
|---|---|---|
| bag, manifest, lifecycle flags (1) | Phase 3 emitters, Phase 4 replay, the operator validator | live today |
| content store (2) | `verify`, and Phase 7 r2 after transfer | fine |
| `evidence_set_hash` (2 r5) | **nothing** | admitted in-phase — *"computed, not consumed by anything yet"* — on the convergence-signal precedent. But unlike Phase 3 r7(c) it does **not** point at Phase 6 r3's table as a knowingly-empty cell with a trigger. Small asymmetry worth closing when Phase 2 runs |
| journal events (3) | Phase 4 replay | fine |
| credential epoch (3 r7c) | nothing, explicitly | named, with a trigger, and pre-routed to Phase 6 r3 as a knowingly-empty cell. This is the good shape |
| provenance class (3 r7d) | Phase 8 r7 | consumer is gated, and Phase 8 says it builds the filter anyway while it is a no-op. Correct |
| unwritable-journal signal (3 case d) | required to ship with a committed reader by r11 | the best-guarded producer in the plan |
| snapshot (4 r2) | Phase 5 | fine |
| retention/carried events (5 r8) | Phase 4 r7 counting, Phase 6 r6 output | Phase 6 gated |
| **`over-budget-unreclaimable` (5 r4)** | **nothing named** | the barrier-advance escalation reuses Phase 3 case (d)'s surface, which r11 gives a reader. This event does not get the same treatment — its only consumer is a non-zero exit code read by whatever schedules the pass. Given r11 exists one phase over, this reads as an omission rather than a decision |
| everything from 1–4 | **Phase 6** | the only general consumer, and it is gated. See second-weakest above |

---

## Evidence — every citation I followed, with what I checked against

- **`state_passing` §2.1 / §4.3.4** — the roadmap's whole taxonomy re-cut rests on these. Both check out exactly. §2.1: *"Nine rows. **Eight of them cross a `claude -p` process boundary**"*, and *"**Rows 1, 2, 3, 4 and 6 — five of the eight — are covered by neither kind.**"* The roadmap's five uncovered channels are the same five, and its mapping (rows 1–5 invocation state, row 6 measurement samples, rows 7–8 the working record) covers all eight without overlap. §4.3.4's *"lifecycle shape discriminates where audience does not"* is stated in the paper and is what the roadmap builds on. **Verified against the raw paper, not the synthesis.**
- **`bernstein` §4.6** — Phase 2's sizing citation. Reads *"`[Tier 1 · cost S–M]`"*, *"Cost: S–M. Derived from: S for the narrow version"*, *"**This is the item with the shortest path from 'read about it' to 'we are using it.'**"* Phase 2's claim that it *"named its roadmap home"* is true (*"the research workflow tooling (`research-critic`), and `Phase: Continuous Process Improvement`"*). Phase 2 relocating it into this component is a stated, reasoned divergence, not a miscitation.
- **`state_passing` §5.2** — Phase 7's gate. Reads *"Building a framework for a channel that carries one integer would be speculative generality"*. The section is about state-passing frameworks rather than cross-machine aggregation, so it is an analogy — but the plan cites it as a warning, not a proof, and characterises it fairly.
- **arXiv 2603.23064, the heartbeat paper** — cited in Phase 7 and Phase 8 for the 91% figure. **I fetched it.** Title *"Mind Your HEARTBEAT! Claw Background Execution Inherently Enables Silent Memory Pollution"*; abstract carries *"promote short-term pollution into durable long-term memory at rates up to 91%"* and the conclusion carries *"prompt injection is not required."* Both phases quote it accurately. One note rather than a finding: the synthesis grades this source **"ONE READING, NOT VERIFIED"**, and neither phase repeats that grading while leaning on it. It happens to hold — but the plan did not know that, and a reader inherits confidence the pool did not issue.

**Priors the plan rests on and DOES name:** the four concepts added beyond the synthesis are each listed with why they are not from the pool, which is the discipline working. **One it does not name:** the roadmap's commitment 6 reverses synthesis §12's split retention (*"The authored record never prunes… The transcript prunes on a schedule"*) and §14's *"Prune `transcript/` inside a run folder and keep the rest"*. The reversal is argued at length and recorded as a ruling, which is right — but the § *Where each adopted concept is planned* table still maps §12 to *"1 (r7, contents table), 5 (r2, the budget)"* as though it were adopted whole. A reader checking coverage there will not learn that half of §12 was overturned.

**A synthesis claim no phase acts on:** § *What this does NOT settle*, first bullet — *"The three questions the journal must answer. Operator's, and they decide the format. **Nothing else does.**"* The roadmap's § *Open inputs* carries four items and this is not one of them. Item 3 (journal format at this volume) is adjacent but is about volume, and Phase 1 § *Notes* explicitly declines the presentation question. **A question the pool says decides the format is not carried by the plan that decides the format.**

---

## What the plan does not settle

Each named at the phase that consumes it. All four of the roadmap's own Open inputs are correctly placed and correctly left unchecked; these are what I would add:

- **Phase 2 r7(a)** — per-run vs. root-level-shared content store is unmade, and it is load-bearing in two other phases: Phase 5 makes a reachability pass a *prerequisite* conditional on it, and Phase 7 r2's "does a shipped bag validate" answer flips on it. It is correctly carried by Phase 2's checklist; flagging it because it is the single unmade decision with the widest reach.
- **Phase 2 r7(c)** — whether this phase fetches at all. It is the difference between "tee an existing tool" and "build an SSRF-capable fetcher whose responses are durably stored and re-servable offline", and it is the whole spread of my Phase 2 estimate.
- **Phase 3 r13(c)** — deliberately unruled, and the doc explains why. No objection; noting it because the roadmap's WD dependency bullet is the only place that says which sliver remains open.
- **`over-budget-unreclaimable` has no reader** (Phase 5 r4). Phase 3 r11 established the rule one phase earlier; Phase 5 does not apply it to its own new signal.
- **Documentation Standard § *Phase Numbering* rule 4 (binding, 2026-08-20)** — *"A phase is CITED BY NAME, never by number — everywhere a human reads"*. This component is fully unconverted (numbered citations throughout the roadmap and all nine phase docs), which the standard permits under its rolling migration — but it also says *"a half-converted one is worse than either state"*, and nothing in the plan carries the conversion. Surfacing rather than filing: the standard is vendored and I may not edit it, and this is a corpus-wide migration rather than this component's phase. *(Separately, that standard has two rules numbered 4, and they point opposite ways — the second says "Cross-references use phase numbers only." Not mine to fix; vendored.)*

---

## Corrections I made, and whether any moved an estimate

Every one is a **determined** defect — one right answer, no design choice. **None changed an estimate.** Everything that was a judgement is reported above and left alone.

| # | Where | What | Why determined |
|---|---|---|---|
| 1 | `roadmap.md` status line + § *The order* | neither recorded that the run-identity phase was **built 2026-08-24**; the list still read "could start today" | its own phase doc states the build date and PR #138 is in the log |
| 2 | `roadmap.md` § *Why the protocol is built as activities* | "Phase 5's requirements **2–6** are ungated" → **2–9** | Phase 5 labels r2–r9 `(ungated)`, and the roadmap's own gate table already says *"the other **eight**"* — the file contradicted itself |
| 3 | `phase1_the_run_bag.md` | same range, stated as **2–5** → **2–9** | same source of truth |
| 4 | `phase3_the_emit_rule.md` | same range **2–5** → **2–9** | same |
| 5 | `phase1_the_run_bag.md` § *Measurement* | *"stated because **496** bytes reads like nothing"* → **523** | four lines above, the same doc says the 496 figure is superseded and *"The old number is not re-stated anywhere"* |
| 6 | `phase3_the_emit_rule.md` § *Dependencies* | *"Phase 1's **496**-byte floor"* → **523** | same rule, second live restatement |
| 7 | `phase3_the_emit_rule.md` § *Schema evolution* | *"§ Open inputs, item **6**"* → item **2** | Open inputs has four items; item 2 is the event-schema-versioning one |
| 8 | `phase5_snapshots_then_retention.md` | two section headings citing the wrong requirement: *"outlive what it deleted — requirement 6"* → **8**, and *"Retention meets a gapped bag — requirement 7"* → **9**; plus the in-body *"Requirement 6's retention events"* → **8's** | r6 is fleet-code-only, r7 is the post-deletion rebuild, r8 is the carry and r9 is incomplete/redacted independence. The list is unambiguous |

**What I found and deliberately did NOT fix:** `phase7_s3_aggregation.md` repeats two near-identical paragraphs — § *The layout is `<machine_id>/<run_id>`* and § *Origin is where an object is* both open *"The prefix is the origin authority, and the in-event `edge_id` is not"* and both restate the credential-scoping control. Which copy to keep is an editorial judgement, and the Documentation Standard's *cross-reference instead of repeating* does not say which side wins. Reported, not edited.

**Nothing else was touched.** No phase renamed, renumbered or deleted; no completion criterion reworded; no checkbox ticked (13 before, 13 after, verified mechanically); no phase-doc reference added or dropped (verified as a Counter diff against `origin/main`); `sprint.md` read and not written.

## Verification

- `bash testing/run-all.sh` → **PASSED** (3,510 unit passed / 8 skipped; 177 integration passed)
- Estimates in `roadmap.md`: **9** against a floor of 9 — one per phase doc, exactly one per phase
- `candidates.md`: one row appended inside the table block, `decision` blank, `status: open`, `component: persistent-memory-protocol`; § *Where things stand* untriaged total moved 71 → 72 and `test_candidates_prose_matches_the_table.py` is green


Pumapumapumas:
## Decision Log

**Rigour tier: one file — 0 mutations.** The deliverable is nine hour figures and eight one-line prose corrections in planning docs. No code path changed, so there is nothing to mutate; the guards that matter here are the workflow's own (estimate count, phase-link parity, checkbox parity, candidate-column parity), and each was run explicitly against `origin/main` before committing.

- **[Low]** Wrote **`~0 hrs remaining`** beside the complete phase instead of omitting an estimate there. Alternatives: omit it (what my task fragment instructs), or size the phase's historical cost. Why: **the fragment and the enforcing code disagree, and I followed the code.** The fragment says a complete phase *"gets none"* and that *"the unsized check compares a TOTAL against a TOTAL and a complete phase is outside both"* — but `plan_verify_activities.sizing_floor` is `max(len(docs), 1)` where `docs` counts **phase-doc files on disk**, so the floor here is 9, not 8, and omitting would have failed the whole run. Zero is also the true remaining cost, so it cannot corrupt a total. I flagged the contradiction under Friction rather than papering over it. **This is the decision I would most want a human to check.**
- **[Low]** Sized the run-identity phase as a **~10 hr residual** rather than not sizing it. Alternatives: treat it as complete and give it zero; size the whole phase including work already done. Why: it is built but explicitly **not** complete — two requirements are open by design, and both carry real remaining work. Zero would understate it and a full-phase figure would double-count ~20 hrs already spent. A residual is the only figure that feeds a scheduling decision honestly, and I labelled it as one in the roadmap so nobody sums it wrongly.
- **[Low]** Put **~45 hrs** on the emit rule, which is ~45% above the largest phase sized anywhere in `sprint.md`. Alternatives: split the figure across the two halves I recommend splitting, or trim it toward the sprint's observed ceiling. Why: sizing a split the operator has not approved would presume the decision, and trimming to make the total look reasonable is the one thing the brief forbids. The number carries the split recommendation beside it instead.
- **[Medium]** Named the **model-issued write half** as the weakest place rather than the Phase 6 gate depth. Alternatives: lead with the gate, which is larger, more obviously alarming, and transitively behind a whole sibling component. Why: the plan **sees** the gate — it split Phase 6 out of Phase 8 specifically to shorten it and records the collision as a scheduling input. What it does not see is that the harvest half has no *standing guard*, only a one-time demonstration. A cold reader's value is the thing the author cannot see, so the unseen one leads and the seen one is reported second with that reasoning stated.
- **[Medium]** Fixed the **496 → 523** byte figure in two places rather than reporting it. Alternatives: report both, since one of them is inside the doc that declared the supersession and could be read as deliberate. Why: the same doc says four lines above that *"The old number is not re-stated anywhere"*, so the restatement contradicts its own sentence and there is exactly one correct value. That is the fragment's determined-defect case, and the fix is two characters against a paragraph of description.
- **[Medium]** Corrected the ungated range to **2–9** in three documents. Alternatives: correct only the roadmap and report the two phase docs. Why: the roadmap contradicted **itself** — its gate table already said *"the other eight"* against prose saying 2–6 — so the answer was already decided inside the file. Leaving two of three stale would have left a reader who opens the phase doc first still wrong, which is the failure mode the correction exists to close.
- **[Medium]** **Did not** deduplicate the two near-identical paragraphs in the S3 phase. Alternatives: collapse one into a cross-reference, which is what the Documentation Standard prefers in general. Why: which copy survives is an editorial judgement — the two sit under different headings doing different rhetorical jobs — and two competent readers could land differently. Reported instead. This is the line between fixing and reporting, and I put it on the report side deliberately.
- **[Medium]** Filed **one** candidate rather than three. Alternatives: also file the missing Architecture-Standard §4 amendment, and the synthesis-§12-coverage-table gap. Why: I checked §4's Memory table against what Phase 4 actually changes and it is **not falsified** — it states each surface's contract and lifecycle, not which one is authoritative, and the authority claim lives in `memory-model.md`, which the roadmap's amendment 2 already covers. The coverage-table gap is a note about a reading aid, not capability that would be added. Both are reported; neither earned a row.
- **[Low]** Left the `evidence_set_hash` and `over-budget-unreclaimable` no-consumer observations as **report-only**. Alternatives: fold them into the candidate as one producer-with-no-consumer proposal. Why: they fail the clustering test — different phases, different remedies, and both are arguably already covered by Phase 6 r3's blank-cell rule when that phase runs. Bundling them into the harvest candidate would have made one row need three separate rulings, which the finding-quality rule calls a defect rather than a formatting choice.

## Deferred Work

- **A standing check for the emit rule's model-issued half** — Why deferred: it is capability that does not exist and was never this run's scope; this is a read-and-size dispatch with a write grant covering `roadmap.md` and `candidates.md` only, and the build decision (enumerating sweep vs. per-run reconciliation) belongs to whoever builds the phase. Tracked at: **`C-6i1u3b1f`** in `docs/standards/architecture/research/candidates.md`, placed in this PR's diff. Verified by: `grep -c "C-6i1u3b1f" docs/standards/architecture/research/candidates.md` → **1** on this branch, cells read `component: persistent-memory-protocol`, `Source: PR for plan-verify (PMP sizing pass, 2026-08-25)`, `decision` blank, `status: open`; and `git show origin/main:docs/standards/architecture/research/candidates.md | grep -c C-6i1u3b1f` → **0**. **Merge caveat stated plainly: this pointer is true on this branch and only becomes true on `main` at merge.** It is placed here rather than merely surfaced because `finding-routing.md` §4 requires a producing run to place its own proposal in its own PR; if this PR is closed unmerged, the row goes with it. The derived untriaged declaration was moved 71 → 72 in the same change and `test_candidates_prose_matches_the_table.py` is green.

**Surfaced with NO pointer, because I could not honestly give one:**

- **The phase-naming migration** (Documentation Standard § *Phase Numbering* rule 4, binding 2026-08-20). This component is fully unconverted and the standard permits that under rolling migration — but it also says a half-converted component is worse than either state, and nothing in the plan carries the conversion. I have no home for it: the standard is **vendored** so I may not edit it, it is not capability being added so it is not a candidate, and it is a corpus-wide migration across 31 components rather than one of this component's phases. An honest "no home" beats a plausible pointer.
- **The same standard carries two rules numbered 4 that point opposite ways** — rule 4 says cite a phase by name and never by number; the second rule 4 says *"Cross-references use phase numbers only."* A reader can quote either and be quoting a binding rule. Vendored, so it is amended upstream in MDC-Master-Planning and re-vendored, never edited here. No pointer offered.

## Post-Run Reflection

**Friction:**

- **The task fragment and the enforcing code disagree about complete phases, and following the fragment fails the run.** The fragment states twice, emphatically, that a complete phase gets no estimate — *"sizing four planned phases when the roadmap lists five is the correct outcome and not a gap to explain away"* — and asserts that *"the unsized check compares a TOTAL against a TOTAL and a complete phase is outside both."* That assertion is **false against the shipped code**: `plan_verify_activities.sizing_floor` returns `max(len(docs), 1)` and `docs` is phase-doc **files on disk**, which includes the complete one. Here that is a floor of 9 against 8 planned phases. A run that obeys its brief literally raises `plan-verify left <component> UNSIZED` and loses everything it did correctly. I only caught it because I read the workflow source before writing; a run that trusted the fragment would have failed at the last guard. The fix belongs in the Python tree — either exclude phases the roadmap marks complete from the floor, or change the fragment to say what the code actually does.
- **The fragment's "one estimate per phase" warning and the floor pull in opposite directions on this component.** It warns that two figures beside one phase satisfy the count while another phase goes unsized, and that *"nothing catches that but you."* True — and the only way to satisfy a floor of 9 with 8 phases needing forecasts is to write a ninth figure. Zero is the honest one, but the fragment gives no guidance for that case, so I had to invent the convention and document it in the roadmap so a later reader does not sum it in.
- **`sprint.md` carries hour totals for three components and none for this one.** That is exactly what this run supplies, so it is not a defect — but it does mean the calibration available to a cold sizer is one component's per-phase figures plus two component-level totals. I leaned on what the component had already shipped (measured LOC across `modules/journal/` and its eleven test modules) as a second anchor, and said so in the report. A stated calibration source in the fragment would make estimates across runs comparable instead of each one improvising its baseline.

**Project-level suggestions (this repo):**

- **The ungated-range drift is a derive-not-declare instance in planning prose.** One range (Phase 5's `(ungated)` labels) was restated in four places and was wrong in three of them, including inside the file that owns it. `C-523klr8n` already names this mechanism and its shipped check covers *figures with a denominator*; a requirement range is neither a denominator figure nor a status, so no gate could see it. Not filing — this is the recurrence shape the routing rules say to **report where I am already working** rather than amend somebody else's row, and it is reported here with its count of four.
- **Requirement-number references inside a phase doc's own section headings have no check.** Two of Phase 5's headings cited requirement numbers its own list contradicts, and both survived the phase doc's authoring plus a full plan pass. A doc-local check — every `— requirement N` heading resolves to an item that exists and whose subject matches — is mechanically cheap and would have caught both. Raising, not filing: it is adjacent enough to `C-w8455f0l`'s exercise-the-registry argument that a triager should decide whether it expands that row rather than becoming a new one.

**Tooling-level suggestions (claude-dot-files):**

- **This is the claude-dot-files repo and the change is in its own workflow tree, so I am flagging rather than editing** — my write grant here is `roadmap.md` plus `candidates.md`, and the fix is in `modules/assistant/plan/`. Two items, both from the Friction section above: (1) reconcile `sizing_floor` with the prompt fragment's complete-phase rule — the docstring already reasons carefully about the all-gated case reaching zero, and the complete-phase case is the same class of off-by-one in the other direction; (2) give the fragment a stated calibration source for estimates, since it currently asks for a basis clause per phase but supplies no fleet-wide anchor to state it against.
- **The fragment's "counted in code, authoritative" preamble told me this component has 9 phase docs and a roadmap carrying 12 phase references, and that was genuinely useful** — it let me skip re-deriving the inventory and go straight to reading. Worth keeping. What it does not carry is the **completion state** of those docs, which is the one input the sizing rule turns on. Adding "N of them are marked complete" to that preamble would remove the ambiguity at its source rather than in the prose.

Issue #149, opened 2026-08-28
RECURRENCE on C-8tv8ewto — the duplicated runner failure block is at ten and the guard that should see it structurally cannot


Pumapumapumas:
---
store: candidates
status: open
filed_by: review-pr
component: workflow-decomposition
---

**RECURRENCE on `C-8tv8ewto` — not a new item. Harvest action: increment its `count` to 2 and append the dated line below under `## Recurrences`, then close this intake. Do NOT create a second candidate.**

**Why this arrives as an intake rather than an edit:** `review-pr` is decide-only and holds no branch, so it cannot commit to `tracked/`. [Tracked Items §5.0](../../docs/standards/documentation/tracked_items_standard.md) exempts intake from §5; the harvest is the writer.

**Recurrence line to append:**

> **2026-08-28 — the population is TEN, not seven, and the remedy is still unapplied.** Measured by `plan-verify` on PR #145: `except RuntimeError` appears in **all eleven** runners under `scripts/workflows/temporal/`, and **ten** carry the byte-identical `# Nothing has been created yet — that is the point of preflight.` / `print(f"\n✗ {exc}", file=sys.stderr)` / `return 1` block. `grep -rn "parse_or_exit"` returns **0 hits**, so the proposed `parse_or_exit(parser, argv)` in `preflight.py` has not landed. The item records seven, true when filed on evidence from PR #93 (2026-08-16). **Growth from 7 to 10 in twelve days with the duplication ratchet green throughout** — the ratchet's population is `ASSISTANT.rglob("prompts/*.md")` (`test_prompt_blocks_are_shared_not_copied.py:137`), so a Python runner cannot add a row to it and cannot turn it red. *Dual-mode children* adds nine more runners on the same template, taking the corpus to twenty. That slope is the argument `C-7q2m4xzb` makes, now with a measured rate.

**Why this is a recurrence and not a re-report:** the new information is three additional consumers that did not exist when the item was filed, measured independently on a different date. Recurrence outranks age in triage, and this item's `count: 1` currently understates a population that has grown by 43% since filing.

**Verified before filing:** `cat tracked/candidates/C-8tv8ewto.md` on PR #145's branch → `status: open`, `count: 1`, `filed: 2026-08-26`, `filed_by: triage-candidates`, `component: workflow-decomposition`, body reads *"Seven consumers now carry it byte-identically"*.

Pumapumapumas:
[...]

Issue #155, opened 2026-08-29
A cross-document pointer at a requirement number silently retargets when a re-plan renumbers, so a builder scopes the wrong work


Pumapumapumas:
---
store: candidates
status: open
filed_by: review-pr
component:
---

**PROPOSAL — cross-document citations of a requirement need a stable key, or at minimum a dated read.**

**Measured on PR #145, 2026-08-29.** An operator re-plan rewrote `phase4_nothing_invisible.md`'s requirement list and **re-used numbers 4, 5 and 6**. `phase6_every_producer_names_its_consumer.md` cites those three numbers as the *source* of its own requirements 1, 3 and 5; `phase5_configuration_a_run_absorbed.md` cites `phase4` requirement 1 and `phase3_dual_mode_children.md` cites requirement 2's verbosity clause. **Six pointers across three documents retargeted in one commit**, every one landing on a requirement that is not the one named. Nothing went red. It survived two `review-pr` passes and three `plan-verify` cold reads, and was found by the producing run reading its own diff.

**Consequence.** A builder following such a pointer reads the wrong requirement and scopes the wrong work — and the load-bearing case is the *source attribution* shape (`phase6`'s three), where the pointer is the only record of what a split carried across.

**Why this is NOT `C-07stuzb9`, which was read in full before filing.** That candidate proposes a gate keyed on the literal `Phase <digit>` in `docs/development/**` prose, and its hard half is exempting a completed phase's own narration. **This is the opposite asymmetry, and the distinction is the whole point:** a *phase* number is stable identity — the file survives a re-plan, so the citation stays correct. A *requirement* number is a position in a list that a re-plan exists to rewrite. Different key, different predicate, different exemption problem, and the two would be ruled differently. Also checked and ruled different: `#152` (a planning doc citing a *tracked-item id* that resolves to nothing — a different namespace and a different resolver).

**Two remedies, and triage should rule between them rather than on the first one.**

- **(a) Date the read**, which is what PR #145 applied by hand: a cross-document requirement pointer carries the date of the requirement list it read (`pre-split requirement 4 (2026-08-18)`). Checkable syntactically. **It is detection, not prevention** — it makes a stale pointer findable, it does not stop one being written.
- **(b) Give requirements a stable key** — a slug or anchor per requirement, cited instead of the ordinal. This is the standard remedy for a cross-document reference keyed on a mutable ordinal, and it removes the failure rather than reporting it. It costs a format change across every phase doc in the repo.

[...]

PR #162, opened 2026-09-01
build-draft: a run derives its context once, at the boundary, and says so

Pumapumapumas:
Implements [Workflow Decomposition Phase 4 — *Nothing a run relies on is invisible*](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/development/edge-assistant/workflow-decomposition/phase4_nothing_invisible.md), derived half.

## Summary

A wrong flag fails loudly at parse time. A wrong **derivation** produces a plausible wrong run — the workflow competently plans the wrong component, opens a real pull request, and nothing goes red. This closes that class, not by auditing derivations but by making the question smaller: a run has **one** place its derived values come from, that place is built once at the dispatch boundary, and the run says what it built before it spends anything.

- **`scripts/workflows/temporal/scripts/dispatch_context.py`** — a frozen `RunContext` extending `RunIdentity`, carrying the repo root, journal root, workflow key, worktree name, PR number and target alongside the three identity fields. Every field carries `field(metadata=derivation(marker=…, algorithm=…, override=…, scope=…))`, and `dataclasses.fields()` is the population that holds that honest.
- **The worktree name is derived once**, in `RunContext._worktree_name`, as `f"{workflow_key}-{int(clock())}"` — **one expression reached by both constructors**, which the refine pass made true (it was written twice). Eleven `worktree_add` call sites now receive the name instead of assembling it.
- **`run_build_minor` is renamed, deliberately.** Its `workflow_key` was `build-minor` while it named its trees `build-<ts>`, so its worktrees were indistinguishable on disk from the `build` parent's — which is exactly what `/cleanup-merged-worktrees` and the journal's `Journal-Worktree` field use to tell one run from another. Trees are now `build-minor-<ts>`.
- **`review_pr` is reconciled.** `run_review_pr.py` recorded `worktree_name=None` with the comment *"the ONE workflow that cuts no worktree"* while `review_pr_workflow` cut one through the same `worktree_add` whose docstring opens *"ISOLATION IS AN INVARIANT, NOT A PARAMETER"*. The bag now records the stem the tree is named from.
- **The journal root is resolved at the boundary** and passed into `open_run_bag`, which retains its own resolution for callers with no context (`validate_bag`, the tests).
- **`ctx.echo()`** states the context on stderr before the bag opens, the worktree is cut, `gh` runs or the model is invoked.

## Deviation summary — planned vs built

| Planned | Built | Note |
|---|---|---|
| r1 frozen context extending `RunIdentity`, ≥9 fields | ✅ as planned | 6 new + 3 inherited |
| Field **docstring** naming marker / algorithm / override / scope | ⚠️ **`field(metadata=derivation(...))` instead** | **Argued deviation, not silent.** Python has no field docstrings — the nearest convention is a `#:` comment the interpreter discards — so a check asserting *four named parts* would have to parse them back out of one prose blob, making the check's population the parser's guesses rather than the four parts. Metadata keeps the statement at the field where a reader meets it and makes each part separately addressable. The population (`dataclasses.fields()`) is unchanged. Reasoned in `derivation()`'s own docstring. |
| r2 check keyed on `worktree_add`'s call sites | ✅ as planned | Matched on the callee's **name**, so `_shared.worktree_add` is as visible as `act.worktree_add` |
| r3 echo gated on **constructed-here** | ⚠️ **gated on `is_the_run`, not on `minted`** | The plan named `minted` as the discriminator. `minted` is **False** for an operator retrying with `--run-id X` — a parent, constructing its own context, and the caller who most needs to see what the retry derived. `writer` is the field that answers *did I build this context or was I handed one*. Corrected with a test that names the reason. |
| r4 dry-run and live print the same object | ✅ as planned | One `render()`; an AST sweep holds every `--dry-run` branch to using it |
| r5 wrong derivation demonstrated | ✅ as planned | Live path, real repo, verbatim transcript |
| r6 field-documentation check over `dataclasses.fields()` | ✅ as planned | Plus a floor keeping `scope`/`algorithm` from degenerating into noun phrases |
| Worktree name from the workflow key at every `worktree_add` | ⚠️ **the per-pass review tree is NOT a context field** | A parent calls `run_review` once per loop-back, each pass cutting its own tree on the PR's branch. By the plan's own test (*run-scoped, derived once at the boundary*) that value is excluded, the same way `paper_currency` and `due_papers` are. `run_review` now **receives** the stem and appends `-review-<pr>-<pass>`; the check's property is *received, never assembled*, which covers both shapes without an exemption list. **Hardened in the refine pass** — as first built the property admitted a parameter *rebound* on the next line and a received stem plus a fresh clock; both were measured passing and both now fail. |

**Plan facts that did not survive verification** (reported, then built on what is true): the evidence table names `run_plan_verify.py:321`, which **does not exist** — the real site is `run_plan_refine.py:360` — and `run_plan_sprint.py` had moved from `:98` to `:137`. The **total of eleven is unchanged**; the membership had drifted again, inside a week, which is that section's own thesis landing on it for the third time. The `worktree_add` call-site count (eleven plus the definition) verified exactly, as did the `_shared.` alias claim and `test_isolation_invariants._classify`'s blindness to it.

### Draft pass — addressed at authoring time

- **[Critical] `run_plan.py`'s boundary block sat outside the handler.** `resolve_identity`'s refusal of a bad `--run-id` and `open_run_bag`'s refusal of a full journal — both `RuntimeError`s carrying operator-facing remedies — reached the operator as a traceback in this one file of eleven and as a one-line diagnostic in the other ten. Pre-existing on `main`; resolving the journal root at the boundary added a **third** raise to the unguarded region. **Fixed**, and the guard that should have caught it now checks placement rather than only presence.
- **[Warning] The review tree lost the PR number.** The stem is `<workflow-key>-<ts>`, so two `review-pr` dispatches aimed at different PRs starting in the same wall-clock second would have shared a directory on their first pass. **Fixed** — `f"{worktree_name}-review-{task.pr_number}-{this_pass}"`, with three collision tests that read the f-string out of the module rather than retyping it.
- **[Warning, standards] The transcript and checkboxes belong in the phase doc.** **Delivered** as helloskyy-io/skyynet-master-planning#4.

### Refine pass — six defects, each confirmed by mutation BEFORE it was fixed

**Rigour tier: safety control.** Six mutations to establish the defects, six negative controls after; predicted red counts stated first, **all twelve exact**.

- **[Critical — `code-reviewer` and `quality-control`, independently] The worktree name was derived TWICE, inside the object built to end exactly that.** `RunContext.build` and `.for_dry_run` each carried `f"{workflow_key}-{int(clock())}"`, and **every** worktree-name assertion in the suite drove the rehearsal — nothing in the tree called `RunContext.build` at all. **Measured: perturbing the LIVE copy alone left 3051 tests green** while a `--dry-run` previewed a name no run would use. That is requirement 4's own defect (a rule with two statements that agree until one is edited) reproduced one level inside requirement 4's fix, and the same shape as `run_build_minor` naming its trees `build-…` under `workflow_key="build-minor"`. **Fixed:** one `RunContext._worktree_name`, and `test_the_REHEARSAL_and_the_LIVE_run_render_THE_SAME_OBJECT` now drives `build` against `for_dry_run` on a pinned clock and compares **fields** as well as renderings — it previously constructed its comparison object from the rehearsal's own field values and could not fail. *Control: re-splitting the expression → predicted 5 red, observed 5.*
- **[Critical — `quality-control`] The `target` field named FOUR targetless workflows; `plan-revision` is a fifth.** A false statement in the one object whose stated purpose is that its fields state themselves — and it is exactly the failure its own guard admits it cannot catch (*"asserts each part is PRESENT, never that it is TRUE"*). **Fixed.**
- **[Warning — `code-reviewer`] The placement guard watched the OLD raise and not the one this PR added.** `_bag_open_outside_the_handler` keyed on `open_run_bag` only, while `run_plan.py`'s own comment says resolving the journal root at the boundary *"added a THIRD raise to the unguarded region"*. **Measured: hoisting `ctx = RunContext.build(...)` above the `try` while leaving `open_run_bag` inside it left the whole tier green** — a `JournalRootError` then reaches the operator as a traceback, which is the regression the guard was written after. **Fixed:** `_MUST_BE_GUARDED = (BAG_OPEN, "build")`, with two new control shapes. *Control: predicted 1 red, observed 1.*
- **[Warning — `code-reviewer`] Requirement 3's ordering was held on ONE entrypoint of eleven.** The entrypoint sweep collected call *names* anywhere in the file. **Measured green:** moving `ctx.echo()` below `open_run_bag`; and `if a.verbose: ctx.echo()`, which the requirement rules out in terms. Only `run_plan_draft` was protected, by the live demonstration. **Fixed:** `test_every_entrypoint_SAYS_IT_BEFORE_THE_BAG_OPENS_and_UNCONDITIONALLY`, with literal-driven controls including the one that must NOT fire — a `--dry-run` early return above the echo is not a gate on it. The `--dry-run` sweep likewise now requires the `for_dry_run` call to sit **inside** the rehearsal branch. *Controls: predicted 1 red each, observed 1 each.*
- **[Warning — `code-reviewer`, confirmed by probing the predicate] The r2 guard admitted two shapes its own property forbids.** A `worktree_name` **parameter rebound on the next line** passed — the half-migration shape (signature converted, body not), at the site the guard's docstring calls *"the one this guard most needs to see"*. And `f"{worktree_name}-{int(time.time())}"` passed — a re-scatter wearing a suffix, referencing a received parameter. **Fixed:** a name leaves `received` the moment anything in the function binds it, and a clock read at the call site is assembly whatever else the argument references. Four new controls. *Control: predicted 1 red, observed 1.*
- **[Warning — both] Prose that was false against the tree, swept as a class rather than at the site reported.**
  - `render()`'s header said *"before anything is created"* while `build()` had just `mkdir`'d the journal root at 0700. Reworded to *"before this run cuts, posts or spends"*, with the reason at the line.
  - Ten entrypoints carried *"BEFORE THE BAG OPENS, THE WORKTREE IS CUT OR ANY `gh` CALL RUNS"*. **Swept all eleven by AST, scoped to `main()`: the claim is false in three** — `run_plan_revision` (`base_ref`), `run_plan_refine` and `run_plan_sprint` (`unclosed_hold`, `pr_branch`). Those reads are **refusals** and correctly precede the bag, so the three comments now say what is true rather than the code being shuffled to match a comment.
  - `docs/file_structure.txt` said *"a local assignment is not received, which is what all eleven sites were"* — while the entry three rows below says the 11th was inline. Corrected; that belief is what shipped `base_ref`'s first guard green.
  - `docs/file_structure.txt` said *five* test modules drive an entrypoint `main()` in two places. **This PR made it six** and bumped the pin without sweeping its own blast radius.
- **[Warning — `code-reviewer`] Two vacuity floors sat below their real populations** — `>= 10` against eleven `worktree_add` calls, `>= 6` against nine context fields, the latter with a message insisting on nine. Both raised; the field count is now one named constant so message and assertion cannot disagree.
- **[Warning — `quality-control`] The census pin moved 33→35 / 22→24 with none of the narration every prior bump in that file carries.** Added.
- **[Info — `code-reviewer`] `context_field_documentation()` was dead code** arguing for a population nobody read. **Fixed by giving it its one caller** — the r6 guard now obtains the population through it, so the argument in its docstring is load-bearing. *Control: making it drop a field → predicted 1 red, observed 1.*
- **[Info — `code-reviewer`] `open_run_bag` silently voids `config_path`/`env` when `journal_root` is supplied.** No caller does this today. **Documented rather than guarded** — a guard for a combination nobody writes is speculation; a silent precedence nobody documented is how a redirected test asserts against the wrong root and passes.
- **[Info — `code-reviewer`] `env`/`clock` untyped, and `derivation()` lives in the identity module.** Types added (`Clock` is now named); a pointer added at the import site.
- **[Info — `code-reviewer`] `_review_tree_name` failed with a bare `ValueError` from `str.index`** if the call it scrapes were reformatted. Now a readable assertion naming the file and forbidding the obvious wrong fix (retyping the template).

**Rejected with reasoning:**

- **[High, structure — `code-reviewer`] Collapse the eleven hand-copied boundary blocks into `open_run_bag(ctx=ctx)`.** Real, and it is the *surfaced candidate below* rather than a rejection of the substance. Not done **here** because it changes all eleven entrypoints' control flow and the six-keyword forwarding is only the visible half — the sequence, the handler placement and the comment are the rest. That is a scope this phase names as out (*"it does not add a new derived value; this phase gives them one home and one voice"*). **What changed in this pass is that the drift is now caught**: the placement guard covers both boundary calls, and the ordering guard covers the echo.
- **[High, structure — `code-reviewer`] Reclassify `review_pr_workflow` as a parent in `test_isolation_invariants`.** Real and surfaced below as a defect. Not fixed here because reclassifying applies a **different set of parametrised invariants** to that module — a behaviour change to an existing guard, in a file this PR does not otherwise touch.
- **[Medium — `code-reviewer`] Extract `_per_pass_tree_name` into `review_pr_workflow` so the test need not scrape the f-string.** The brittleness was real and is fixed at its actual cost (a readable failure). Adding an indirection to production code purely so a test can import it moves the second statement of the rule rather than removing it — and the scraping is what makes the test *derived* rather than a copy, which is this phase's own thesis.
- **[Note] The bag records a stem, not the exact directory.** `Journal-Worktree` is `review-pr-<ts>` while the tree is `review-pr-<ts>-review-<pr>-<pass>`. Deliberate: the pass number is not knowable at bag-open (it costs a `gh` round trip) and the bag opens before the first side effect. **The refine pass carried out the disposition** — the stem-vs-path relationship is now stated in `open_run_bag`'s own docstring, where a consumer meets the field, rather than only at the two call sites.
- **[Low] `review-pr-<ts>-review-<pr>-<pass>` says "review" twice.** Cosmetic; for every other workflow the stem is the workflow key and the name reads correctly.

**Surfaced, not filed — for `review-pr` to route:**

> ### A dispatch boundary helper would close the drift class structurally — EXPANSION of `C-8tv8ewto`
>
> **This is an EXPANSION of an existing candidate, not a new one, and the draft pass surfaced it without that pointer.** [`tracked/candidates/C-8tv8ewto.md`](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/tracked/candidates/C-8tv8ewto.md) already covers the eleven runners' byte-identical `try/except RuntimeError → print(f"\n✗ {exc}") → return 1` block, with `parse_or_exit(parser, argv)` in `preflight.py` as its remedy; issue #149 is its open recurrence intake, measuring the population at ten as of 2026-08-28. **Verified:** `cat /opt/skyy-net/skyynet-master-planning/tracked/candidates/C-8tv8ewto.md` → `status: open`, `count: 1`, `component: workflow-decomposition`; `gh issue view 149` → OPEN.
>
> **What this PR adds to it:** the same eleven files now also hand-copy a *build context → echo → open bag* sequence **inside** that handler, and the two are one refactor rather than two — a `dispatch_boundary(argv, repo_root, workflow_key, …)` context manager subsumes `parse_or_exit`. **The population grew again in this PR**, which is the recurrence signal `C-8tv8ewto`'s own count exists to carry.
>
> **Harvest action:** append this to `C-8tv8ewto` as a recurrence with the widened remedy. **Do not open a second candidate.** `decision` stays blank — that is `triage-candidates`' output. **Component:** `workflow-decomposition`.

> ### The standards amendment this phase owes `workflow-scripts.md`
>
> **Consequence:** the rule this PR implements is now enforced only by per-value checks. Nothing states it, so the next run-scoped derived value scatters before anyone notices the pattern is already ruled on.
>
> **Remedy:** amend [`workflow-scripts.md`](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/standards/workflow-scripts.md) § *9. Repo Root Operation* — the section that today governs the one derivation this fleet already rules on — with: *a run-scoped derived value is a field on the run context, computed once at the dispatch boundary and passed down; it is not re-derived by a callee.* It also finishes a sentence [`guide/workflows.md`](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/guide/workflows.md) already starts — *"Isolation is established once by the parent and passed down"* — which is this rule stated for one value.
>
> **Store:** `tracked/standards/`. **Surfaced rather than filed** per [`finding-routing.md` §7](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/standards/finding-routing.md): a build dispatch is a producing run, and the phase doc says so in terms — *"Surface it, do not file it, and never edit the standard."*

> ### `test_isolation_invariants._classify` misclassifies `review_pr_workflow`
>
> **Consequence:** `_classify()` keys on the literal `"act.worktree_add("`, and `review_pr_workflow` reaches the function as `_shared.worktree_add`. It is classified a **child** while it cuts a worktree, so `test_child_does_not_create_its_own_worktree` asserts it cuts none — a green assertion that is actively false. The same literal is reused at `test_every_parent_opens_a_run_bag.py:636` and `:684` and at `test_isolation_invariants.py:134`, deliberately, so the miss is shared rather than isolated. *(Line numbers re-derived against this branch's HEAD; the draft pass's `470`/`513` were correct when written and moved when this pass extended the file.)*
>
> **Remedy:** key those predicates on the callee's AST name, the way this PR's two new guards do.
>
> **Store:** `tracked/issues/` — a **defect** in something already built. **Not fixed here** because reclassifying `review_pr_workflow` as a parent applies a different set of invariants to it, which is a behaviour change to an existing guard and outside this phase's scope. **Component:** `workflow-decomposition`.

## Test results

`./testing/run-all.sh` → **RESULT: PASSED** (`python/unit` PASS, `python/integration` PASS, `python/e2e` SKIP — no `tests/e2e` in the tree). Unit tier: **4256 passed, 8 skipped, 0 failed**; integration **95 passed**. The draft pass measured 3151 in the temporal component alone; the tier total moved because the corpus gates now collect against the sibling planning repo.

**The twelve mutations, predicted before each run and all twelve exact.** Six established the defects above; six are negative controls on what replaced them.

| # | Mutation | Predicted | Observed |
|---|---|---|---|
| 1 | Perturb `build()`'s worktree derivation only (the duplicated expression) | 0 — the defect | **0** |
| 2 | Hoist `RunContext.build` above the `try` in `run_plan.py` | 0 — the defect | **0** |
| 3 | Move `ctx.echo()` below `open_run_bag` in `run_plan.py` | 0 — the defect | **0** |
| 4 | Rebind `worktree_name` inside `run_review` (half-migration) | 0 — the defect | **0** |
| 5 | `f"{worktree_name}-{int(time.time())}"` at a call site | detector silent — the defect | **silent** |
| 6 | `if a.verbose: ctx.echo()` | 0 — the defect | **0** |
| 7 | Re-split the derivation after the fix | 5 | **5** |
| 8 | Hoist `RunContext.build` after the fix | 1 | **1** |
| 9 | Move `ctx.echo()` below bag-open after the fix | 1 | **1** |
| 10 | Rebind `worktree_name` after the fix | 1 | **1** |
| 11 | `context_field_documentation()` drops a field | 1 | **1** |
| 12 | `if a.verbose: ctx.echo()` after the fix | 1 | **1** |

**And the r5 demonstration was reproduced independently**, not read off the diff: a scratch repo with `RIGHT`/`WRONG` components, `run_plan_draft.main(["development/edge-assistant/WRONG", "--repo", …])` on the live path with `open_run_bag` replaced by a raise. The echo named `target : development/edge-assistant/WRONG` before the stand-in fired, and `.claude/worktrees/` did not exist afterwards.

**Coverage.** One new source artifact with logic — `dispatch_context.py` — with four corresponding test modules, all in `scripts/workflows/temporal/tests/unit/` per the Testing Standard's tier layout, all pytest:

| New / changed source | Test |
|---|---|
| `dispatch_context.py` (behaviour, both constructors) | `test_dispatch_context.py` — 31 tests |
| `dispatch_context.py` (field documentation, r6) | `test_every_context_FIELD_STATES_ITSELF.py` |
| The eleven call sites (r2) | `test_a_worktree_NAME_comes_from_the_RUN_CONTEXT.py` |
| The echo on the live path (r5) | `test_a_WRONG_TARGET_is_NAMED_before_the_run_costs_anything.py` |
| `run_plan.py`'s handler placement, and `RunContext.build`'s | `test_every_parent_opens_a_run_bag.py` (extended, with its own controls) |

**Three census guards went red on the new modules and were updated, not silenced** — `_PINNED` (33, 22) → (35, 24), now with the narration every prior bump in that file carries; the entrypoint-driving-module pin 5 → 6, **and its two stale restatements in `docs/file_structure.txt`**; and one prose figure reworded, because it stated a count the journal sweep reads as a claim about the *entrypoint* population.

## Success criteria

- [x] **r1** — A frozen run context exists, constructed once at the dispatch boundary, extending `RunIdentity`, carrying nine fields
- [x] **r2** — No entrypoint and no workflow module assembles a run-scoped derived value for itself, held by a check keyed on the call site, matched on the function name, and blind to neither a rebound parameter nor a clock
- [x] **r3** — A run states its context once, on the live path, before the first side effect, gated on constructed-here — **and the ordering and the gating are now held across all eleven entrypoints rather than one**
- [x] **r4** — The `--dry-run` preview and the live run print the same object, from **one** derivation, with the live constructor in the tested population
- [x] **r5** — A wrong derivation is demonstrated visible, verbatim, on the live path — reproduced independently in the refine pass
- [x] **r6** — Every field states its marker, algorithm, override and scope of effect, with `dataclasses.fields()` as the population, reached through the module's own accessor
- [x] The full suite passes and nothing that depended on the quiet path or on a worktree name's spelling broke
- [x] **The three departures from the requirements' wording are recorded where a planner reads them** — helloskyy-io/skyynet-master-planning#4
- [ ] **The standards amendment against `workflow-scripts.md` § 9** — surfaced above, not filed. The phase doc rules that a producing run surfaces and `review-pr` files, so this is deliberately open rather than met.

### DIFF
--- a/docs/file_structure.txt
+++ b/docs/file_structure.txt
@@ -626,7 +626,7 @@ claude-dot-files/
 │                   │                                #   is DERIVED from splitlines() rather than from a remembered list, and the round trip
 │                   │                                #   is the property: for any accepted value, reading the file returns what was written
 │                   ├── test_dispatch_context.py     # The run context's behaviour: the worktree name follows the workflow key (every key,
-│                   │                                #   because the drift was in one of eleven), the object is frozen, the echo is gated on
+│                   │                                #   because the drift was in a single runner), the object is frozen, the echo is gated on
 │                   │                                #   `is_the_run` rather than on `minted` — a retry supplies its name and is exactly the
 │                   │                                #   caller who needs the echo — and the rehearsal renders the same object as the live
 │                   │                                #   run. Also two AST sweeps over the entrypoints: every one builds and echoes, and
@@ -674,7 +674,11 @@ claude-dot-files/
 │                                                    #   run_*.py by AST (never substring: the sibling sweep shipped green because a COMMENT
 │                                                    #   satisfied its grep) and fails when one starts a run with no bag. Also checks ORDER —
 │                                                    #   the bag precedes the first side effect — with a self-contained non-conforming fixture
-│                                                    #   for both, and names the swept scope in its own failure message
+│                                                    #   for both, and names the swept scope in its own failure message. And PLACEMENT: the
+│                                                    #   call must sit inside a `try` whose except catches RuntimeError, which its failure
+│                                                    #   message had demanded in prose since it was written while nothing checked it —
+│                                                    #   run_plan.py had drifted above the handler, turning a bad --run-id and a full journal
+│                                                    #   into tracebacks in one file of eleven and one-line diagnostics in the other ten
 │
 ├── testing/                                         # TIER 1 + TIER 2 of the Testing Standard's three-tier hierarchy
 │   ├── README.md                                    # How to run the suite and how to add a test
--- a/docs/file_structure.txt
+++ b/docs/file_structure.txt
@@ -223,9 +223,20 @@ claude-dot-files/
 │           │                                        #   a run's CONTENT here — that is Phase 3. Imports no workflow module, so a
 │           │                                        #   measurement helper can load it without dragging in temporalio
 │           ├── scripts/                             # Kickoff entrypoints, until an SDK path replaces them — plus preflight.py,
-│           │                                        #   validate_bag.py and dispatch_identity.py, which are NOT kickoffs: they are what a
-│           │                                        #   run checks before, what an operator checks after, and who NAMES the run. All three
-│           │                                        #   live here because they are launch concerns rather than workflow concerns
+│           │                                        #   validate_bag.py, dispatch_identity.py and dispatch_context.py, which are NOT
+│           │                                        #   kickoffs: they are what a run checks before, what an operator checks after, who
+│           │                                        #   NAMES the run, and what the run DERIVED. All four live here because they are
+│           │                                        #   launch concerns rather than workflow concerns. None is named run_*.py, and that
+│           │                                        #   is load-bearing — the entrypoint sweeps glob that prefix, so a helper wearing it
+│           │                                        #   is swept as a runner and fails five guards at once (measured 2026-09-01)
+│           │                                        #   dispatch_context.py is Workflow Decomposition Phase 4's answer to "what did this
+│           │                                        #   run work out for itself": a frozen RunContext extending RunIdentity, built ONCE at
+│           │                                        #   the dispatch boundary, carrying the repo root, journal root, workflow key, worktree
+│           │                                        #   name, PR number and target — each field stating its marker, algorithm, override and
+│           │                                        #   scope of effect. It echoes itself on stderr before the first side effect, and the
+│           │                                        #   same render() is what every --dry-run previews, so a rehearsal cannot show
+│           │                                        #   something other than what runs. The worktree name lived at ELEVEN sites in THREE
+│           │                                        #   spellings before it moved here, one of them already drifted from its workflow key
 │           │                                        #   dispatch_identity.py is Phase 9's answer to "who names a run": it declares --run-id
 │           │                                        #   and --writer once for every entrypoint, validates a supplied name, and mints one
 │           │                                        #   ONLY when nothing supplied it — then prints it, because a name nobody was told is a
@@ -307,6 +318,18 @@ claude-dot-files/
 │                   │                                #   success — $8.60, nine findings left open (MDC #204)
 │                   ├── test_a_gh_reply_is_checked_for_SHAPE_not_only_parsed.py # Every `gh`-reply `json.loads` asks what SHAPE came
 │                   │                                #   back — a decode guard alone lets `{"message": …}` through to a `TypeError`
+│                   ├── test_a_WRONG_TARGET_is_NAMED_before_the_run_costs_anything.py # Phase 4 r5, DEMONSTRATED rather
+│                   │                                                   #   than asserted: a real entrypoint's live main(), pointed at the wrong
+│                   │                                                   #   component, with bag-open replaced by a raise. A wrong target is the
+│                   │                                                   #   one derived value that fails SILENTLY — the run plans the wrong
+│                   │                                                   #   thing competently and opens a real PR. Carries the verbatim transcript
+│                   ├── test_a_worktree_NAME_comes_from_the_RUN_CONTEXT.py # No worktree_add call assembles the name it cuts with.
+│                   │                                                   #   Keyed on the CALL SITE, matched on the function NAME so a module
+│                   │                                                   #   alias cannot hide it (review_pr_workflow reaches it as _shared., and
+│                   │                                                   #   the act.-keyed predicates in test_isolation_invariants are blind to
+│                   │                                                   #   that today). Received = a .worktree_name attribute, or a PARAMETER
+│                   │                                                   #   of that name — a local assignment is not received, which is what all
+│                   │                                                   #   eleven sites were. A fourth spelling fails without being anticipated
 │                   ├── test_a_new_branch_STARTS_FROM_THE_DEFAULT_BRANCH.py # A run with no --pr is cut from the DEFAULT BRANCH, never the
 │                   │                                                   #   operator's HEAD. Keyed on worktree_add's ref argument, so an inline
 │                   │                                                   #   literal is as visible as an assignment — the 11th call site was inline.
@@ -602,6 +625,17 @@ claude-dot-files/
 │                   │                                #   more. Sweeps every f"{label}: {value}" for an undeclared value; the battery beside it
 │                   │                                #   is DERIVED from splitlines() rather than from a remembered list, and the round trip
 │                   │                                #   is the property: for any accepted value, reading the file returns what was written
+│                   ├── test_dispatch_context.py     # The run context's behaviour: the worktree name follows the workflow key (every key,
+│                   │                                #   because the drift was in one of eleven), the object is frozen, the echo is gated on
+│                   │                                #   `is_the_run` rather than on `minted` — a retry supplies its name and is exactly the
+│                   │                                #   caller who needs the echo — and the rehearsal renders the same object as the live
+│                   │                                #   run. Also two AST sweeps over the entrypoints: every one builds and echoes, and
+│                   │                                #   every --dry-run branch renders a context instead of assembling its own preview
+│                   ├── test_every_context_FIELD_STATES_ITSELF.py # Phase 4 r6. Every field of RunContext names its marker, its
+│                   │                                #   algorithm, its override-or-none and its scope of effect. The population is
+│                   │                                #   dataclasses.fields() — a language primitive, which is why this is NOT the tree-wide
+│                   │                                #   enumeration the phase deleted: that one had no population to read (four marker
+│                   │                                #   conventions, zero hits across 225 files). Asserts PRESENCE, never truth
 │                   ├── test_dispatch_identity.py    # The BEHAVIOURAL half of Phase 9 r3 and r4, against a real journal root — the sweeps
 │                   │                                #   are source greps and both requirements fail in ways a grep cannot see. r3's failure
 │                   │                                #   is SILENT (two bags for one run reads as two runs forever); r4's is a child quietly
--- a/docs/file_structure.txt
+++ b/docs/file_structure.txt
@@ -263,7 +263,7 @@ claude-dot-files/
 │           └── tests/                               # TIER 3 — this code unit owns its tests (Testing Standard § per-unit ownership)
 │               ├── conftest.py                      # Component-scoped import path (replaces the per-file sys.path preambles), AND the
 │               │                                    #   session-wide autouse fixture that points every bag the suite opens at a temp
-│               │                                    #   root — the FIX for the defect test_the_suite_never_writes_… guards, since five
+│               │                                    #   root — the FIX for the defect test_the_suite_never_writes_… guards, since six
 │               │                                    #   unit modules drive an entrypoint main() and were writing real bags into the
 │               │                                    #   operator's ~/.local/state/. Redirects CONFIG_PATH, not XDG_STATE_HOME, so a
 │               │                                    #   configured root cannot silently lose the protection
@@ -328,8 +328,11 @@ claude-dot-files/
 │                   │                                                   #   alias cannot hide it (review_pr_workflow reaches it as _shared., and
 │                   │                                                   #   the act.-keyed predicates in test_isolation_invariants are blind to
 │                   │                                                   #   that today). Received = a .worktree_name attribute, or a PARAMETER
-│                   │                                                   #   of that name — a local assignment is not received, which is what all
-│                   │                                                   #   eleven sites were. A fourth spelling fails without being anticipated
+│                   │                                                   #   of that name, NOT rebound inside it. Ten sites were local assignments
+│                   │                                                   #   and the 11th was an inline argument with none — an assignment-keyed
+│                   │                                                   #   guard would have walked past it, which is how base_ref's first guard
+│                   │                                                   #   shipped green. A fourth spelling fails without being anticipated, and a
+│                   │                                                   #   CLOCK read at the call site is assembly whatever else it references
 │                   ├── test_a_new_branch_STARTS_FROM_THE_DEFAULT_BRANCH.py # A run with no --pr is cut from the DEFAULT BRANCH, never the
 │                   │                                                   #   operator's HEAD. Keyed on worktree_add's ref argument, so an inline
 │                   │                                                   #   literal is as visible as an assignment — the 11th call site was inline.
@@ -592,7 +595,7 @@ claude-dot-files/
 │                   │                                #   as one checked at all eight. Missing / mismatched / unlisted are distinguished — the
 │                   │                                #   third is a write landing AFTER the seal, invisible to any check that walks the manifest
 │                   ├── test_the_suite_never_writes_to_the_operators_journal.py # Making bag-open structural made the SUITE a
-│                   │                                #   dispatch: five modules drive an entrypoint main(), so every pytest run left a real bag
+│                   │                                #   dispatch: six modules drive an entrypoint main(), so every pytest run left a real bag
 │                   │                                #   under ~/.local/state — 24 in one day, and the integration tier was grading them as
 │                   │                                #   dispatch output. Keyed on the observable (did the real root change), not the mechanism
 │                   ├── test_journal_containment.py  # The CLASS behind four escapes found one-per-review-pass: an externally-supplied string
@@ -629,8 +632,10 @@ claude-dot-files/
 │                   │                                #   because the drift was in a single runner), the object is frozen, the echo is gated on
 │                   │                                #   `is_the_run` rather than on `minted` — a retry supplies its name and is exactly the
 │                   │                                #   caller who needs the echo — and the rehearsal renders the same object as the live
-│                   │                                #   run. Also two AST sweeps over the entrypoints: every one builds and echoes, and
-│                   │                                #   every --dry-run branch renders a context instead of assembling its own preview
+│                   │                                #   run. Also AST sweeps over the entrypoints: every one builds and echoes, the echo
+│                   │                                #   PRECEDES the bag and is unconditional (presence alone was blind — moving it below
+│                   │                                #   bag-open left the tier green), and every --dry-run branch renders a context inside
+│                   │                                #   that branch instead of assembling its own preview
 │                   ├── test_every_context_FIELD_STATES_ITSELF.py # Phase 4 r6. Every field of RunContext names its marker, its
 │                   │                                #   algorithm, its override-or-none and its scope of effect. The population is
 │                   │                                #   dataclasses.fields() — a language primitive, which is why this is NOT the tree-wide
@@ -678,7 +683,10 @@ claude-dot-files/
 │                                                    #   call must sit inside a `try` whose except catches RuntimeError, which its failure
 │                                                    #   message had demanded in prose since it was written while nothing checked it —
 │                                                    #   run_plan.py had drifted above the handler, turning a bad --run-id and a full journal
-│                                                    #   into tracebacks in one file of eleven and one-line diagnostics in the other ten
+│                                                    #   into tracebacks in one file of eleven and one-line diagnostics in the other ten.
+│                                                    #   Placement covers RunContext.build as well as open_run_bag — both raise a
+│                                                    #   RuntimeError carrying an operator remedy, and a guard watching only the older of
+│                                                    #   the two was measured blind to the newer one
 │
 ├── testing/                                         # TIER 1 + TIER 2 of the Testing Standard's three-tier hierarchy
 │   ├── README.md                                    # How to run the suite and how to add a test
--- a/scripts/workflows/temporal/modules/assistant/build/build/build_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/build/build/build_workflow.py
@@ -108,7 +108,7 @@ def run_build(task: BuildInput, repo_root: Path, worktree_name: str) -> BuildRes
     # --- Steps 2 & 3: REFINE then DISPOSITION, bounded by helper.MAX_LOOPS ---
     loops = 0
     verdict = _refine_then_dispose(task, description, pr, repo_root,
-                                   worktree, notes, correction=False)
+                                   worktree, worktree_name, notes, correction=False)
 
     while helper.should_loop_back(verdict, loops):
         loops += 1
@@ -121,7 +121,7 @@ def run_build(task: BuildInput, repo_root: Path, worktree_name: str) -> BuildRes
                      + (" This is the last automated pass."
                         if loops == helper.MAX_LOOPS else ""))
         verdict = _refine_then_dispose(task, description, pr, repo_root, worktree,
-                                       notes, correction=True,
+                                       worktree_name, notes, correction=True,
                                        loops_left=helper.MAX_LOOPS - loops)
 
     if verdict is Verdict.HOLD_NEEDS_ASSISTANCE:
@@ -155,7 +155,7 @@ def run_build(task: BuildInput, repo_root: Path, worktree_name: str) -> BuildRes
 
 
 def _refine_then_dispose(task: BuildInput, description: str, pr: str,
-                         repo_root: Path, worktree: Path,
+                         repo_root: Path, worktree: Path, worktree_name: str,
                          notes: list[str], *, correction: bool,
                          loops_left: int = 0) -> Verdict:
     """One refine pass followed by one disposition pass."""
@@ -205,9 +205,13 @@ def _refine_then_dispose(task: BuildInput, description: str, pr: str,
     if hold is not None:
         return hold
 
+    # THREADED FROM THE RUN'S CONTEXT, NOT REBUILT. `run_review` cuts a
+    # per-pass tree on the PR's branch and takes the run's own worktree name
+    # as its stem, so the tree it makes is traceable to the run the bag
+    # recorded. See `review_pr_workflow.run_review`.
     result = review_pr.run_review(
         ReviewInput(pr_number=pr, repo_target=task.repo_target, verbose=task.verbose),
-        repo_root,
+        repo_root, worktree_name=worktree_name,
     )
     notes.extend(result.notes)
     return Verdict(result.verdict.value)
--- a/scripts/workflows/temporal/modules/assistant/build/build_minor/build_minor_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/build/build_minor/build_minor_workflow.py
@@ -72,7 +72,7 @@ def run_build_minor(task: BuildInput, repo_root: Path, worktree_name: str) -> Bu
 
     loops = 0
     verdict = _refine_then_dispose(task, description, pr, repo_root,
-                                   worktree, notes, correction=False)
+                                   worktree, worktree_name, notes, correction=False)
 
     # Same bound as the major tier and for the same reason: self-correction
     # plateaus at roughly 3-5 passes, and past it the model justifies rather than
@@ -84,7 +84,7 @@ def run_build_minor(task: BuildInput, repo_root: Path, worktree_name: str) -> Bu
                      + (" The last automated pass."
                         if loops == helper.MAX_LOOPS else ""))
         verdict = _refine_then_dispose(task, description, pr, repo_root, worktree,
-                                       notes, correction=True,
+                                       worktree_name, notes, correction=True,
                                        loops_left=helper.MAX_LOOPS - loops)
 
     if verdict is Verdict.HOLD_NEEDS_ASSISTANCE:
@@ -106,7 +106,7 @@ def run_build_minor(task: BuildInput, repo_root: Path, worktree_name: str) -> Bu
 
 
 def _refine_then_dispose(task: BuildInput, description: str, pr: str,
-                         repo_root: Path, worktree: Path,
+                         repo_root: Path, worktree: Path, worktree_name: str,
                          notes: list[str], *, correction: bool,
                          loops_left: int = 0) -> Verdict:
     ci_settled = wait_for_ci(pr, repo_root=repo_root)
@@ -132,10 +132,14 @@ def _refine_then_dispose(task: BuildInput, description: str, pr: str,
     if hold is not None:
         return hold
 
+    # THREADED FROM THE RUN'S CONTEXT, NOT REBUILT. `run_review` cuts a
+    # per-pass tree on the PR's branch and takes the run's own worktree name
+    # as its stem, so the tree it makes is traceable to the run the bag
+    # recorded. See `review_pr_workflow.run_review`.
     result = review_pr.run_review(
         ReviewInput(pr_number=pr, repo_target=task.repo_target,
                     verbose=task.verbose, review_type=ReviewType.BUILD),
-        repo_root,
+        repo_root, worktree_name=worktree_name,
     )
     notes.extend(result.notes)
     return Verdict(result.verdict.value)
--- a/scripts/workflows/temporal/modules/assistant/plan/plan/plan_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/plan/plan/plan_workflow.py
@@ -79,7 +79,7 @@ def run_plan(*, component: Path, repo_root: Path, worktree_name: str,
 
     verdict = _refine_size_and_dispose(
         component=component, repo_root=repo_root, worktree=worktree,
-        sprint_path=sprint_path, candidates_path=candidates_path, pr=pr,
+        worktree_name=worktree_name, sprint_path=sprint_path, candidates_path=candidates_path, pr=pr,
         repo_target=repo_target, notes=notes, correction_pass=False,
         verbose=verbose,
     )
@@ -118,7 +118,7 @@ def run_plan(*, component: Path, repo_root: Path, worktree_name: str,
         notes.append(f"HOLD (redispatch): loop-back {loops} of {routing.MAX_LOOPS}.")
         verdict = _refine_size_and_dispose(
             component=component, repo_root=repo_root, worktree=worktree,
-            sprint_path=sprint_path, candidates_path=candidates_path, pr=pr,
+            worktree_name=worktree_name, sprint_path=sprint_path, candidates_path=candidates_path, pr=pr,
             repo_target=repo_target, notes=notes, correction_pass=True,
             verbose=verbose,
         )
@@ -142,6 +142,7 @@ def run_plan(*, component: Path, repo_root: Path, worktree_name: str,
 
 
 def _refine_size_and_dispose(*, component: Path, repo_root: Path, worktree: Path,
+                             worktree_name: str,
                              sprint_path: Path, candidates_path: Path, pr: str,
                              repo_target: str | None, notes: list[str],
                              correction_pass: bool, verbose: bool) -> routing.Verdict:
@@ -179,10 +180,14 @@ def _refine_size_and_dispose(*, component: Path, repo_root: Path, worktree: Path
     if hold is not None:
         return hold
 
+    # THREADED FROM THE RUN'S CONTEXT, NOT REBUILT. `run_review` cuts a
+    # per-pass tree on the PR's branch and takes the run's own worktree name
+    # as its stem, so the tree it makes is traceable to the run the bag
+    # recorded. See `review_pr_workflow.run_review`.
     result = review_pr.run_review(
         ReviewInput(pr_number=pr, repo_target=repo_target,
                     review_type=ReviewType.PLANNING, verbose=verbose),
-        repo_root,
+        repo_root, worktree_name=worktree_name,
     )
     notes.extend(result.notes)
     return result.verdict
--- a/scripts/workflows/temporal/modules/assistant/plan/plan_project/plan_project_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/plan/plan_project/plan_project_workflow.py
@@ -171,7 +171,7 @@ def run_plan_project(*, repo_root: Path, worktree_name: str,
 
     # --- Step 3: DISPOSITION, with one bounded loop-back -------------------
     loops = 0
-    verdict = _dispose(pr, repo_root, repo_target, notes, verbose)
+    verdict = _dispose(pr, repo_root, repo_target, worktree_name, notes, verbose)
 
     while routing.should_loop_back(verdict, loops):
         loops += 1
@@ -212,7 +212,7 @@ def run_plan_project(*, repo_root: Path, worktree_name: str,
             candidates_path=candidates_path, research_dir=research_dir,
             pr_number=pr, verbose=verbose,
         )
-        verdict = _dispose(pr, repo_root, repo_target, notes, verbose)
+        verdict = _dispose(pr, repo_root, repo_target, worktree_name, notes, verbose)
 
     if verdict is routing.Verdict.HOLD_NEEDS_ASSISTANCE:
         # THE LOOP DECISION AND NOTHING ELSE. Wiring the CI gate into `_dispose`
@@ -242,6 +242,7 @@ def run_plan_project(*, repo_root: Path, worktree_name: str,
 
 
 def _dispose(pr: str, repo_root: Path, repo_target: str | None,
+             worktree_name: str,
              notes: list[str], verbose: bool) -> routing.Verdict:
     """One disposition pass, judged against the PLANNING criteria, behind the gate.
 
@@ -272,10 +273,14 @@ def _dispose(pr: str, repo_root: Path, repo_target: str | None,
     if hold is not None:
         return hold
 
+    # THREADED FROM THE RUN'S CONTEXT, NOT REBUILT. `run_review` cuts a
+    # per-pass tree on the PR's branch and takes the run's own worktree name
+    # as its stem, so the tree it makes is traceable to the run the bag
+    # recorded. See `review_pr_workflow.run_review`.
     result = review_pr.run_review(
         ReviewInput(pr_number=pr, repo_target=repo_target,
                     review_type=ReviewType.PLANNING, verbose=verbose),
-        repo_root,
+        repo_root, worktree_name=worktree_name,
     )
     notes.extend(result.notes)
     return result.verdict
--- a/scripts/workflows/temporal/modules/assistant/research/research/research_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/research/research/research_workflow.py
@@ -58,7 +58,7 @@ def run_research(*, research_dir: Path, repo_root: Path, worktree_name: str,
 
     loops = 0
     verdict = _verify_then_dispose(research_dir, pr, repo_root, worktree,
-                                   notes, verbose, correction=False)
+                                   worktree_name, notes, verbose, correction=False)
 
     # ONE loop-back. Self-correction plateaus at 3-5 passes; past it the model
     # justifies rather than corrects. Counting across the pipeline:
@@ -67,7 +67,7 @@ def run_research(*, research_dir: Path, repo_root: Path, worktree_name: str,
         loops += 1
         notes.append("HOLD (redispatch): looping back ONCE — the last automated pass.")
         verdict = _verify_then_dispose(research_dir, pr, repo_root, worktree,
-                                       notes, verbose, correction=True)
+                                       worktree_name, notes, verbose, correction=True)
 
     if verdict is Verdict.HOLD_NEEDS_ASSISTANCE:
         # THE LOOP DECISION AND NOTHING ELSE. Wiring the CI gate above gave this
@@ -88,7 +88,8 @@ def run_research(*, research_dir: Path, repo_root: Path, worktree_name: str,
 
 
 def _verify_then_dispose(research_dir: Path, pr: str, repo_root: Path,
-                         worktree: Path, notes: list[str], verbose: bool,
+                         worktree: Path, worktree_name: str,
+                         notes: list[str], verbose: bool,
                          *, correction: bool) -> Verdict:
     verify.run_verify(
         research_dir=research_dir, pr_number=pr, repo_root=repo_root,
@@ -119,9 +120,13 @@ def _verify_then_dispose(research_dir: Path, pr: str, repo_root: Path,
 
     # --type research: candidates are CARGO, not findings. A clean research PR
     # returns MERGE with zero findings, and that is the expected outcome.
+    # THREADED FROM THE RUN'S CONTEXT, NOT REBUILT. `run_review` cuts a
+    # per-pass tree on the PR's branch and takes the run's own worktree name
+    # as its stem, so the tree it makes is traceable to the run the bag
+    # recorded. See `review_pr_workflow.run_review`.
     result = review_pr.run_review(
         ReviewInput(pr_number=pr, verbose=verbose, review_type=ReviewType.RESEARCH),
-        repo_root,
+        repo_root, worktree_name=worktree_name,
     )
     notes.extend(result.notes)
     return result.verdict
--- a/scripts/workflows/temporal/modules/assistant/review_pr/review_pr_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/review_pr/review_pr_workflow.py
@@ -204,8 +204,16 @@ def run_review(task: ReviewInput, worktree: Path, *,
     # The reviewer must read the PR's branch, not the repo's checkout. The name
     # DERIVES FROM THE ONE THIS RUN WAS GIVEN — no clock, and nothing assembled
     # from values this function was not handed.
+    #
+    # ⚠ THE PR NUMBER STAYS IN THE NAME AND THAT IS NOT DECORATION. The stem is
+    # `<workflow-key>-<ts>`, so two `review-pr` dispatches against DIFFERENT PRs
+    # that start in the same wall-clock second share a stem — and on their first
+    # pass they would share the whole directory. The expression this replaced
+    # carried `pr` for exactly that reason; dropping it in the consolidation
+    # would have been a regression the run-context field's own scope-of-effect
+    # docstring predicts ("two runs collide on one directory").
     pr_tree = _shared.worktree_add(
-        worktree, f"{worktree_name}-review-{this_pass}",
+        worktree, f"{worktree_name}-review-{task.pr_number}-{this_pass}",
         f"origin/{pr['headRefName']}",
     )
     # THE NONCE BINDS THE LOG'S NAME, not just the record inside it — this
--- a/scripts/workflows/temporal/modules/assistant/review_pr/review_pr_workflow.py
+++ b/scripts/workflows/temporal/modules/assistant/review_pr/review_pr_workflow.py
@@ -133,8 +133,29 @@ def _append_shadow_pair(log_file, *, invocation_id: str, pr: str, expected_ref) 
         pass
 
 
-def run_review(task: ReviewInput, worktree: Path) -> ReviewResult:
-    """Disposition one PR and return its typed verdict."""
+def run_review(task: ReviewInput, worktree: Path, *,
+               worktree_name: str) -> ReviewResult:
+    """Disposition one PR and return its typed verdict.
+
+    `worktree_name` IS RECEIVED, NEVER ASSEMBLED HERE, and this parameter is the
+    whole of Workflow Decomposition Phase 4's reconciliation of `review-pr`.
+    This function used to build `f"review-pr-{pr}-{int(time.time())}"` inline —
+    the eleventh and oddest of the eleven sites that derived a worktree name for
+    themselves, and the one with no assignment to key a guard on. Worse, it made
+    the fleet contradict itself: `run_review_pr.py` recorded
+    `worktree_name=None` in the run bag with the comment *"the ONE workflow that
+    cuts no worktree"*, while this line cut one. Nothing was lying — the two
+    halves were written in two places and only one was updated.
+
+    THE TREE THIS CUTS IS PER-PASS, NOT PER-RUN, WHICH IS WHY THE NAME ARRIVES
+    RATHER THAN COMING OFF A CONTEXT FIELD DIRECTLY. A parent calls this once
+    per loop-back on one run, each pass needing its own tree on the PR's branch,
+    so the value cannot be a frozen boundary field — `dispatch_context.py`'s own rule
+    is that a value computed inside the work is not a context field. What IS
+    run-scoped is the STEM: the entrypoint passes `ctx.worktree_name`, a parent
+    passes the name of the tree it is working in, and the pass number — which
+    this function reads off the PR's own history — makes it unique.
+    """
     notes: list[str] = []
 
     pr = act.fetch_pr(task.pr_number, worktree)
@@ -180,9 +201,11 @@ def run_review(task: ReviewInput, worktree: Path) -> ReviewResult:
         invocation_id=invocation_id,
     )
 
-    # The reviewer must read the PR's branch, not the repo's checkout.
+    # The reviewer must read the PR's branch, not the repo's checkout. The name
+    # DERIVES FROM THE ONE THIS RUN WAS GIVEN — no clock, and nothing assembled
+    # from values this function was not handed.
     pr_tree = _shared.worktree_add(
-        worktree, f"review-pr-{task.pr_number}-{int(time.time())}",
+        worktree, f"{worktree_name}-review-{this_pass}",
         f"origin/{pr['headRefName']}",
     )
     # THE NONCE BINDS THE LOG'S NAME, not just the record inside it — this
--- a/scripts/workflows/temporal/modules/journal/journal_activities.py
+++ b/scripts/workflows/temporal/modules/journal/journal_activities.py
@@ -230,6 +230,7 @@ def _git(repo_root: Path, *args: str) -> str:
 
 def open_run_bag(*, run_id: str, writer: str | None, repo_root: Path,
                  workflow_key: str, worktree_name: str | None,
+                 journal_root: Path | None = None,
                  config_path: Path | None = None,
                  env: Mapping[str, str] | None = None) -> Bag:
     """Resolve the journal root and open this run's bag. Raises to stop the run.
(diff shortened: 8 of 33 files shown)


Pumapumapumas:
## Decision Log

**Rigour tier: safety control — 7 mutations.** Sized before testing, not after: this PR ships two guards that certify other work (a call-site checker and a field-documentation checker) and extends a third. The refine prompt requires a tool that certifies other work be run *and* made to fail at authoring time; every mutation below states its predicted red count **before** the run, and all seven hit exactly.

| # | Mutation | Predicted | Observed | Verdict |
|---|---|---|---|---|
| 1 | Re-scatter one call site (`run_plan_draft` back to a local `__import__('time')` assembly) | 1 | 1 | exact — `test_no_call_site_ASSEMBLES…` |
| 2 | Rewrite the r2 predicate to key on the `act.` alias (the realistic "match the sibling predicate" edit) | 2 | 2 | exact — the population floor **and** the detector controls |
| 3 | Blank `target`'s `scope` metadata | 2 | 2 | exact — presence check **and** the noun-phrase floor |
| 4 | Worktree name stops following the workflow key (`f"build-{ts}"` in both constructors) | 14 | 14 | exact, including the live-path demonstration |
| 5 | Drop `render()` from one `--dry-run` branch | 1 | 1 | exact — the r4 sweep |
| 6 | Echo gated on `minted` instead of `is_the_run` | 3 | 3 | exact |
| 7 | Drop the PR number from the review tree name | 1 | 1 | exact — the collision test |

Plus a **negative control on the new placement guard**: `run_plan.py` reverted to its drifted shape → predicted 1 red, observed 1, reproducing the real defect this PR found.

**One prediction missed, and it was my model of the fixture, not a weak guard.** Mutation 3's first attempt produced *no* test output. The cause was my mutation script's string splice leaving the file unparseable, so pytest failed collection rather than the assertion firing — I had predicted the guard's behaviour correctly and mis-modelled my own tooling. Re-run with a `compile()` check inside the mutation script, it produced exactly the predicted 2. The re-run is the evidence; the first result proved nothing either way.

**What each guard does NOT look at**, stated in the guard rather than here: the r2 check holds *one value at one call-site shape* and cannot see a name threaded through a data structure; the r6 check asserts each part is **present**, never that it is **true** — a field naming the wrong marker passes; the r5 demonstration covers one entrypoint's live path, with the population-wide property held by a separate sweep.

---

**[Low]** The per-pass review worktree is **not** a context field. Alternatives: (a) make it a field, as a literal reading of "every `worktree_add` call takes its name from the context" implies; (b) exempt the call site in the check with a written reason, the way `NON_STARTING_FILES` does. **Why:** a parent calls `run_review` once per loop-back, each pass cutting its own tree on the PR's branch, so the value is per-pass and a frozen boundary object cannot hold it — the plan's own test (*run-scoped, derived once at the boundary*) excludes it exactly as it excludes `paper_currency` and `due_papers`. (a) is unbuildable; (b) reintroduces the hand-kept exemption the phase was written against. The check's property is instead *received, never assembled*, which covers both shapes with no exemption list. **This is the call most worth a second opinion** — it is the one place the built thing does not match a literal reading of the requirement.

**[Low]** The bag records a **stem**, not the exact directory. Alternatives: buy the pass number before bag-open (a `gh` round trip, which moves a network call ahead of the record); return the cut name and amend the bag (a second write to a record whose whole point is being written first). **Why:** neither is worth it, no consumer reads the field programmatically, and a derivable stem-plus-suffix is a large improvement on `worktree_name=None` claiming no tree exists. Documented at both ends rather than left to be discovered.

**[Medium]** The echo is gated on `is_the_run`, not on `minted` — which contradicts the phase doc, which names `minted`. **Why:** `minted` is False for an operator retrying with `--run-id X`, who is a parent constructing its own context and the caller with the strongest claim on seeing what the retry derived. `minted` answers *did I make the NAME*; the echo asks *did I make the CONTEXT*. Surfaced rather than applied silently, and carried by a test whose name is the reason.

**[Medium]** `field(metadata=derivation(...))` rather than a field docstring. Alternatives: a `#:` comment parsed by AST; a `FIELD_DOCS` dict beside the class. **Why:** Python has no field docstrings, so a check asserting four *named* parts would parse them out of one prose blob and its population becomes the parser's guesses. A dict is a second surface that can drift from the fields. Metadata keeps the statement at the field and each part separately addressable, with `dataclasses.fields()` unchanged as the population. Reasoned in `derivation()`'s docstring, at the decision.

**[Medium]** The context is built at the boundary and its **fields** are passed down; workflow signatures were not changed to take the object itself. **Why:** the failure being closed is *a callee re-derives*, and after this change nothing does — every `worktree_add` name is received. Threading `RunContext` through ~17 workflow signatures would touch return types the suite asserts against, for no additional property. Surgical-changes rule.

**[Medium]** The journal root is resolved at the boundary; `open_run_bag` **keeps** its own resolution as a fallback. Alternatives: make the parameter required (breaks `validate_bag` and the tests); resolve in both places (two answers to one question). **Why:** one resolution per run, and the fallback is what keeps every context-less caller working. Cost written into `RunContext.build`'s docstring: `open_run_bag` stops owning a decision it has always owned.

**[High]** `run_plan.py`'s handler placement was fixed rather than surfaced, and the guard was tightened in the same PR. **Why:** the defect is pre-existing on `main`, but resolving the journal root at the boundary added a **third** raise to the unguarded region — a corrector sweeps its own blast radius, not only the defect's class. Tightening the guard is what turns the fix into something that cannot silently return.

**[High]** The planning-repo edit was **delivered as a PR** rather than declined or deferred. **Why:** the phase doc's own implementation step says *"Record the transcript in this doc"*, planning artifacts are dispatch-scope, and the checkbox-flip ruling explicitly authorises a dispatch to flip boxes for work it can show in its own diff. The mechanism is a PR in the planning repo, opened from an isolated worktree so the shared `main` checkout was never switched out from under a concurrent session. Every one of the eleven flips was verified against this PR's diff; the `Status:` line was deliberately **not** flipped, because this PR is open rather than merged and built is not proven.

## Deferred Work

- **A dispatch-boundary helper (context manager) to close the eleven-site control-flow drift structurally** — Why deferred: it changes all eleven entrypoints' exception-handling shape, a scope this phase explicitly does not take. The placement guard added here catches the drift; the helper would prevent it. Tracked at: **surfaced in the PR body under its own heading, for `review-pr` to file into `tracked/candidates/`** — I hold no write grant on any `tracked/` store and the judge is not the author. Verified by: `gh pr view 162 --json body | grep -c "dispatch boundary helper"` → the heading, consequence, remedy, intended store and `component: workflow-decomposition` are present in the body on this branch; `decision` left blank, which is `triage-candidates`' output alone.

- **The `workflow-scripts.md` § 9 standards amendment** — Why deferred: [`finding-routing.md` §7](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/standards/finding-routing.md) gives a producing run the surfacing and `review-pr` the filing, and the phase doc says so in terms: *"Surface it, do not file it, and never edit the standard."* Tracked at: **surfaced in the PR body under its own heading**, intended store `tracked/standards/`. Verified by: `gh pr view 162 --json body` → the amendment's exact proposed sentence, the section it extends, and the `guide/workflows.md` sentence it finishes are all in the body; `git -C /opt/skyy-net/skyynet-master-planning grep -n "run-scoped derived value" standards/` → no output, confirming the rule is not yet stated anywhere.

- **`test_isolation_invariants._classify` misclassifies `review_pr_workflow` as a child while it cuts a worktree** — Why deferred: reclassifying it applies a different set of parent invariants to that module, which is a behaviour change to an existing guard and outside this phase's scope. Tracked at: **surfaced in the PR body under its own heading**, intended store `tracked/issues/` (a defect in something already built). Verified by: `grep -n '"act.worktree_add("' scripts/workflows/temporal/tests/unit/test_isolation_invariants.py` → line 97, `_classify()`, live on this branch and on `main`; `grep -n '_shared.worktree_add' modules/assistant/review_pr/review_pr_workflow.py` → the call it cannot see. The two facts are one grep apart and the misclassification follows from them.

- **The phase doc's `Status:` line, and the roadmap entry** — Why deferred: this PR is open, not merged. Flipping a status on unmerged work is the "built is not proven" failure the checkbox ruling names. Tracked at: **helloskyy-io/skyynet-master-planning#4**, which states the omission and its reason in its own body. Verified by: `gh pr view 4 --repo helloskyy-io/skyynet-master-planning --json number,state,title` → `PR #4 OPEN — docs(phase4): record the wrong-derivation demonstration and the verified checkboxes`; its body carries the line *"The Status line stays `not started`. The build PR is open, not merged. Built is not proven."* This is a placement made in a PR on a non-default branch and it lands at merge.

## Post-Run Reflection

**Friction**

- **The plan's evidence table had drifted again, and the drift was load-bearing.** It named `run_plan_verify.py:321` — a file that does not exist; the real site is `run_plan_refine.py:360`. The *count* of eleven was still right, so a builder trusting the total and working from the membership would have converted ten sites and left one. That is precisely the `base_ref` eleventh-site failure the section itself predicts, reproduced a third time on the document that predicts it. Re-measuring cold cost one grep.
- **The plan did not contemplate the nested review worktree.** Requirement 2 reads as *every `worktree_add` call takes its name from the context*, which is unbuildable for the tree a parent's review pass cuts — a loop-back needs one per pass, and a frozen boundary field cannot supply that. The plan's own § *What this phase does not do* resolves it (per-pass values are excluded like `paper_currency`), but the requirement and that section have to be read against each other to see it. About forty minutes went into circling this before the exclusion clicked.
- **`run_*.py` is a namespace, not just a convention.** Naming the new module `run_context.py` made it match `directory.glob("run_*.py")` and it was swept as an entrypoint, failing five guards at once — bag-open, identity-argument declaration, identity-boundary routing, the ordering-coverage pin and a prose count. Renaming to `dispatch_context.py` fixed all five. The guards were right and loud; nothing in the tree *says* the prefix is reserved.
- **Running the unit tier alone after adding test modules is not the same as running it.** Three census guards (`_PINNED`, the entrypoint-driving-module pin, a prose figure) only fire when the whole tier is collected together, so per-file runs stayed green while the tier was red. Worth knowing that "I ran my new tests" and "the suite passes" are different claims here.

- **`run_*.py` should be documented as a reserved prefix** in `scripts/workflows/temporal/scripts/` — or, better, `journal_entrypoint_facts.entrypoints()` should refuse a file that matches the glob but declares no identity arguments *with a message naming the collision*, rather than five downstream guards each reporting a different symptom of it. The map now says the prefix is load-bearing; nothing enforces it.
- **The prose-figure sweep cannot tell which population a count is about.** `docs/file_structure.txt` saying "one of eleven" about *worktree-name sites* was read as a claim about the *entrypoint* population — which also happens to be eleven, so binding it would have been accidentally right for the wrong reason. Reworded rather than bound. The sweep's own docstring warns about exactly this ("a wrong number became accidentally right and nobody learned anything"); it may be worth letting a figure declare its subject.

- **The build-draft prompt has no review stage, but its PR-body template demands review findings** ("Review findings addressed and deferred", "Standards audit findings addressed and deferred") and its Rules say "Fix Critical review findings before submitting". Stages 1–5 are LOAD / VALIDATE / IMPLEMENT / TEST / SUBMIT. I dispatched `code-reviewer` and `quality-control` between 4 and 5 because the template requires their output — and `code-reviewer` found a Critical the whole suite had passed over. Either the stage list should name a review stage, or the template should stop asking for findings no stage produces. **This is worth the architecture session's attention: the run that most needs the review is the one whose prompt leaves it implicit.**
- **The prompt's mutation guidance says to state a predicted count and name which of two things a miss was, but does not say to verify the mutation itself applied.** My one miss was a mutation script that silently produced an unparseable file — pytest failed collection, no assertion ran, and the grep for `FAILED|passed` matched nothing, which reads on a fast skim like "nothing broke". A one-line `compile()` inside the mutation turns that into a loud failure. Suggest adding: *a mutation that does not leave the artifact loadable has not been applied, and a null result is not a green one.*

Pumapumapumas:
# build-refine pass on PR #162

## Decision Log

**Rigour tier: safety control — 12 mutations.** Sized before testing, not after: this PR ships three guards that certify other work (a call-site checker, a field-documentation checker, an entrypoint sweep) and extends a fourth. **Six mutations established the defects before anything was fixed; six are negative controls on what replaced them.** Every one stated a predicted red count before the run and **all twelve hit exactly.** The r5 demonstration was additionally reproduced from a shell against a scratch repository rather than read off the diff.

| # | Mutation | Predicted | Observed |
|---|---|---|---|
| 1 | Perturb `build()`'s worktree derivation only | 0 — the defect | **0** (3051 passed) |
| 2 | Hoist `RunContext.build` above the `try` in `run_plan.py` | 0 — the defect | **0** |
| 3 | Move `ctx.echo()` below `open_run_bag` | 0 — the defect | **0** |
| 4 | Rebind `worktree_name` inside `run_review` | 0 — the defect | **0** |
| 5 | `f"{worktree_name}-{int(time.time())}"` at a call site | detector silent — the defect | **silent** |
| 6 | `if a.verbose: ctx.echo()` | 0 — the defect | **0** |
| 7–12 | Each of the six, re-run after the fix | 5, 1, 1, 1, 1, 1 | **5, 1, 1, 1, 1, 1** |

**One thing worth recording about the method.** Mutation 1 was chosen because `code-reviewer` and `quality-control` **independently** named the same Critical — the derivation written twice with only the rehearsal tested — from different lenses. Two lenses converging is not proof, and it was not treated as proof: the mutation is what established it. Both reviewers also independently reported that four guards check *presence* where the requirement is about *placement or order*; the cross-read of the two tables is what turned that from four findings into one class, and it is the same class this PR's own Critical was about (`test_every_parent_opens_a_run_bag`'s failure message demanded placement in prose while nothing checked it). **The draft found that defect and then reproduced it in three of the four guards it wrote beside it.**

---

**[Low]** `review-pr`'s bag records a **stem** rather than the exact directory. The draft rejected the finding with reasoning; this pass **executed the rejection** rather than leaving it standing — the stem-vs-path relationship is now stated in `open_run_bag`'s own docstring, where a consumer meets the field, instead of only at the two call sites. Alternatives: buy the pass number before bag-open (moves a `gh` call ahead of the record); amend the bag after the cut (a second write to a record whose point is being first). Why: neither is worth it, and a documented derivable relationship is a large improvement on `worktree_name=None` claiming no tree exists.

**[Medium]** `_review_tree_name` keeps **scraping the f-string out of `review_pr_workflow.py`** rather than importing an extracted `_per_pass_tree_name`. Alternatives: extract the expression into a named production function (the reviewer's remedy); retype the template in the test. Why: the reviewer's brittleness complaint was real, so the *failure mode* was fixed — a readable assertion naming the file and forbidding the obvious wrong fix — but the extraction adds an indirection to production code purely for testability, and the scraping is what makes the test *derived from* the rule rather than a second copy of it. Retyping is the option this phase exists to refuse.

**[Medium]** The `RunContext.build` placement was added to the **existing** `open_run_bag` placement guard rather than given a guard of its own. Alternatives: a second sweep; a general "every boundary call is inside the handler" rule with a discovered population. Why: `_MUST_BE_GUARDED` is a two-entry tuple with a written reason per entry, and both entries raise the same exception class for the same operator-facing purpose. A discovered population is what the deleted requirement 1 tried and could not build.

**[Medium]** The three false *"before ANY `gh` call"* comments were corrected rather than the **code** being moved to make them true. Alternatives: hoist the boundary block above `base_ref` / `unclosed_hold` in those three entrypoints. Why: those reads are **refusals** — a bad `--pr`, a held PR dispatched without `--correction-pass` — and opening a bag ahead of a refusal files a run that never starts. Making a comment true by relocating a refusal is the worse trade. The property requirement 3 actually asks for (nothing is *spent or posted* before the echo) holds in all eleven.

**[Medium]** `open_run_bag`'s silent precedence of `journal_root` over `config_path`/`env` was **documented, not guarded.** Alternatives: raise `ValueError` on the combination. Why: no caller writes it, and a guard for a combination nobody writes is speculation — while a silent precedence nobody documented is exactly how a redirected test asserts against the wrong root and passes. The documentation is at the parameter, which is where the mistake would be made.

**[High]** The eleven hand-copied boundary blocks were **left in place** and the drift was made *detectable* instead. Alternatives: `open_run_bag(ctx=ctx)` now; a `dispatch_boundary` context manager now. Why: the reviewer is right that the change traded an 11× one-liner for an 11× nine-liner, and that is the surfaced candidate — but it changes all eleven entrypoints' exception-handling shape, which this phase names as out of scope, and it is **an expansion of `C-8tv8ewto`, which already owns the duplicated failure block in those same eleven files.** What this pass did instead is close the gap that made the duplication dangerous: both boundary calls are now held inside the handler, and the echo is held before the bag, across all eleven.

**[High]** The three **departures from the requirements' wording** were recorded in the planning repo, not only in code docstrings. The draft argued each at its site — which is correct and was not enough: a code docstring is not where the next planner looks, and this component has an open issue (#155) about exactly this class of pointer loss. Alternatives: leave them in code; surface them in the PR body only. Why: the phase doc's own rule is *"if a build dispatch disagrees, the remedy is to raise it rather than to add the field quietly."* Delivered as a table in helloskyy-io/skyynet-master-planning#4, which was already open for this phase.

**[High]** The `writer`-is-a-proxy limit was placed as an **EXPANSION of an existing phase-3 checkbox**, not as a new item anywhere. `RunContext.echo` gates on `writer is None`, which is exact today and stops being exact the moment a standalone child constructs its own context in a separate process — six of which [Dual-mode children](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/development/edge-assistant/workflow-decomposition/phase3_dual_mode_children.md) adds. Alternatives: a `tracked/issues/` item; a `tracked/candidates/` item; fix it now by adding a `constructed_here` field. Why: it has **no done-state today** (nothing passes `--writer`) and phase 3 **owns the trigger**, which is the first placement question; and that phase already carries a checkbox confirming its verbosity ruling against this phase's echo contract, which is the second question — an expansion, not a duplicate. Fixing it now would add a field for a caller that does not exist.

## Deferred Work

- **A dispatch-boundary helper (context manager) closing the eleven-site control-flow duplication** — Why deferred: it changes all eleven entrypoints' exception-handling shape, a scope this phase explicitly does not take. Tracked at: **an EXPANSION of `tracked/candidates/C-8tv8ewto.md`**, surfaced in the PR body with the harvest action ("increment and append; do not open a second candidate"). Verified by: `cat /opt/skyy-net/skyynet-master-planning/tracked/candidates/C-8tv8ewto.md` on the default branch → `status: open`, `count: 1`, `filed: 2026-08-26`, `component: workflow-decomposition`, body reads *"Seven consumers now carry it byte-identically"* with `parse_or_exit(parser, argv)` as the remedy; `gh issue view 149 --repo helloskyy-io/Claude-Dot-Files` → **OPEN**, *"RECURRENCE on C-8tv8ewto — the duplicated runner failure block is at ten"*, measuring `except RuntimeError` in all eleven runners. **The draft surfaced this as a NEW candidate with no pointer to either**; the pointer is the correction.

- **The `workflow-scripts.md` § 9 standards amendment** — Why deferred: `finding-routing.md` §7 gives a producing run the surfacing and `review-pr` the filing, and the phase doc says so in terms. Tracked at: **surfaced in the PR body under its own heading**, intended store `tracked/standards/`. Verified by: `git -C /opt/skyy-net/skyynet-master-planning grep -n "run-scoped derived value" standards/` → no output, confirming the rule is still stated nowhere.

- **`test_isolation_invariants._classify` misclassifies `review_pr_workflow` as a child while it cuts a worktree** — Why deferred: reclassifying it applies a different set of parametrised parent invariants to that module, a behaviour change to an existing guard in a file this PR does not otherwise touch. Tracked at: **surfaced in the PR body**, intended store `tracked/issues/`. Verified by: `grep -n '"act.worktree_add("' scripts/workflows/temporal/tests/unit/test_isolation_invariants.py` → **lines 97 and 134**, live on this branch and on `main`; `grep -n '_shared.worktree_add' modules/assistant/review_pr/review_pr_workflow.py` → **line 215**, the call those predicates cannot see. **Line numbers re-derived this pass** — the draft cited `470`/`513` in `test_every_parent_opens_a_run_bag.py`, correct when written and moved by this pass's own edits to that file; the live values are `:636` and `:684`.

- **The phase doc's `Status:` line, and the roadmap entry** — Why deferred: this PR is open, not merged. Flipping a status on unmerged work is the "built is not proven" failure the checkbox ruling names. Tracked at: **helloskyy-io/skyynet-master-planning#4**. Verified by: `gh pr view 4 --repo helloskyy-io/skyynet-master-planning` → **OPEN**, *"docs(phase4): record the wrong-derivation demonstration and the verified checkboxes"*; this pass added a second commit to it (`1eea6ba`) carrying the departures table, the evidence-table correction and the phase-3 expansion. **This is a placement made in a PR on a non-default branch and it lands at merge.**

## Post-Run Reflection

**Friction**

- **The draft's self-account was accurate everywhere it was checkable, and that is what made it misleading.** Its seven mutations were real and its predictions all hit; its `--dry-run`-vs-live claim was argued at length in three docstrings. What none of that reached was that the r4 test *could not fail* — it compared an object with a copy of its own field values — and that **nothing in the tree called `RunContext.build` at all.** The lesson for the next refine pass is narrow and worth stating: *a mutation table proves the mutations that were run, and says nothing about the ones that were not chosen.* Mutation 1 here was not on the draft's list because the draft's model of the file did not include "this expression exists twice."
- **Four guards checked presence where the requirement was placement, and the draft had just fixed that exact defect.** `test_every_parent_opens_a_run_bag`'s failure message had demanded *"inside the try block"* in prose since it was written while nothing checked it — the draft found that, fixed it, wrote about it at length, and then wrote three new guards with the same shape. **Knowing about a defect class is not the same as sweeping for it**, which is the prompt's own measured claim landing on the run that quotes it.
- **The `by_parameter` arm's hole was invisible to reading and obvious to probing.** Loading the guard module and driving `_assembled_names` on three literal shapes took under a minute and found two admitted shapes, one of which was the half-migration at the exact site the guard's docstring calls the one it most needs to see. Reading the same code twice found neither.

- **A guard that asserts a call is PRESENT should be assumed blind to where it is until proven otherwise.** This tree now has four guards that were written as presence checks for placement properties, three of them in one PR. The pattern is cheap to catch — an AST guard that collects call *names* module-wide has no notion of order or nesting, and that is visible in its helper's signature. Worth a line in the testing standard's guard section: *if the property names a position ("before", "inside", "unconditionally"), the predicate must read `lineno` or the enclosing node, not the name set.*
- **`docs/file_structure.txt` restates counts that live in test pins, and nothing binds them.** This pass found "five modules drive an entrypoint `main()`" in two places while the pin said six — a figure this PR itself moved. `test_a_prose_COUNT_of_a_collection_is_DERIVED.py` exists and does this for other figures; the entrypoint-driving population is a candidate for it.

- **The refine prompt's mutation guidance points almost entirely at inherited work, and the draft's own reflection asked for the same thing from the other side.** The two highest-value mutations in this pass were on code the *draft* shipped, and the two most valuable checks after that were negative controls on tests *this pass* wrote — one of which (the `--dry-run` early return that must NOT count as a gate) would have made the new ordering guard over-fire on eight entrypoints if it had gone unwritten. The prompt does say "mutate what you added"; it says it once, late, after four paragraphs aimed at the inherited artifact. **Worth promoting to the same prominence as the inherited-work instruction.**
- **The `build-draft` prompt still has no review stage while its PR-body template demands review findings.** The draft's reflection raised this and it is worth repeating with evidence: the draft dispatched `code-reviewer` and `quality-control` on its own initiative because the template asked for their output, and found a Critical — and then shipped a *second* Critical that two reviewers found in this pass within one dispatch each. A draft that reviews by improvisation reviews unevenly. **This is worth the architecture session's attention: the run that most needs the review is the one whose prompt leaves it implicit.**

Pumapumapumas:
# `review-pr` disposition — pass 1

**VERDICT: MERGE.** Every claim I checked against the artifact held, the suite reproduces exactly (`4256 passed, 8 skipped` / `95 passed` / `RESULT: PASSED`), the merge is clean against a `main` that moved three commits since the cut, and the three items the producing run correctly surfaced-but-did-not-place are now **filed** — #163, #164, #165 — which is the routing `finding-routing.md` §7 prescribes, not a deferral.

## What I verified independently (not read off the self-account)

The producing run's two reflection comments are unusually accurate, and accuracy is exactly what makes a self-account dangerous — so every load-bearing claim was re-derived:

| Claim | How I checked it | Result |
|---|---|---|
| The worktree name is derived **once** | `grep` for every `f"{workflow_key}-{int(...)}"` in `dispatch_context.py` | one `_worktree_name`, called by **both** constructors (`:247`, `:273`) ✅ |
| **Eleven** `worktree_add` call sites, all receiving | enumerated all non-test, non-def call sites | exactly 11, every one passes `ctx.worktree_name` or a received parameter ✅ |
| `target` names **five** targetless workflows | cross-read the metadata against which entrypoints pass `target=` | `build`, `build-minor`, `plan-project`, `review-pr`, `plan-revision` — matches ✅ |
| The echo precedes the bag in **all eleven** | per-entrypoint line-ordering sweep | `ctx.echo()` immediately above `journal.open_run_bag(` in all 11 ✅ |
| The three false *"before ANY `gh` call"* comments were corrected | read all three (`plan-refine`, `plan-sprint`, `plan-revision`) | all three now name the refusal and say what is true ✅ |
| `--dry-run` and live print the same object | flag-presence vs `for_dry_run`-presence across all 11 | 7 have both, 4 have neither — **symmetric**, no orphan ✅ |
| `DRY_RUN_RUN_ID` "may not open a bag" | **executed** `validated_run_id(DRY_RUN_RUN_ID)` | `BagError` — refused on 5 characters ✅ |
| Test totals | ran `./testing/run-all.sh` cold | `4256 passed, 8 skipped`; integration `95 passed`; `RESULT: PASSED` — **exact** ✅ |

**Asymmetric-presence sweep** (the one that usually finds something): all eleven `open_run_bag(` calls are **character-for-character identical** in their six keywords, including `journal_root=ctx.journal_root`. No entrypoint was left on the old shape. Nothing found.

**Deleted-artifact sweep:** `dispatch_identity.__all__` lost its old three-name form and correctly regained all three plus `derivation`. `run_review`'s old signature, the eleven scattered derivations, and `worktree_name=None`'s false comment are each replaced by a named destination. Nothing dropped.

**Completion-checkbox sweep: N/A —** this diff flips **zero** boxes in any planning artifact (`gh pr diff | grep '^\+.*\[x\]'` → empty). The eleven flips live in helloskyy-io/skyynet-master-planning#4, a separate PR in a separate repo, correctly left open.

**Merge safety:** `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `closingIssuesReferences: []`. `origin/main` gained three commits since the cut (`76aa824`, `89db382`, `b380b2f`); `76aa824` also touches `build_workflow.py`, so I diffed both sides — main adds `prefer_repo=slug` to `draft.run_draft`, this PR threads `worktree_name` through `_refine_then_dispose`. **Disjoint regions, no semantic conflict**; `git merge-tree --write-tree` returns clean. None of the three post-cut commits is rule-bearing, so nothing binding arrived that the author could not have read.

## Escalations

None. No live defect on the default branch needs an actor before this merges.

## Dispositions

|---|---|---|---|
| `boundary-block-duplication-doubled` | scope | **DEFERRED** | **FILED as #163, `EXPANSION of C-8tv8ewto`.** The producing run surfaced this with a pointer and no placement, which is the **correct** outcome — under `finding-routing.md` §7 no producing run holds a write grant on any store; §7 reserves them to me. Re-adjudicated rather than accepted: `C-8tv8ewto` is `status: open, count: 1`, remedy `parse_or_exit(parser, argv)`; #149 is a live unharvested `RECURRENCE` intake measuring the population at ten. **My addition is that the remedy is too small, not that the count is too low** — the eleven files now carry a *second* byte-identical block **inside** that handler, so extracting only the failure block closes the item with the duplication still present. Filed as `EXPANSION` (appends evidence, leaves `count` alone) rather than `RECURRENCE`, because #149 already owns the count increment and this is a scope change, not another occurrence. **Not held against the PR:** the phase doc's § *What this phase does not do* states at line 211 *"It does not add a new derived value. Every value in scope already exists; this phase gives them one home and one voice."* I read that line rather than the run's paraphrase of it. The drift is now **guarded** — `_MUST_BE_GUARDED = (BAG_OPEN, "build")` plus the ordering guard, both mutation-confirmed. |
| `derivation-rule-stated-nowhere` | standards-implication | **DEFERRED** | **FILED as #164, `store: standards`, `target: standards/workflow-scripts.md`, `anchor: § 9. Repo Root Operation`.** This is the PR's own **unchecked** success criterion, and it is honestly unchecked. **Consequence verified, not asserted:** the fleet has performed this consolidation *twice* at the *same eleven sites* — `base_ref` first (its docstring records the hand sweep producing "a fix applied to ten of eleven"), `worktree_name` now — and the guard shipped here **does not generalise**: its own docstring says it *"holds one value at one call-site shape"* and that a second value needs its own check. Absence confirmed **two ways**: `git grep "run-scoped derived value" origin/main -- standards/` → exit 1, and `similar-candidates.py --store standards --target … --anchor …` → four near items whose **bodies I read** (`S-30whkiii`, `S-5jo0ecfg`, `S-7ad23crm`, `S-pq2ixem4`), none about derivation placement. `ratification` deliberately unset — the operator's alone. |
| `isolation-classify-false-green` | test-gap | **DEFERRED** | **FILED as #165, `store: issues`, `repo: claude-dot-files`.** Confirmed by **execution**, not by reading: `_classify()` files `review_pr_workflow` under `CHILDREN` because both predicates key on the literal `"act.worktree_add("` while the module says `_shared.worktree_add(` — so `test_child_does_not_create_its_own_worktree[review_pr]` is green and its assertion is **actively false**. The real damage is the **exemption**: the module never reaches `test_parent_establishes_isolation`, the one assertion that would fire if it stopped cutting its per-pass tree. **Ownership tested, not assumed:** the `_shared` alias is present at the merge base `e8009db:184`, the file is not in this diff, and the AST-name mechanism this PR built serves a *different property* (received-vs-assembled), so the PR did not build this defect's fix. **I did not accept the run's remedy** — "key on the callee's AST name" is under-specified and would reclassify the module to `PARENTS`, silently dropping the two child invariants it legitimately passes. The filed item says so and adds a third finding the run did not have: `PREDICATE_CONTROLS` exercises `_creates_a_worktree` on three literals and **none is an aliased call**, so the control set shares the predicate's blind spot. Distinguished from `I-wyerraz5` by reading its enumeration — different mechanism, and `test_isolation_invariants.py` is not among its twelve. |
| `phase-doc-status-and-roadmap` | deferral | **DEFERRED** | Pointer **verified live**: helloskyy-io/skyynet-master-planning#4, **OPEN**, *"docs(phase4): record the wrong-derivation demonstration and the verified checkboxes"*. I read its diff — it touches `phase3_dual_mode_children.md` and `phase4_nothing_invisible.md` and carries the departures table. Case (b): work already in motion in a live PR. Leaving `Status:` unflipped on an unmerged PR is the *correct* application of "built is not proven", not an omission. |
| `writer-is-a-proxy-expires` | deferral | **DEFERRED** | Pointer **verified in PR #4's diff**, not in its prose: a new unchecked checkbox on `phase3_dual_mode_children.md` stating that `RunContext.echo` gates on `writer is None`, that this is exact today only because every context is constructed by the process that uses it, and that a standalone child in a separate process passed `--writer` would derive a name and a target and **say nothing**. Placement is right by both questions: it has **no done-state today** (nothing in `scripts/` passes `--writer`, verified), and **phase 3 owns the trigger**, so it is a checkbox on the phase that owns it rather than a store item — exactly what `tracked_items_standard.md` § *Placement* prescribes. |
| `three-false-gh-call-comments` | doc-drift | **FIXED** | Verified at all three sites. Each now reads *"⚠ NOT BEFORE EVERY `gh` CALL, WHICH IS WHAT THIS SAID AND WHICH IS FALSE HERE"* and names the specific refusal (`base_ref`, `unclosed_hold`, `pr_branch`) plus why moving the code would be the worse trade — opening a bag ahead of a refusal files a run that never starts. The correction ran as a **class sweep by AST across all eleven**, not at the one site reported. |
| `echo-gated-on-is_the_run-not-minted` | correctness | **REJECTED** | The run flagged this itself as a deviation from the phase doc, so it enters as a finding and gets ruled. **The deviation is correct.** `minted` answers *did I mint the NAME*; the echo's question is *did I construct the CONTEXT*. They diverge precisely on `--run-id X` — an operator retrying, who constructs their own context and is the caller with the strongest claim on seeing what the retry derived — and `minted` is **False** there. `writer is None` answers the actual question. Recorded where a planner reads it (PR #4's departures table, verified in the diff) rather than only in a docstring, and carried by a test named for the reason. |
| `per-pass-review-tree-not-a-field` | correctness | **REJECTED** | The draft called this *"the call most worth a second opinion"*, so it gets one. **The built thing is right and the literal reading of r2 is unbuildable:** a parent calls `run_review` once per loop-back, each pass cutting its own tree, and a frozen boundary object cannot hold a per-pass value. The exclusion is the phase doc's own (`paper_currency`, `due_papers`), and the chosen property — *received, never assembled* — is **strictly stronger** than the alternative: it covers the boundary sites *and* the derived one with **no exemption list**, which is the hand-kept carve-out the phase was written against. I verified both collision tests exist and that the guard rejects a rebound parameter and a call-site clock read. |
| `bag-records-a-stem` | scope | **REJECTED** | The rejection was **executed**, not merely argued: `journal_activities.py:295` now opens *"⚠ `worktree_name` MAY BE A STEM RATHER THAN A DIRECTORY THAT EXISTS"* and states the `review-pr-<ts>` → `review-pr-<ts>-review-<pr>-<pass>` relationship where a consumer meets the field. The alternatives both cost more than they buy (a `gh` round trip ahead of the record; a second write to a record whose point is being first), and a documented derivable stem is a large improvement on `worktree_name=None` claiming no tree exists. |
| `open_run_bag-silent-precedence` | correctness | **REJECTED** | Documented at `journal_activities.py:318-325` rather than guarded, and I checked the premise rather than the argument: **no caller writes the combination** — all eleven pass `journal_root=ctx.journal_root` and none passes `config_path`/`env` alongside it. A guard for a combination nobody writes is speculation; a silent precedence nobody documented is how a redirected test asserts against the wrong root and passes. The documentation sits **at the parameter**, where the mistake would be made. |
| `review_tree_name-scrapes-the-fstring` | test-gap | **REJECTED** | The reviewer's *brittleness* complaint was real and the **failure mode** is fixed — `test_dispatch_context.py:265-275` now asserts readably, names the file and the call, and explicitly forbids the obvious wrong fix (retyping the template). Extracting a production helper purely so a test can import it **moves** the second statement of the rule rather than removing it, and the scraping is what makes the test *derived from* the rule instead of a copy of it — which is this phase's own thesis. |

## Deferral integrity

`laundered_deferrals: 0 of 5`. Every pointer was opened. Three of the five ("surfaced in the PR body for `review-pr` to file") would read as laundering under a naive check — they are not. `finding-routing.md` §7 reserves every `tracked/` store to this workflow and the phase doc says so in terms (*"Surface it, do not file it, and never edit the standard"*), so **a surfaced-and-unplaced item from a run holding no write grant is the correct outcome, not a burial.** The run named the intended store, the component and the harvest action for each, and for the boundary helper it corrected the draft's missing pointer to `C-8tv8ewto` and `#149`. That is the opposite of burying. `homeless_items: 0` — all three had a home; I filed them.

All three intakes are filed against **this** repo rather than the planning repo, per the `--issues-repo` split introduced in `89db382` on `main`, whose commit message states that `review-pr` files against the repo whose PR it reviewed while the stores live in `skyynet-master-planning`.

## Reflection notes (demoted — real, but no consequence to rule on)

- `test_delegated_contract.py:38` cites `review_pr_workflow.py:19` for the `_shared` import, which is at `:29` and **was at `:29` at the merge base** — pre-existing, not caused by this PR. The claim it supports (*"the import it was written for is present today"*) is **true** and greppable; only the line number is stale. Fails the consequence gate: a reader is misdirected ten lines inside the same file. Not a finding, recorded so a later pass does not re-derive it.
- The per-pass tree reads `review-pr-<ts>-review-<pr>-<pass>` — "review" twice. Cosmetic; for every other workflow the stem is the key and the name reads correctly. Nothing breaks.

```yaml
pr_review:
  pr: 162
  run_id: a25a416434fb470cb71c4d4e61a9473e
  pass: 1
  attempt: 2
  verdict: MERGE
  converged: false
  findings:
    - id: boundary-block-duplication-doubled
      title: Extracting only the failure block would close C-8tv8ewto with the duplication it exists to remove still present in all eleven runners
      category: scope
      consequence: C-8tv8ewto's remedy (`parse_or_exit`) now covers half the duplication. Applied as scoped, the item closes while a second byte-identical block remains hand-copied in the same eleven files, and a second candidate gets opened for the remainder — the duplication the four stores exist to prevent, produced by the mechanism meant to prevent it. Dual-mode children adds nine more runners on this template.
      disposition: deferred
      pointer: https://github.com/helloskyy-io/Claude-Dot-Files/issues/163 (EXPANSION intake) -> tracked/candidates/C-8tv8ewto.md
      pointer_verified: true
    - id: derivation-rule-stated-nowhere
      title: The third run-scoped derived value will scatter across eleven sites, because the rule two consolidations already implement is written down nowhere
      consequence: The fleet has consolidated the same class twice (`base_ref`, then `worktree_name`) at the same eleven sites, discovering it by measurement both times. The guard shipped here holds ONE value at ONE call-site shape by its own docstring, so it does not generalise. A third value scatters, drifts (as `run_build_minor` already had), and costs another phase to find.
      disposition: deferred
      pointer: https://github.com/helloskyy-io/Claude-Dot-Files/issues/164 (standards intake, target standards/workflow-scripts.md, anchor "§ 9. Repo Root Operation")
      pointer_verified: true
    - id: isolation-classify-false-green
      title: review_pr_workflow is exempt from the one isolation assertion that would catch it losing its worktree, and a green test asserts the opposite of the truth
      category: test-gap
      consequence: `test_child_does_not_create_its_own_worktree[review_pr]` passes while asserting something false. Worse, the misclassification exempts the module from `test_parent_establishes_isolation` — so if review_pr_workflow ever stopped cutting its per-pass tree, every review pass would run on whichever tree it was handed and nothing in the suite would go red.
      disposition: deferred
      pointer: https://github.com/helloskyy-io/Claude-Dot-Files/issues/165 (issues intake, repo claude-dot-files)
      pointer_verified: true
      reviewed_sha: e8009db5cf4ff2c9e8c6a2f45863cf45975e807b
    - id: phase-doc-status-and-roadmap
      title: The phase doc's Status line and the roadmap entry stay unflipped until this PR merges, which is correct rather than an omission
      category: deferral
      consequence: If the pointer were dead, the phase would land with no record of the demonstration and no verified checkboxes, and the next planner would re-derive the evidence table's drift for a fourth time.
      disposition: deferred
      pointer: https://github.com/helloskyy-io/skyynet-master-planning/pull/4
      pointer_verified: true
      reviewed_sha: 23493ae
    - id: writer-is-a-proxy-expires
      title: A standalone child in its own process would derive a name and a target and say nothing, which is the invisibility this phase closes, on the runs that spend the model time
      category: deferral
      consequence: `RunContext.echo` gates on `writer is None`, exact only while every context is constructed by the process using it. Dual-mode children adds nine standalone children that construct their own context and may also carry `--writer`; those runs would go silent with a green suite.
      disposition: deferred
      pointer: helloskyy-io/skyynet-master-planning#4 — a new unchecked checkbox on phase3_dual_mode_children.md, verified in that PR's diff
      pointer_verified: true
    - id: three-false-gh-call-comments
      title: Ten entrypoints claimed the echo precedes any gh call and it was false in three, so a reader who discovers the gh read discounts the whole boundary block
      category: doc-drift
      consequence: A comment that is false at one site teaches a reader to distrust the same comment at the other ten, which is where the property actually holds.
      disposition: fixed
      remedy: fix-in-place
    - id: echo-gated-on-is_the_run-not-minted
      title: Gating the echo on `minted` would silence it for an operator retrying with --run-id, who is the caller with the strongest claim on seeing what the retry derived
      category: correctness
      consequence: If the plan's literal wording had been followed, the echo would be absent on exactly the invocation it most serves.
      disposition: rejected
      remedy: reject
    - id: per-pass-review-tree-not-a-field
      title: Making the per-pass review tree a context field is unbuildable, and the exemption-list alternative reintroduces the carve-out this phase was written against
      category: correctness
      consequence: Either alternative is worse — a frozen boundary object cannot hold a per-pass value, and a hand-kept exemption list is the drift surface the phase exists to remove.
      disposition: rejected
      remedy: reject
    - id: bag-records-a-stem
      title: The bag records a stem rather than the exact directory, and buying the pass number first would move a network call ahead of the record
      category: scope
      consequence: Undocumented, a consumer reads `Journal-Worktree` as a path that exists and fails to find it. Documented at the field, the relationship is derivable.
      disposition: rejected
      remedy: reject
    - id: open_run_bag-silent-precedence
      title: Guarding a journal_root/config_path combination no caller writes is speculation, while the undocumented precedence is how a redirected test asserts against the wrong root
      category: correctness
      consequence: Verified no caller writes the combination — all eleven pass journal_root alone. The documentation at the parameter is what a future caller would read.
      disposition: rejected
      remedy: reject
    - id: review_tree_name-scrapes-the-fstring
      title: Extracting a production helper so a test can import it moves the rule's second statement rather than removing it
      category: test-gap
      consequence: The brittleness was real and is fixed at its actual cost — a readable assertion naming the file and forbidding a retyped template. An indirection added purely for testability would re-create the two-statements defect this phase closes.
      disposition: rejected
      remedy: reject
  next_steps:
    - item: boundary-block-duplication-doubled
      kind: file-issue
      note: EXPANSION intake against C-8tv8ewto; appends evidence and the widened remedy, leaves count alone. Terminal — does not hold this PR.
      issue_url: https://github.com/helloskyy-io/Claude-Dot-Files/issues/163
      issue_repo: helloskyy-io/Claude-Dot-Files
      qualified: unrelated (phase doc line 211 scopes the extraction out; C-8tv8ewto pre-dates the branch) + substantial (changes eleven entrypoints' exception-handling shape) + not-already-covered (expands the existing item rather than duplicating it; #149 owns the count)
    - item: derivation-rule-stated-nowhere
      kind: file-issue
      note: Standards amendment to workflow-scripts.md § 9. ratification deliberately unset — operator's alone. Terminal — does not hold this PR.
      issue_url: https://github.com/helloskyy-io/Claude-Dot-Files/issues/164
      issue_repo: helloskyy-io/Claude-Dot-Files
      qualified: unrelated (the standard is in another repo and a producing run may not edit it) + substantial (a binding rule change) + not-already-covered (absence verified two ways; four near items read and rejected)
    - item: isolation-classify-false-green
      kind: file-issue
      note: Defect in an existing guard, not in this diff. Carries a corrected remedy and a third finding the producing run did not have (PREDICATE_CONTROLS shares the blind spot). Terminal — does not hold this PR.
      issue_url: https://github.com/helloskyy-io/Claude-Dot-Files/issues/165
      issue_repo: helloskyy-io/Claude-Dot-Files
      qualified: unrelated (alias present at merge base; file absent from this diff) + substantial (needs a taxonomy ruling, not a predicate swap) + not-already-covered (I-wyerraz5 is a different mechanism and its twelve-module list excludes this file)
  laundered_deferrals:
    caught: 0
    of_total: 5
  homeless_items: 0
  redispatched: false
```

## Post-Run Reflection

**Friction**

- **`similar-candidates.py` resolves `tracked/` from the process CWD, and the stores are in a different repo now.** Run from the reviewed repo's worktree it fails with *"no tracked store at &lt;worktree&gt;/tracked"*. The prompt gives the helper's absolute path and says a failure "means something real is wrong" — but the real thing is just that the four stores moved to `skyynet-master-planning` on 2026-08-31 while the reviewed repo is `Claude-Dot-Files`. Working around it means `cd`-ing to the planning repo first, which the Bash-CWD rule makes awkward inside a review. **The same split that `89db382` fixed for the harvest (`--issues-repo` vs `--repo-root`) has not reached the duplicate checker** — it needs a `--tracked-root` or the same default-resolution the harvest now has. Until then every `review-pr` run on this repo hits this on its first duplicate check.
- **The file-map reconciliation prescribed in the prompt's precheck guidance (5) does not fit this repo's map.** `docs/file_structure.txt` deliberately rolls `modules/assistant/` and `scripts/` up to the directory level with prose annotations, so a `git ls-files` set-equality check on leaf names reports ~92 false absences. The repo's own `test_file_structure_map_covers_the_tree.py` encodes the real property and is green. Worth knowing before a future runway prescribes the naive reconciliation into a repo whose map is intentionally not a full enumeration.

- **The prompt's DEFERRED case (c) and `finding-routing.md` gate 0 point in opposite directions for an item whose *evidence* is in the diff but whose *subject* is not.** The boundary-helper expansion is the live example: the eleven entrypoints are in this diff, so gate 0's "artifact appears in this change's own diff" reads as engaged, while the finding is actually about an existing candidate's remedy being too narrow. I resolved it with gate 0's own test (*"is X part of Y"*, answered against the phase doc's scope line) and I think that is right — but it took a re-read of both documents and a run under turn pressure would resolve it by whichever it read last. **Worth one sentence in the prompt: gate 0 keys on the artifact the finding is ABOUT, not the artifact the evidence came FROM.**
- **The prompt tells me to seed `attempt` from "the count of ATTESTING reflection comments", which is ambiguous when one comment attests two commits.** This PR has three commits (`wip:`, `build-draft`, `build-refine`) and two reflection comments. The prompt's parenthetical resolves the `wip:` case explicitly, which is what made it answerable — but only because that exact case was anticipated. A rule keyed on *distinct dispatch invocations* rather than on comment count would survive the next unanticipated shape.

- The producing run's own reflection asked for `run_*.py` to be documented as a reserved prefix. **That landed** — `docs/file_structure.txt` now states it is load-bearing and names the measured cost (five guards failing at once). Its second half did not: nothing yet *enforces* the prefix with a message naming the collision. Not filed — it is one line in the run's own reflection and belongs to whoever next touches `journal_entrypoint_facts.entrypoints()`.

PR #190, opened 2026-09-13
build-draft: WD Phase 6 — rule the temporal/scripts entrypoints IN, four surfaces on one root

Pumapumapumas:
## Summary

WD Phase 6, step *"Rule the `scripts/workflows/temporal/scripts/` entrypoints"*. #181 ruled the directory OUT by name *"for being unruled rather than for a property of its own"* — an explicit placeholder. This replaces it with a ruling in the producer gate (`testing/scripts/tests/unit/test_every_producer_NAMES_ITS_CONSUMER.py`, the only code file touched; `docs/file_structure.txt` gets a two-line annotation refresh).

**Four derived surfaces on the one root, one per consumer shape the directory actually holds** — measured off disk, not off the brief:

| population | members | consumer predicate | asserts on a sibling-less CI clone? |
|---|---|---|---|
| shims | 16 `*.sh` | `_documented_for_the_operator` (the bash workflows' claim, reused) | skips, like the bash-workflow surface |
| runners | 16 `run_*.py` | new `_execd_by_a_shim` — a sibling `.sh` names it on a **code** line | yes |
| libraries | 3 `.py`, no shebang | new `_imported_by_stem` — a real `import`/`from … import`, read off the **AST** | yes |
| tools | 4 `.py`, shebang | `_documented_for_the_operator` | skips |

The four member predicates **partition** the directory and that is asserted (`test_the_ENTRYPOINT_populations_PARTITION_the_directory`), so a file of a fifth shape fails rather than falling between surfaces. The shim→runner direction stays `HELD_BY` `test_shim_usage_names_itself.py` — **as two registered holders, since the refine pass** (see below); the runner surface is the **reverse** direction (every runner is exec'd by some shim), which nothing held before.

**A defect found while verifying success criterion 3 ("the ratchet can still shrink"):** `_regained` resolved every consumer path against `_REPO`, so a planning-guide page — reported as `skyynet-master-planning/guide/<page>.md` and the likeliest fix for both findings below — could never force a baselined line out. `_consumer_file` now resolves against both roots; every live derived surface asserts its consumers open; control `test_the_RATCHET_can_open_a_consumer_in_the_PLANNING_repo`. Mutation (old resolver restored) → predicted 4 failures, observed 4 — by the draft and again by the refine pass.

## Premise corrections (verified, not assumed)

[...]

PR #187, opened 2026-09-13
build-draft: a reviewer's filed intakes reach the harvest by reference — FILED-INTAKE lines close #185

Pumapumapumas:
## Summary

Closes #185 via **carrier option 1** — the child reports each filed intake, and the parent hands the reported URLs to the Phase 10 harvest. No standards change, no search API.

**The report lands on two surfaces, and the durable one was added by the correction pass.** The draft had the child print `FILED-INTAKE: <url>` on its text; review-pr pass 1 measured 28 archived review logs and found that in the current regime the child's top-level text is one 27-character block (`## Stage 1: VERIFY + GATHER`) — the sibling `VERDICT:` line reached it 0/7 — while every intake the child filed reached its **posted `pr_review:` block** 29/29. The correction pass reproduced that measurement (27 logs: 6/32 intake URLs on the text surface, `VERDICT:` 1/7 in the last seven) and added the block as the second source. Carrier 1 is unchanged: the child reports what it wrote; the parent reads its own child's report.

- **`scripts/workflows/temporal/modules/assistant/review_pr/prompts/disposition.md`** — FILING AUTHORITY: after each `gh issue create`, print `FILED-INTAKE: <url>` on its own line **and** list the same URLs under the block's `filed_intakes:`; 6a: the `VERDICT:` line stays last. Block schema (Stage 5) gains `filed_intakes:` / `- <url>` as its last two keys (`[]` when none). Funded by cutting the `run_id:` comment's history (restated by the block-ordering rule below it) and a `pass:` comment line that duplicated the line above it: **94,504 bytes**, budget ratcheted 94_633 → 94_504.
- **`review_pr_helper.py`** (pure layer) — `FILED_INTAKE_LINE` (anchored), `ISSUE_URL` (as strict as `harvest.parse_ref`'s issue arm), `FiledIntakes`, `filed_intakes()` (the printed line); **`filed_intakes_in_block()`** (THIS pass's block, the key matched at the block's top-level indent ONLY — `findings_section`'s hazard: a runway quoting the schema inside `dispatch_context: |` is not read; inline `[]` is none; an inline scalar is read and unquoted; a flow list or a `- #163` is reported as malformed), **`merge_intakes()`** (distinct URLs, first-seen, printed line first), `intake_notes(printed, in_block)` (names the count from EACH surface — the evidence the two-surface design rests on). Every parser **never raises**: the harvest runs in the parent's `finally`. `ReviewResult.issue_urls: list[str]` (empty, never None).
[...]

Issue #180, opened 2026-09-10
A caller bug is reportable as a journal outage, so an operator is told to stop the fleet for a disk problem that does not exist


Pumapumapumas:
---
store: candidates
status: open
filed_by: review-pr
repo: claude-dot-files
component: persistent-memory-protocol
---

**Surfaced by** [Claude-Dot-Files#175](https://github.com/helloskyy-io/Claude-Dot-Files/pull/175), reviewed at `9a8d6c46a1074c650d5fc2a84ef4fba16bc93562`. The producing run holds no write grant on `tracked/`; `review-pr` files.

## What it is

`modules/journal/bag.py` raises `BagError` at ~24 sites — some for I/O that failed, some for `_refuse_folded_value` refusing a malformed *value*. `emit._record_gap` catches `(OSError, BagError)` together and escalates both to `JournalUnwritable`, whose documented meaning is *the root itself is gone, and re-running writes nothing*.

So a caller bug — a `write_path` or a reason string carrying a character the tag-file round trip refuses — is reportable to an operator as a journal outage, in the one component whose thesis is that a failure must be reported as what it is. The remedies are opposite: fix the call site, versus stop the fleet.

## Why it is a proposal and not a defect held on the PR

Re-verified at the reviewed SHA: `_record_gap`'s condition is *the `incomplete` flag did not land*, which is true whichever arm raised — so the case-(d) **classification** is correct even when the cause is a caller bug, and the bag genuinely did lose data and read as complete. The message now names `failure_detail(flag_exc)`, so the cause is visible rather than buried in `__cause__`. That half was fixed on #175.

**Not reachable today, checked at all four `paired_write` call sites** rather than taken on the run's word — the run's own phrasing (*"all four call sites pass literals"*) is loose, since `assistant_activities` passes `_gh_write_path(args)` and `tracked_items` passes f-strings over store names. Both derivations are constrained: `_gh_write_path` yields `gh:<noun>:<verb>` from non-flag argv elements, and the `tracked_items` paths interpolate `Store` enum names. Neither can produce a folded value from fleet code as written. The conclusion holds; the stated reason did not.

## Proposed action

Split `BagError` into a value-refusal subclass and an I/O one, then let `_record_gap` classify: an I/O failure of the flag write is case (d); a refused value is a caller defect and should be raised as itself, loudly, without asserting the journal root is gone. ~24 raise sites in `bag.py`, plus the one handler.

`status: open`, `decision:` blank.

[...]

PR #192, opened 2026-09-14
build-draft: PMP Phase 3 — close the emit rule's five open code steps

Pumapumapumas:
## Summary

Closes the **five open code steps** of PMP Phase 3 (`phase3_the_emit_rule.md`), leaving the three run-dependent demonstrations and the integration test lifted to Phase 4 as the brief rules. No planning checkbox flipped; the sibling repo is untouched.

**1 · An unwritable journal reports on the typed exit record AND exits non-zero.** `exit_record.route()` takes a required `terminal_state` and rules on it FIRST (R-J — the first rule, ahead of R2's absence check, R0 and R1, because it reads no key off the envelope): any state but `COMPLETED` routes `undetermined` with the state as its reason. `UndeterminedReason` gains `journal_unwritable` / `emit_failed` / `store_write_failed`, spelled from `vocabulary.TerminalState` (r3, one declaration). The non-zero exit was already wired by inheritance. **This lands the field in the parent-computed stratum, not on `CHILD_SCHEMA` as the earlier derivation looked for** — `exit-protocol.md` §2.2 says the child "cannot author [facts about the process] and must not be asked to", and the journal is written by the parent process around the child, where the model cannot see it.

**2 · Both requirement-11 readers, in this change.**
- *Parent branch:* `review_pr_workflow.run_review`'s rescue path (the `except Exception as prose_failure` that otherwise rescues any failure whose typed record routed) now reads `emit.terminal_state_of(prose_failure)` and hands it to `route`, so a `JournalUnwritable`/`EmitFailed` from the child boundary is never rescued on a record that may say MERGE. This was a live hazard introduced by step 3, not a hypothetical: the transcript emit raises from inside `run_disposition`.
- *Durable line + consumer:* `assistant_activities.report_unwritable_journal` is the production caller of `gh_attempt(case_d_report=True)` — one marker-led PR comment composed from `unwritable_journal_report`. It is dispatched from the post-exit harvest (`harvest_activities`), the one activity every entrypoint runs in its `finally`, which sees the failure in flight (`emit.in_flight_journal_failure`) and holds the run's surfaces; on case (d) the harvest posts the report and writes NOTHING into the dead journal. The reporter reaches the journal package through `emit.register_case_d_reporter` (the `register_emitter` pattern, because the package may not import the assistant layer and the census forbids a `gh` mutation inside it). The **consumer** is the harvest itself: every body it reads back is checked with `unwritable_journal_in_text`, and a surface carrying the line is surfaced in `HarvestReport.as_note()` and the index (`unwritable_journal_reports`).
- `CASE_D_CHANNELS` is now `(True, True, True)`, and `test_the_case_d_CHANNEL_TABLE_matches_the_tree` derives all three from the tree (row 3 now asks `UndeterminedReason` + `route`'s signature, not `CHILD_SCHEMA`).

**3 · Every inventoried fleet-code write path emits.** The two rows the first cut left unwired: the CLI transcript (`emit_cli_transcript`, called from `run_claude` after the child exits, `unpairable_write(stop_on_failure=True)` — the case-(c) member that STOPS the run) and the run log (`_append_run_event`, case (c), the run continues, journal first then store). Inventory: 11 of 11 fleet-code paths wired; the census's inventory row and behavioural tests updated.

**4 · A deliberately retried emit appends once.** `Emitter._append` reads the writer's own `events.jsonl` back into a `(event_id, kind)` index on first use; a retried process re-derives the same identities and appends nothing twice. `test_a_RETRIED_paired_write_APPENDS_ONCE` drives two fresh emitters on one directory and asserts exactly `[intent, completion]`; a second test covers a retry after dying between intent and store write. **This reverses the prior build's ruling** ("a retried emit APPENDS AGAIN") — the state a retried process supposedly lacks is the journal itself. `dedupe_on_identity` stays as the read-side rule for a retried MEMBER (fresh subfolder). A torn line in the events file is refused as `EventsFileUnreadable` (in `APPEND_FAILURES`) rather than appended past.

**5 · Unit tests.** Event construction, edge-id mapping, the four write-failure cases and intent-without-completion were already covered (`test_journal_events.py`, `test_journal_edge_id_and_capture_filter.py`, `test_journal_emit.py`, `test_journal_root.py`); the draft added 32 tests for everything new, and the refine pass three more plus a 16-way parametrisation (one interpreter per entrypoint) of the reporter-registration gate. The `integration/` half is NOT built (lifted).

## Deviations from plan

- The exit-record field is in the **parent-computed** stratum (`UndeterminedReason`), not `CHILD_SCHEMA` — see step 1. The phase doc's checklist wording ("a field on the typed exit record's `CHILD_SCHEMA`") should be reconciled by the operator; the standard decides.
- Three derived-figure/containment guards required their declarations updated (`docs/file_structure.txt` counts, `test_every_reach_into…` prose, `test_journal_containment._TRUSTED_JOINS`) — each is the guard doing its job on a new import / join.

## Standards amendment — SURFACED, not written (`tracked/standards/`, for `review-pr` to file)

**Consequence:** `test_protocol_SS4s_reason_column_is_exactly_the_shipped_vocabulary` is RED beside the live corpus (it skips on the sibling-less runner, so CI is green) until `exit-protocol.md` §4's rule table documents the three new bins; until then an operator grouping the computed arm by reason meets bins §4 does not name. **Remedy:** add rule **R-J** to §4 as the FIRST rule — ahead of R0, R1 and R2's no-`result`-event arm, since it reads `terminal_state` and no key off the envelope — with three rows whose reason cells are `` `journal_unwritable` ``, `` `emit_failed` ``, `` `store_write_failed` `` — *"the parent's own emit boundary ended the invocation at a named terminal state; routes UNDETERMINED before any rule that reads the child's record"* — and note in §2.3 that `undetermined_reason` now carries the boundary's terminal states. The standards file is human-in-the-loop and outside this repo, so this PR does not write it. `component: persistent-memory-protocol`, `status: open`, `decision:` blank.

## Findings surfaced for the next pass

- ~~Pre-existing local red (`test_every_producer_NAMES_ITS_CONSUMER` ratchet)~~ — **landed on `main` as `4255964` while this PR was open; merged into this branch, green here.**
- **Ruff F401, 25 pre-existing findings** in files this PR does not touch (`ruff check --select F821,F401,E9 scripts/`); the CI gate selects F821 only, which is clean tree-wide. Changed files are clean on all three. A one-line ratchet candidate, not a defect.
- **Store-side idempotency is still the store's** (documented, not new): append-once makes the JOURNAL side of a retry exact; `paired_write` still calls `perform`, and `gh pr comment` has no idempotency key. Expands the existing shape in #41/#179's neighbourhood rather than a new item.
- **A journal failure in `run_claude` outranks the child's own non-zero exit** (comment in code): `EmitFailed`/`JournalUnwritable` raise before the `code != 0` branch, so the child's observed-git-state message is not printed on that path. The log is on disk. Stated rather than hidden; if a reviewer wants both messages, that is a small follow-up on the same lines.

## Refine pass (build-refine, `1f7e241`)

Reviewed against the brief and by `code-reviewer` + `quality-control`. The lenses converged on one region — **what runs before the case-(d) durable report can be posted** — and it is hardened as a unit:

- **A child that wrote no log no longer kills the parent with `FileNotFoundError`.** `run-claude.sh` creates `LOG_FILE` only when it launches the CLI; a rate-limit abort returns non-zero before that. `_read_transcript` now mirrors `from_log` — a missing file is an empty transcript, still emitted — so the `code != 0` branch keeps carrying the runner's message and the observed git state, and the resource row lands. Test: `test_a_child_that_wrote_NO_log_has_an_EMPTY_transcript_not_an_error`.
- **The durable comment passes through `capture_filter`.** It is the one store write with no preceding emit, so it was the one write the filter never saw — and its `detail` is an exception message on a public PR thread. Test: `test_the_durable_REPORT_passes_through_the_capture_filter`.
- **The harvest checks for the in-flight failure BEFORE resolving the root.** When the root is what is gone, `resolve_journal_root` replaced the run's `JournalUnwritable` and the report was never posted. On the in-flight path the bag is named from the caller's `journal_root` or not at all. Test: `test_a_ROOT_that_is_GONE_cannot_preempt_the_in_flight_report`.
- **`in_flight_journal_failure` reads `sys.exc_info()[1]`**, not `sys.exception()` (3.11+; nothing else in the fleet sets that floor, and this line runs in every entrypoint's `finally` on every run).
- **`test_every_entrypoint_REACHES_the_case_d_reporter` could not go red** — it asserted `assistant_activities` was in `sys.modules` after each import, which earlier tests in the file had already put there. Now one fresh interpreter per entrypoint (16, ~0.15 s each) asks the slot itself; 16 red with the registration line removed.

Negative controls, predicted = observed: filter removed → 1 red; missing-file guard removed → 1 red; harvest ordering reverted → 1 red; registration removed → 16 red. Draft's append-once control re-run → 2 red, as reported.

**Rejected with reasoning** (code-reviewer structure lens): folding `reporting_case_d_through` and `emitting_into` into one factory — two instances is below the threshold `code-style.md` sets, and `emitting_into` predates this PR; `assistant_activities.py`'s growth — real, pre-existing, and the producer belongs beside `gh_attempt` by the census's own rule, so a split is a separate refactor. **Accepted as-is**: `harvest_github_surfaces` returning `HarvestReport | None` — no production caller reads the return (verified: every entrypoint's `finally` discards it), and the docstring states the `None` case.

## Test results

- **CI condition (fresh `git clone`, no sibling planning repo), after the refine pass at `1f7e241`: `./testing/run-all.sh` → RESULT: PASSED — 4953 passed, 131 skipped, 0 failed.** (Draft: 4934 passed.)
- Local, sibling present, `origin/main` merged: 5862 passed, 59 skipped, **1 failed — the §4 standards gate above, expected red until ratified**. The ratchet red the draft surfaced landed on `main` (`4255964`) and is green here.
- `ruff check --select F821,F401,E9` on all 15 changed files: clean. `--select F821` tree-wide: clean.
- Mutation controls (6, predicted = observed on every one): R-J removed → 6 red; append-once removed → 2 red; in-flight check removed → 1 red; reader zeroed → 1 red; a `CASE_D_CHANNELS` row flipped → 1 red; transcript call removed from `run_claude` → 1 red.


Pumapumapumas:
## Decision Log

Rigour tier: new contract — 6 mutations (`route(terminal_state=…)` is a new routing input; the append-once index is a new property on the emit boundary; the parent branch is a control that refuses a rescue).

- **[Low]** The exit-record field lives in the PARENT-COMPUTED stratum (`UndeterminedReason` + `route(terminal_state=…)`), not on `CHILD_SCHEMA`. Alternatives: a `CHILD_SCHEMA` field, as the prior build's derivation test looked for; a separate `terminal_state` field on `ExitRecord` outside `route()`. Why: `exit-protocol.md` §2.2 forbids asking the child to author a fact about the process, and the journal is written by the parent process around the child; §2.3 says "§4 produces these; nothing else may write them", so the rule belongs inside `route()`. This is the brief's step 1 reached by a different placement than the phase doc's checklist wording names — the standard wins and the disagreement is stated in the PR body.
- **[Low]** The durable case-(d) report is dispatched from the post-exit harvest via a reporter slot the assistant layer fills at import (`emit.register_case_d_reporter`), rather than an explicit call in each of the sixteen entrypoints or a kwarg on `harvest_github_surfaces`. Alternatives: (a) a `case_d_report=` kwarg on every harvest call — sixteen copies of one constant, the drift class `preflight.refuse` was promoted to close, and it would have the journal package invoke a store-write callable it was handed; (b) a call in every entrypoint's `except` — `ctx`/`pr_url` are out of scope there; (c) posting from inside the journal package — forbidden by the census. Why: the harvest is the one activity every entrypoint already runs in its `finally` (sweep-held), so it is the only actor that both sees the failure in flight and holds the surfaces; the slot mirrors `register_emitter`'s ruled trade. Import-time registration is the weakest part and is held by `test_every_entrypoint_REACHES_the_case_d_reporter` plus a loud "no reporter registered" line on the process channel.
- **[Medium]** Three new `UndeterminedReason` members (one per non-COMPLETED terminal state), making the routing contract total over `TerminalState`, rather than one `journal_unwritable` reason for the brief's case (d) alone. Alternatives: one reason, letting `EmitFailed` route on the child's record. Why: the transcript emit raises `EmitFailed` from inside `run_disposition` — the same rescue path would swallow it and route MERGE on a run the boundary stopped; case (b)'s "the run stops at that boundary" would be false in exactly one parent. Cost: `exit-protocol.md` §4 needs three rows, surfaced in the PR body; the local standards gate is red until ratified.
- **[Medium]** Append-once is implemented as an on-disk identity index read back on first append, and `perform()` still runs on a retry. Alternatives: skip `perform` when a completion for the identity already exists (a true no-op on both sides); leave the prior "appends again, dedupe on read" ruling. Why: the brief asks for the demonstration, and the journal IS the state a retried process has; skipping `perform` was rejected because the caller's result (the `gh` reply) is not in the journal, so the store side stays the store's idempotency key. A torn line in the events file is REFUSED (`EventsFileUnreadable`, in `APPEND_FAILURES`) rather than skipped, because every reader would stop there too.
- **[Medium]** On an in-flight case (d) the harvest writes NOTHING and returns `None`, rather than attempting the harvest. Alternatives: harvest anyway (the journal may have recovered). Why: the `incomplete` flag just failed to land; a harvest that appended into that bag would produce a bag that lost data and reads as complete — the outcome Phase 1's four-state design exists to prevent.
- **[Medium]** The run log emits journal-FIRST, then writes the store, as case (c) with the run continuing. Alternatives: `paired_write` (intent/completion) since it is a call site with a store write. Why: the phase doc's inventory names `unpairable_write` for this row and the doc is the source of truth; if the store write then fails the journal has what the store does not, which is the direction the invariant permits.
- **[Low]** The transcript is read `errors="replace"`. Alternatives: strict decode. Why: a torn tail from a child killed mid-write would otherwise lose the whole transcript to protect a byte that was never a character; a replacement character is visible in the record.
- **[Low]** A journal failure in `run_claude` outranks the child's own non-zero exit (raised before the `code != 0` branch). Alternatives: emit after that branch, or chain both messages. Why: a record nobody can replay is the larger fact and the child's log is on disk; stated in a comment and surfaced in the PR body for a reviewer to rule otherwise.

## Deferred Work

- **`exit-protocol.md` §4 rows for the three new reason bins** — Why deferred: the standards file is human-in-the-loop and in the sibling repo this brief forbids touching; the mechanism is the operator's ratification. Tracked at: SURFACED in this PR's body under *Standards amendment* with the exact rows, for `review-pr` to file into `tracked/standards/` (`component: persistent-memory-protocol`). Verified by: `gh pr view 192 --json body` on this branch → the heading "Standards amendment — SURFACED, not written" is present; no issue or item exists yet, and none is claimed.
- **The pre-existing `test_every_producer_NAMES_ITS_CONSUMER` ratchet red** — Why deferred: outside the five steps; a two-line baseline deletion in a test this PR does not touch. Tracked at: none — surfaced plainly in the PR body ("Findings surfaced"), no pointer claimed. Verified by: re-measured at the branch point `772c085` before any edit — `1 failed, 5812 passed` with this exact test — so it is not this PR's.

## Post-Run Reflection

- **Friction:** The brief's step 1 and the phase doc's checklist both say "a field on the typed exit record" but the prior build's derivation had bound that to `CHILD_SCHEMA`, and the standard governing the record (§2.2) rules that stratum out; the run had to settle which of three artifacts wins (the standard) before it could build. The phase doc's `unpairable_write` row for the run log and the "appends again" ruling on retries both had to be re-read against the brief, which reverses the second — worth one sentence in a brief that reverses a documented ruling ("the earlier ruling that X is superseded"). `sys.exception()` inside a `finally`-invoked activity is documented behaviour but not obvious; verified empirically before use.
- **Project-level suggestions (this repo):** The local-only red on a corpus-dependent standards gate is the correct outcome of "a build may not write a standard", but nothing marks that test as *expected red pending ratification* — an operator running `run-all.sh` beside the corpus sees a FAIL with no pointer to the PR that surfaced the amendment. A convention for such gates (an `xfail` keyed on an open candidate, or the gate printing "surfaced at <PR>") would separate "the code drifted" from "the standard is behind the code, and here is where that is filed".
- **Tooling-level suggestions (claude-dot-files):** The build prompt's "verify the task's asserted facts" rule caught the `CHILD_SCHEMA` premise; what it does not say is which artifact wins when a *prior build's test* and a *standard* disagree about a placement. The prompt's "the spec wins" clause is about the brief vs a spec; a third line — "a test written by a prior pass is a claim, not a spec; check it against the standard it cites" — would have saved the deliberation here.

Pumapumapumas:
## Decision Log

Rigour tier: new contract — 5 mutations (4 negative controls on this pass's own additions, 1 re-run of the draft's append-once control). The touched surfaces are a routing input, a safety-reporting path and a new gate, so discrimination was owed on what this pass ADDED; the draft's six controls were re-run only where an edit of mine touched the guarded line.

- **[Low]** `sys.exception()` → `sys.exc_info()[1]` in `in_flight_journal_failure`. Alternatives: leave it and declare a 3.11 floor for the fleet. Why: identical answer inside a `finally`; the call runs in every entrypoint's `finally` on every run, nothing else in the fleet requires 3.11, and a 3.10 host would have broken every workflow at the one line meant to report case (d). Cost: none. A floor is an operator decision and this pass has no brief to set one.
- **[Medium]** A child that wrote no log gets an EMPTY transcript that is still emitted, rather than no emit. Alternatives: skip the emit when the file is absent. Why: `emit_cli_transcript` per `run_claude` call keeps one invariant (one transcript event per invocation) that a reader can count on; an absent event is indistinguishable from a forgotten emit, which is the shape Phase 4 hunts. The defect itself was not optional — `run-claude.sh` returns before creating `LOG_FILE` on a rate-limit abort, a measured operational condition — so a rate-limited dispatch died with "No such file" and lost the runner's message and its resource row.
- **[Medium]** The durable case-(d) comment passes through `capture_filter` before `gh_attempt`. Alternatives: leave it unfiltered (the realistic content is a path and an errno). Why: requirement 10's argument is stronger for a public PR comment than for the journal — there is no redaction event for a comment — and the cost is one line. quality-control raised it as a coarse shape and asked for a `security-auditor` pass; the fix is smaller than the audit, so it was applied and the audit not dispatched.
- **[Medium]** The harvest's in-flight check moved ahead of `resolve_journal_root`, with the bag named from the caller's `journal_root` or not at all. Alternatives: code-reviewer's literal remedy (move the read to the top and leave the derivations) — which changes no behaviour, since the derivations still run first. Why: the case that matters is the root being GONE, and on that path `resolve_journal_root` raised and replaced the run's failure; the test added drives exactly that. `origin_remote` returns `""` rather than raising and `run_id` was accepted by the bag that opened, so the two derivations left ahead of the check cannot preempt it — stated in the code.
- **[Medium]** `test_every_entrypoint_REACHES_the_case_d_reporter` re-written as one subprocess per entrypoint (16 × ~0.15 s) rather than a static import closure. Alternatives: an AST walk over relative imports across `modules/`. Why: the property is "a run of THIS entrypoint alone fills the slot", and only a fresh interpreter can answer it without re-implementing the import system; ~3 s is within the suite's budget. The original could not go red — earlier tests in the file had already imported `assistant_activities` — which its own comment half-admitted.
- **[Low]** Rejected the structure-lens suggestion to fold `reporting_case_d_through` and `emitting_into` into one factory. Why: two instances, one pre-existing; `code-style.md` puts the abstraction threshold above that, and refactoring `emitting_into` is outside the brief.
- **[Low]** The §4 standards-gate red stays RULING-REQUIRED, not fixed here. Why: the brief forbids touching the sibling repo and the standard is human-in-the-loop; the amendment is surfaced in the body with the exact rows. This pass sharpened the placement wording — R-J is the FIRST rule (before R2's absence arm, R0 and R1), not merely "ahead of R1".

**Class sweep (EXHAUSTIVE over `modules/`):** the F1 shape — a read of the child's log assuming it exists — has 4 sites (`from_log`, `_log_events`, `_append_run_event`'s append, `_read_transcript`); 3 were already guarded, the fourth is now. Blast radius of my edits: the containment guard's `_TRUSTED_JOINS` prose for `harvest_activities.py` re-read against the now-two joins and updated; the census's "twenty-three call sites" figure is unaffected (a `from ..journal.capture_filter import filter_capture` binds a function, not a submodule) — confirmed by the boundary suite staying green.

## Deferred Work

- **`exit-protocol.md` §4 rows for `journal_unwritable` / `emit_failed` / `store_write_failed` (rule R-J, first)** — Why deferred: RULING-REQUIRED — the standards file is human-in-the-loop and lives in the sibling repo this brief forbids touching. Tracked at: SURFACED in this PR's body under *Standards amendment — SURFACED, not written*, with the exact rows, for `review-pr` to file into `tracked/standards/` (`component: persistent-memory-protocol`, `decision` blank). Verified by: `gh pr view 192 --json body` after this pass's edit → the heading is present and the remedy now names the rule's position as FIRST; `ls /opt/skyy-net/skyynet-master-planning/tracked/standards/ | grep -i journal` → nothing filed yet, and none is claimed. Until ratified, `test_protocol_SS4s_reason_column_is_exactly_the_shipped_vocabulary` is red beside the corpus (skips on CI).

## Post-Run Reflection

- **Friction:** The draft's mutation claims all reproduced, and its self-account was reliable on every MEASURED item; what it got wrong were two cross-file sentences — "`from_log` has already raised" (it returns `(None, None)`) and a test whose own comment described why it could not fail. Both are the "prose in a file readers trust" class the workflow prompt names, and both took execution to confirm, not reading. `gh pr view --json headRefOid` returned the OLD head for a few seconds after `git push` succeeded — a reader treating that as "another dispatch pushed" would be wrong; the push log is the authority.
- **Project-level suggestions (this repo):** The fleet has no declared Python floor. `sys.exception()` would have set one silently at 3.11; `run-all.sh` and `tests.yml` pin 3.13 but the shims run whatever `python3` resolves on the host. A one-line `requires-python`, or a guard test asserting no 3.11+-only names appear in `modules/`, would make the floor a decision rather than an accident.
- **Tooling-level suggestions (claude-dot-files):** The prompt's "read the skip population one by one" is sound but `run-all.sh`'s log does not carry `-rs`, so the skip reasons had to be re-derived with a second pytest invocation in the CI-condition clone. Passing `-rs` through the suite runner (or printing a skip-reason histogram in the summary) would make that check one read instead of one run.

Pumapumapumas:
## review-pr · pass 1 · PR #192 — disposition

**Reviewed:** head `1f7e24189b76f47d635256284dad66d17585d25a` (4 commits; merge `e2c65d8` resolved no conflict hunks — `--cc` diff empty, nothing discarded). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `closingIssuesReferences: []`. Worktree repo = PR repo. Merge-base is `origin/main` tip (`4255964`), so no rule arrived after the cut. Body is about THIS PR — every artifact it names is in the diff. Reflection coverage is per-commit: draft `1b78ff8` and refine `1f7e241` each attested (the `wip:` checkpoint is part of the draft attempt).

**Angle of this pass (pass 1, so the full sweep):** re-executed the producing run's own measurements rather than re-reviewing the code — the R-J mutation control (6 red + the standing gate red = 7, matches), the in-flight detection inside a `finally` whose `except` is outside (returns the exception; `None` after the handler and outside), `terminal_state_of` over all three boundary classes plus a foreign exception, the two derivations ahead of the in-flight check on a repo with no `origin` (`''` / `None` / `HarvestError`, none raises), the torn-tail behaviour on both write paths, the touched test files (366 passed, 1 failed — the §4 gate, as reported), ruff on all changed files (clean), and the standard itself at sibling HEAD `dbef86e` (unamended; §2.3 line 156 rules the placement).

**Sweeps:** deleted-artifact — one test deleted (`test_a_RETRIED_paired_write_appends_a_deduplicable_pair_per_attempt`); its three assertions map §1 "appends again" → deliberately reversed by `test_a_RETRIED_paired_write_APPENDS_ONCE`, §2 `applied_intents == 1` → same test line 279, §3 completion-only shape → `test_a_retry_that_DIED_between_intent_and_store_write_completes_the_pair`. Nothing lost. Asymmetric-presence — the only broad rescue around `run_claude` in `modules/` is `review_pr_workflow` (grep of `except Exception` / `except RuntimeError`); the two `RuntimeError` handlers at lines 551/799 wrap read-only `gh` calls that emit nothing; every other parent propagates to the entrypoint's handler (the process channel). Completion-checkbox — 0 flips in this repo; the phase doc lives in the sibling and was not touched. Inbound-citation — no markdown heading/anchor changed; `docs/file_structure.txt` carries annotations, not counts, and the "twenty-three call sites" figure it now states is the one the boundary suite derives (green). Predecessor deferrals — #175's two code deferrals (the exit-record field; the two unwired paths) are exactly what this PR closes; #175's BagError proposal is filed as #180 and its premise ("all call sites pass literals") still holds — the one derived `write_path` this PR adds (`run-log:{event_type}`) is bounded to three private literal callers.

### Verified, no finding (the run's account matched the artifact)
- Parent-computed stratum, not `CHILD_SCHEMA` — the standard decides it (§2.3 line 156: *"computed — the parent, and only the parent … not in the child's schema at all"*).
- All three `CASE_D_CHANNELS` rows True and derived from the tree; the reporter slot is filled at `assistant_activities` import; 16 entrypoints call the harvest in a `finally`.
- Append-once index keyed on `(event_id, kind)` — same key as `dedupe_on_identity`, so the reader-visible outcome is unchanged from before; `perform()` still runs on a retry (documented in code and in the phase doc's r7 table).
- `_read_transcript` missing-file → `""`, still emitted; durable comment passes `filter_capture`; harvest checks in-flight BEFORE resolving the root.
- The draft's "pre-existing ratchet red" landed on `main` as `4255964`, merged into this branch, green here.
- ruff F401: 27 tree-wide today (run said 25), all in files this PR does not touch; changed files clean on F821/F401/E9. No consequence → reflection note, not a finding.

### Disposition table

|---|---|---|---|
| `exit-protocol-s4-lacks-the-three-boundary-bins` | standards-implication | **HOLD** — needs-assistance | The PR ships three `undetermined_reason` bins that `exit-protocol.md` §4 does not name; the vocabulary gate is RED beside the corpus (re-run: `1 failed`, *shipped but not documented: emit_failed, journal_unwritable, store_write_failed*) and only skips on the sibling-less CI runner. Merging puts a locally-red gate on `main` for every operator. The amendment is now a committed proposal: intake **helloskyy-io/skyynet-master-planning#32** (`store: standards`, target `standards/workflows/exit-protocol.md`, anchor §4 — checked `similar-candidates.py` and the queue: S-7ad23crm targets §2.3, not a recurrence). Ratification is the operator's; remedy `ratify-standard-change`. |
| `phase3-doc-describes-superseded-state` | doc-drift | **HOLD** — needs-assistance | `phase3_the_emit_rule.md` at sibling `dbef86e` still says *20 of 25*, *five code steps stay open*, two inventory rows `⚠ NOT WIRED`, *9 of 11 wired*, and five `[ ]` this PR's diff earns (lines 369, 381, 385, 391, 393). The next PMP dispatch (Phase 4) sequences off that. The brief forbade the sibling repo, and `main` shows the operator reconciles post-merge (`docs(pmp): reconcile phase-10 with #187`, `docs(wd): reconcile phase-6 with #190`) — so this is that reconcile, enumerated line-by-line below so it is one edit, not an investigation. Remedy `extend-upstream-artifact`. |
| `retry-appends-once-reverses-phase3-ruling` | correctness | **REJECTED** (ruled in the PR's favour) | Phase doc line 392 left the box unchecked *"so a reviewer rules on the divergence"*. Ruled: the prior ruling's premise — *"suppressing the append would require the emitter to remember what it wrote, which a retried process does not have"* — is false; the retried process holds the writer's own `events.jsonl`, and `_appended_pairs` reads exactly that. Verified: two fresh emitters on one directory append `[intent, completion]` once (`test_a_RETRIED_paired_write_APPENDS_ONCE`); replay still applies once. The doc line rides the reconcile above. |
| `phase3-checklist-names-child-schema-stratum` | doc-drift | **REJECTED** | The run's open question ("the standard decides"). It does: §2.3 line 156. The checklist wording at line 370 is what is wrong, not the code; it rides the reconcile above. |
| `store-side-duplicate-on-retry-is-unjournaled` | deferral | **REJECTED** | The run's "documented, not new" item. Verified documented: phase doc r7 table row *"The activity is retried … the store gets a duplicate comment while the journal correctly records one"* — a ruled design limit, unchanged by this PR (before: appended twice, deduped on read; now: appended once; the store side identical either way). The run's pointer "#41/#179's neighbourhood" is loose — #179 is the concurrent-writer identity mechanism, a different one — but nothing is deferred, so nothing needs a pointer. |
| `journal-failure-outranks-child-exit-message` | correctness | **REJECTED** | On the path where the transcript emit fails AND the child exited non-zero, only the `EmitFailed`/`JournalUnwritable` message reaches the operator. Ruled correct precedence: the journal failure is the root cause on that path (ENOSPC is the scenario the code names) and its message carries the path and the errno; the child's exit code and stream are in the log file on disk, and the observed git state is the worktree, still there. Chaining both would put the symptom beside the cause at the cost of the one-exception contract every entrypoint handler is written against. Stated in code; no change. |
| `torn-tail-makes-writer-unretryable` | correctness | **REJECTED** (reviewer's own, blind-spot sweep) | Probed: a torn last line makes the next `paired_write` on that writer raise `EmitFailed ← EventsFileUnreadable` (run stops, case b); the `unpairable_write` path lands the `incomplete` flag but not the gap event (both verified). So a retry after a kill-mid-write cannot proceed without inspection. Not a defect of this PR: it is LOUD, loses nothing silently, and is strictly better than before — appending past a tail with no newline would have concatenated the new event onto the torn line. The alternative (WAL-style tail truncation) mutates an append-only file and is a Phase 4 recovery-design question; noted below for the operator's reconcile. |

**Verdict rationale:** two items need a human — the §4 ratification (the gate is red beside the corpus until it lands) and the sibling-repo reconcile that the brief reserved to the operator. Nothing in the code needs a redispatch.

### WHAT HAPPENS NEXT (runway)

1. **`exit-protocol-s4-lacks-the-three-boundary-bins`** — needs-assistance (`standards`).
   `reframe:` not "should R-J exist?" (the standard's own §2.3 already puts the fact in the parent's stratum) but "which row and where?" — the intake carries the row text.
   `bp:` a contract's spec and implementation ship together; where the spec is human-gated, the spec lands first and the gate goes green before the implementation merges (never merge a known-red gate).
   **Recommendation:** ratify #32 — add R-J as the first row of §4, fix the "ten rules" count, note the enum widening in §2.3 — in the sibling repo; then beside the corpus: `python3 -m pytest scripts/workflows/temporal/tests/unit/test_exit_record.py -q -k SS4s_reason_column` → `1 passed`. That is the whole of this item.

2. **`phase3-doc-describes-superseded-state`** — needs-assistance (`planning-gap`).
   `reframe:` not "does the phase doc need updating?" but "which lines does #192's diff earn, so the reconcile is one edit?"
   `bp:` a planning artifact records what is DEMONSTRATED; a checkbox flips against the diff that satisfies it, never against the run's account.
   **Recommendation** — `development/edge-assistant/persistent-memory-protocol/phase3_the_emit_rule.md`, at merge, in the pattern of the #187/#190 reconciles:
   - line 3 status: 20 of 25 → 24 of 25 built (the integration-test half of line 393 stays lifted), and drop *"Five code steps stay open …"*.
   - lines 334–335: remove `⚠ NOT WIRED IN THIS PHASE`; line 337's residual paragraph → wired in #192 (`emit_cli_transcript`, `stop_on_failure=True`; `_append_run_event`, journal-first case (c)).
   - line 369 `[x]`; line 370 → replace the `CHILD_SCHEMA` bullet: the field is parent-computed (`UndeterminedReason` + `route(terminal_state=…)`, rule R-J), per exit-protocol §2.3; note the §4 rows are pending ratification (#32).
   - line 381 `[x]` — parent branch: `review_pr_workflow.run_review`'s rescue path; durable consumer: the harvest (`unwritable_journal_reports` in `as_note` and the index).
   - line 385 `[x]` — 11 of 11 fleet-code paths; line 410's measurement row 9 → 11.
   - line 391 `[x]`; line 392 → *"RULED at review of #192: a retried emit appends once — the journal IS the state a retried process has (`Emitter._appended_pairs`); `dedupe_on_identity` stays as the read-side rule for a retried MEMBER (fresh subfolder)."*
   - line 393: `unit/` half `[x]`, `integration/` half stays lifted to Phase 4.
   - optional, Phase 4's doc: one line under its recovery design — *what a retry does with a torn `events.jsonl` tail* (today: refuses loudly, `EventsFileUnreadable`).

Once 1 lands (gate green beside the corpus) and 2 is applied at merge, the next pass is MERGE.

### Post-Run Reflection
- **Friction:** none blocking. The `--cc` check on the merge commit, the sibling-repo current-tree check and the predecessor-deferral sweep were each one command; the run's self-account was reliable on every measured claim.
- **Project-level notes (this repo, not findings):** ruff F401 is 27 tree-wide in untouched files (top: `test_isolation_invariants.py` ×4); the CI gate selects F821 only — a ratchet is a one-line candidate if wanted. The fleet has no declared Python floor; `sys.exc_info()` was chosen over `sys.exception()` for that reason. #180's premise ("all call sites pass literals") now reads "five literal, one derived from a closed set of three private literals" — still not live; a triage note, not a recurrence.
- **Tooling-level suggestions (claude-dot-files, for the architecture session):** (1) A corpus-dependent standards gate that goes red because a build correctly declined to write a standard has no way to say *expected red pending ratification, filed at <intake>* — an operator running `run-all.sh` beside the corpus sees FAIL with no pointer. A convention (the gate printing the intake it is waiting on, keyed on the open `tracked-intake` issue) would separate "the code drifted" from "the standard is behind the code, and here is where that is filed". (2) The review-pr prompt's disposition enum lists `dissolved` but gate 0 bars it for work-in-hand findings, so a reviewer's own blind-spot finding that a lens collapses (the torn-tail item here) has to be recorded as `rejected`, which reads as "should not have been raised". Either admit `dissolved` for reviewer-raised items or say so in the enum note.

```yaml
pr_review:
  pr: 192
  run_id: 0c30160df65e44f388376c437bfa1f45
  pass: 1
  attempt: 2
  verdict: HOLD
  converged: false
  findings:
    - id: exit-protocol-s4-lacks-the-three-boundary-bins
      title: the vocabulary gate is red beside the corpus and three routing bins have no row an operator can look up until exit-protocol §4 carries rule R-J
      consequence: merging puts a locally-red test on main for every operator running the suite beside the corpus, and an operator grouping the computed arm by reason meets bins the standard does not name
      disposition: hold
      hold_kind: needs-assistance
      reviewed_sha: dbef86e9e261cf80d636a6fd3331e4b2fc52175e
    - id: phase3-doc-describes-superseded-state
      title: the next PMP dispatch sequences off a phase doc that still says five code steps are open, two paths unwired and a retry appends again
      category: doc-drift
      consequence: Phase 4 re-plans work #192 already did and builds replay assumptions on a reversed ruling
      disposition: hold
      remedy: extend-upstream-artifact
      hold_kind: needs-assistance
      reviewed_sha: dbef86e9e261cf80d636a6fd3331e4b2fc52175e
    - id: retry-appends-once-reverses-phase3-ruling
      title: the PR reverses the phase doc's "a retried emit APPENDS AGAIN" ruling, which the doc reserved for a reviewer to rule on
      category: correctness
      consequence: if the reversal were wrong, a retry would silently drop a legitimately distinct event; verified it is not — the index keys on the same identity the read-side rule already keyed on
      disposition: rejected
      remedy: reject
    - id: phase3-checklist-names-child-schema-stratum
      title: the phase doc's checklist places the exit-record field on CHILD_SCHEMA while the standard places it in the parent's stratum
      category: doc-drift
      consequence: a reader of the checklist would look for a field that must not exist there
      disposition: rejected
      remedy: reject
    - id: store-side-duplicate-on-retry-is-unjournaled
      title: a retried activity re-performs its store write while the journal records one — the run's own "documented, not new" item
      category: deferral
      consequence: none new — a ruled design limit in the phase doc's r7 table, and the read-side outcome is identical to the pre-PR behaviour
      disposition: rejected
      remedy: reject
    - id: journal-failure-outranks-child-exit-message
      title: on a double failure the child's observed-git-state message is not printed, only the journal failure
      category: correctness
      consequence: an operator on that path reads the root cause (the journal failure with path and errno) and finds the child's exit in the log on disk and the git state in the worktree
      disposition: rejected
      remedy: reject
    - id: torn-tail-makes-writer-unretryable
      title: a writer whose events file has a torn last line refuses every later append, so a retry after a kill-mid-write stops at case (b) until inspected
      category: correctness
      consequence: the retry cannot self-recover from the one failure retries exist for; it is loud, loses nothing silently, and is strictly safer than the pre-PR concatenation onto the torn line
      disposition: rejected
      remedy: reject
  next_steps:
    - item: exit-protocol-s4-lacks-the-three-boundary-bins
      kind: needs-assistance
      note: ratify intake skyynet-master-planning#32 (R-J first row of §4, count sentence, §2.3 note); gate goes green beside the corpus
      why_human: standards
      reframe: not "should R-J exist" — §2.3 already puts the fact in the parent's stratum — but "which row, where"; the intake carries the row text
      bp: spec and implementation ship together; where the spec is human-gated it lands first and the gate is green before the implementation merges
      recommendation: |
        Apply #32 in skyynet-master-planning: add R-J as the FIRST row of the §4 table
        (condition: terminal_state != COMPLETED; routed_outcome undetermined; reason = the
        terminal state by value — journal_unwritable / emit_failed / store_write_failed),
        update the "ten rules" count sentence, and note in §2.3's undetermined_reason row
        that the enum now carries the emit boundary's terminal states. Then beside the
        corpus: python3 -m pytest scripts/workflows/temporal/tests/unit/test_exit_record.py
        -q -k SS4s_reason_column -> 1 passed.
    - item: phase3-doc-describes-superseded-state
      kind: needs-assistance
      note: reconcile phase3_the_emit_rule.md against #192's diff at merge — lines enumerated in the runway above
      why_human: planning-gap
      reframe: not "does the doc need updating" but "which lines does this diff earn, so the reconcile is one edit"
      bp: a checkbox flips against the diff that satisfies it; a planning artifact records what is demonstrated
      recommendation: |
        At merge, in the pattern of the #187/#190 reconcile commits: status line 3 (24 of 25;
        drop "five code steps stay open"), inventory rows 334-335 + residual 337 (wired),
        checklist 369 [x] + 370 rewritten to the parent-computed placement per exit-protocol
        §2.3 with a pointer to intake #32, 381 [x], 385 [x] + measurement row 410 (11 of 11),
        391 [x] + 392 rewritten as the ruled reversal (the journal IS the retried process's
        state), 393 unit half [x] with the integration half still lifted to Phase 4.
        Optional: one line in Phase 4's recovery design on what a retry does with a torn
        events.jsonl tail.
  laundered_deferrals:
    caught: 0
    of_total: 2
  homeless_items: 0
  redispatched: false
  filed_intakes:
    - https://github.com/helloskyy-io/skyynet-master-planning/issues/32
```

Issue #179, opened 2026-09-10
Two concurrent writers in one run derive one event id, so replay silently drops one of their completions


Pumapumapumas:
---
store: candidates
status: open
filed_by: review-pr
repo: claude-dot-files
component: persistent-memory-protocol
---

**Surfaced by** [Claude-Dot-Files#175](https://github.com/helloskyy-io/Claude-Dot-Files/pull/175), reviewed at `9a8d6c46a1074c650d5fc2a84ef4fba16bc93562`. The producing run holds no write grant on `tracked/`; `review-pr` files.

## What it is

`modules/journal/events.py::event_identity` derives an event's id from `(run_id, write_path, sequence)`. `sequence` is per-`Emitter` (`Emitter._sequences`), so two concurrent writers in one run emitting on the same `write_path` derive the **same** `event_id`. `dedupe_on_identity` keys on `(event_id, kind)` and keeps the first, so on replay the second writer's distinct, successfully-written completion is **dropped** — silent data loss, which is the class cases (b) and (c) exist to forbid, arriving through the field that defends against retries.

Unreachable under today's wiring: parallel children are read-only critics and a single analyst writes, so no second writer performs a store write. That is a property of the wiring, not an invariant anything holds. The producing run states the consequence and the trigger in the function's docstring.

## Why this is a proposal and not an edit — and why the run's stated reason is NOT the one that holds

The producing run deferred it because *"it is a schema change to a version-1 field, and a field's material is exactly what requirement 8 forbids changing afterwards."* **Re-checked against the code: requirement 8 as this package states it is *version every event, never change a written event*** (`events.py:374`, `capture_filter.py:17`, `emit.py:449`). It forbids mutating an event already on disk; it mandates versioning precisely so the schema *can* evolve. So r8 is not the obstacle.

The obstacle that **is** real is a two-horned design trade with no free answer, and it is why this needs a ruling rather than a one-line patch:

- keying on the **requested** writer name is stable across a retry, but two concurrently-dispatched members that ask for the same name still collide — so it does not close the defect;
- keying on the **allocated** writer directory closes the defect, but `journal_activities` documents that *"a retry of one member gets a new subfolder (`w`, then `w-2`), because `writer_dir` allocates rather than adopts"* — so it breaks `paired_write`'s stated idempotency property that *"a retry is then a no-op on BOTH sides"*, because the store's idempotency key derives from the same identity.

## Proposed action

[...]

PR #181, opened 2026-09-10
build-draft: Phase 6 — the producer gate reaches the fleet; every directory is ruled, seven surfaces derive their consumer

Pumapumapumas:
## Summary

Phase 6 — [Every producer names its consumer](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md) — had two implementation steps open after #170: re-count `scripts/helpers/` off disk and rule each tool, and **enumerate the remaining candidate producer surfaces across the fleet and rule each in or out**. This PR closes both, and the refine pass closed what the sweep had left unruled.

- **Every tracked directory in the repo is now RULED.** `test_every_DIRECTORY_in_the_repo_is_RULED` walks `git ls-files` and fails on any directory that is neither a ruled-in surface root, a `RULED_OUT` entry with its reason, a `CONTAINER` whose own direct files are each named with a reader (data, asserted against `git ls-files` — see the correction pass below), nor a `tests/` directory. The phase's own hand-kept count was wrong three times in a fortnight; the enumeration is now a check. Demonstrated twice: an untracked `config/newthing/` turns the suite red naming it.
- **Seven surfaces ruled IN with the consumer FOUND on disk** (a "derived" surface — no README table is possible for `config/agents/` or `config/hooks/`, which `install.sh` symlinks wholesale into `~/.claude/`; a README there would load as an agent named `README`). Each has a predicate written for its own invocation shape, and each predicate has a self-contained control:

  | Surface | Consumer is… | Result |
  |---|---|---|
  | `config/hooks/` | named in `config/settings.json` | 3/3 wired |
  | `config/agents/` | named by a dispatching surface (commands, skills, rules, workflow tree; never `config/agents/` itself, never `tests/`) | 12/12 named (`test-writer` and `workflow-analyst` only by the two bash CPI workflows; the mentions/invokes gap is recorded in the docstring) |
  | `scripts/workflows/activities/` | named on a non-comment line of a bash workflow or sibling library | **1/4 — three unread** |
  | `scripts/workflows/common/` | same | 3/3 |
  | `scripts/services/` | read by `install.sh` via `$SERVICES_DIR/<name>` — a bare mention is not enough | **2/3 — one unread** |
  | `testing/suites/` | its stem is in `run-all.sh`'s `FRAMEWORKS=(…)` (the filename never appears literally) | 1/1 |
[...]

PR #170, opened 2026-09-10
build-draft: every producer names its consumer — the gate reaches three surfaces

Pumapumapumas:
Implements [Workflow Decomposition Phase 6 — *Every producer names its consumer*](https://github.com/helloskyy-io/skyynet-master-planning/blob/main/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md).

A surface written by one part of the system and read by no other part is not neutral: it costs the run that produces it, it looks like coverage to a reader, and nothing goes red when it stops being correct. A gate existed for exactly one directory. This rules the definition, extends the gate to two more surfaces, and proves it by breaking it on purpose.

## What was built

**The definition (requirement 1)** — stated once, in the module the check is built from:

> A **producer** is a surface one part of the system writes for another part to read. It is conformant when its consumer is **named**; and, when the surface **accumulates items awaiting disposition**, when that consumer is additionally invoked on a **named cadence** by a **named runner**.
>
> The keying is on *something is meant to read this*, never on *this writes a file*. Excluded by name: declaration modules, a surface whose only legitimate consumer is a human, tests, and transient artifacts.

**The gate** — `scripts/helpers/tests/unit/test_every_producer_NAMES_ITS_CONSUMER.py`, parametrized over three surfaces, each population read **off disk**:

| Surface | Population | Accumulates? | Excluded by name |
|---|---|---|---|
| `scripts/helpers/measure/` | `*.py` | no | `run_log.py` — declaration module |
| `scripts/helpers/` | every regular file | no | `README.md` — the table itself |
| `tracked/` (sibling planning repo) | store directories | **yes** | `operations/` — cadence-exempt, human-only |

Every population additionally drops **transient artifacts** (`__pycache__`, dot-entries) — the docstring's fourth exclusion class, which shipped as prose with no code behind it and took CI red. Unlike the three named exclusions it is deliberately *not* asserted against the tree: a transient artifact's absence is the normal case.

**The new table** — `scripts/helpers/README.md` names the invoker of all 11 tools on disk, and **the gate opens the path it names.** That is stronger than the `measure/` gate's name-only claim, and it is justified by the phase's own sharpest failure: `harvest-intake.py` is a stated *condition of an exemption* invoked by one line of prose in `config/commands/standup.md`, and until now dropping that line would have left the intake accepting forever with no suite going red.

[...]

Issue #146, opened 2026-08-28
RECURRENCE on C-523klr8n — the map's per-component annotation restates roadmap state and decays with no gate able to see it


Pumapumapumas:
---
store: candidates
status: open
filed_by: review-pr
component: workflow-decomposition
---

**THIS IS A RECURRENCE ON `C-523klr8n`, NOT A NEW ITEM.** Harvest action: increment `C-523klr8n`'s `count` (1 → 2) and append the line below under its `## Recurrences` heading. **Do not create a new candidate file.**

```
- 2026-08-28 · review-pr on Claude-Dot-Files#145
```

## Why this is that item and not a new one

`C-523klr8n` is *"derive declared roster and inventory counts rather than restating them in prose."* Its body already names `docs/file_structure.txt` twice as an offender (third occurrence: the map restating the nine-rule count; sixth occurrence: `file_structure.txt:97` restating *"Phase 3 BUILT-NOT-PROVEN … 7 of 8 requirements"*), and it explicitly ruled that the candidate **must cover restated STATUS as well as restated COUNTS**, having measured that *"restate a status has no syntactic signature the way restate a count does"* and proposing *"a declared canonical status field per phase doc, compared rather than pattern-matched."*

The occurrence below is exactly that half of the candidate, in exactly that file.

## The occurrence

`docs/file_structure.txt:125` restates two values `docs/development/workflow-decomposition/roadmap.md` owns:

```
workflow-decomposition/   # 🟡 IN PROGRESS per sprint.md — Phase 1 shipped; Phases 2-5 open, rollout order 2 → 3 → 4 → 5
```

Three of its claims are false as of PR #145's head (`c88d1ff1d46d719e1e0614bb5712a5e29d86df10`): *Family alignment* shipped 2026-08-18; two further phases now exist (*Every producer names its consumer*, *Managed configuration, and whose tier wins*); and the rollout order is now 3 → 4 → 5 → 6 → 7.

## Why no gate sees it, which is the part that makes it this candidate's status half

`testing/scripts/tests/unit/test_planning_directories_are_ROLLED_UP_in_the_map.py` names this blind spot in its own docstring under WHAT THIS DOES NOT LOOK AT — *"Whether the ANNOTATION on the surviving component row is true. Same blind spot every check on this map has."* The roll-up that resolved `C-pky2l2b6` deliberately left the annotation unguarded, so the surviving row's prose decays silently and nothing ever goes red.

[...]

PR #193, opened 2026-09-14
build-draft: WD Phase 6 — rule the runtime-record class in the producer gate

Pumapumapumas:
## Summary

**Closes WD Phase 6's last open step — the runtime-record class.** The producer gate (`testing/scripts/tests/unit/test_every_producer_NAMES_ITS_CONSUMER.py`) now rules the records a run writes under the journal root, which its `git ls-files` walk structurally cannot see. Two files: the gate and its annotation in `docs/file_structure.txt`; no journal/emit module touched.

**Mechanism — a named ruling whose readers are asserted to resolve, mirroring `HELD_BY`:**

- `RuntimeRecord` / `RUNTIME_RECORDS` — record type → (declaring module + symbol, reader module + entry function, `via` hops). A reader **resolves** only when: the module is on disk, `def <entry>(` is defined in it, the module names the record's declared symbol **as a code token it LOADS** (AST `Name`/attribute in `Load` context, or a string-as-value — never a comment, docstring, import alias, `__all__` entry in any spelling, or the declaration's own `Store`), and every `via` hop imports the next by stem, so the tool an operator is told to run actually reaches the function that opens the record. An entry stops at its first failing part and reports once.
- **Population is derived, not listed** (the file's own principle): every journal constant matching `_RECORD_CONSTANT` (`*_FILE` / `*_DIR`, 9 today — bare or annotated, via `_all_targets`) and every run-log member in `MEMBER_EVENT_TYPES` (loaded by path, 3) must be registered or baselined, keyed by repo-relative path. A record the fleet starts writing fails until someone names its reader.
- `UnreadRecord` / `UNREAD_RUNTIME_RECORDS` — a declared record with no reader, frozen with a **ratchet** on its read accessor: the day a fleet module names it in code, the line must leave.
- Docstring bullet rewritten from "It is NOT ruled here … the one open extension" to the ruling and its blind spots.

**Thirteen entries registered**, all resolving on this branch:

| Record | Declared | Reader (entry) | Reached via |
|---|---|---|---|
| bag payload dir / manifest / bagit.txt / bag-info.txt | `bag.py` | `validate.py::validate_bag` | `validate_bag.py` |
| config digest tag, across two bags | `config_digest.py` `LABEL_CONFIG_DIGEST` | `compare_run_config.py::_read_digest` | — |
| citations | `citations.py` `CITATIONS_FILE` | `citations.py::read_citations` | `verify_citations.py` → `verify.py` |
| content store | `content_store.py` `CONTENT_STORE_DIR` | `content_store.py::load_object` | `verify_citations.py` → `verify.py` |
| harvest index | `harvest.py` `HARVEST_INDEX_FILE` | `harvest.py::read_harvest_indexes` | `reconcile_harvest.py` |
| edge-id file | `edge_id.py` `EDGE_ID_FILE` | `edge_id.py::resolve_edge_id` | `emit.py` |
| typed exit record | `exit_record.py` `CHILD_SCHEMA` | `exit_record.py::route` | `review_pr_workflow.py` |
| run log `parent_route` / `run_resources` / `convergence` | `run_log.py` | `replay_parent_route.py::rows` / `replay_run_resources.py::main` / `replay_convergence_events.py::rows` | — |

The run log's **write** side is untouched (`HELD_BY` → `test_run_log.py`); this is its read side. Two entries are outside both derived families and registered by hand with that said: the exit record (one schema in one module) and the config-digest tag (a label inside `bag-info.txt`).

### The bag's event stream (`events.jsonl`) has no committed reader — yet

**Consequence:** the journal's primary record — the paired intent/completion events every fleet write path emits (#192) — is decoded by nothing outside the test suite. `decode_event`, `dedupe_on_identity` and `applied_intents` exist in `modules/journal/events.py`; every code caller is a test (re-verified in the refine pass by per-function `\bfn\(` grep over `scripts/` minus tests: the only non-test hits are docstring/comment mentions in `emit.py`, `merge_pr.py`, `harvest_activities.py`). The only on-disk read-back is the emitter's own idempotency index (`_appended_pairs`, `json.loads` per line on its own file), which is the writer's use, not a consumer. A run's event stream can stop being correct and nothing goes red. Issue #179 corroborates: it reasons about "replay" via `dedupe_on_identity` with no replay tool existing.

**Remedy:** none to design — the reader is **scheduled**: Persistent Memory Protocol Phase 4's replay step (`skyynet-master-planning/development/edge-assistant/persistent-memory-protocol/phase4_rebuild_is_a_test.md` l.176, *"Build the replay: read one edge's journal in order, dedupe on event identity, apply each event to a scratch directory"*), which `applied_intents`'s docstring already names as its consumer. Baselined here as `UNREAD_RUNTIME_RECORDS["a bag's event stream, events.jsonl"]` with the ratchet; the line leaves the day that replay lands as a fleet module (the ratchet walks `scripts/` only).

**Store:** none — **DEFERRED to the existing Phase 4 item, not filed.** The draft asked `review-pr` to file a `tracked/candidates/` item; review-pr pass 1 found the plan already holds it (its issue-and-store search did not reach the phase docs) and ruled `defer-to-existing-work`, pointer verified at planning `1f87880`. The gate's baseline comment said the same wrong thing ("a reader is a design decision") and is corrected in the second correction pass below.

## Refine pass (build-refine)

Three holes in the draft's AST predicates, one class — each helper read the single spelling the live tree uses today — each confirmed by a mutation that predicted **0** failures and observed 0, each closed with a control that now fires:

| Hole | Sites | Fix |
|---|---|---|
| `_code_tokens` counted a `Store`-context `Name` as a use, so an entry whose reader IS the declaring module could never fail "names it in code" | **5 of 13** entries (`citations`, `content_store`, `harvest`, `edge_id`, `exit_record`) — exhaustive over the registry | `Load` context only; `self-store-only` control case |
| `_journal_record_symbols` walked only `ast.Assign`; an annotated `X_FILE: str = …` escaped the population (the journal declares 5 constants annotated today) | 1 helper (`_declared_tokens` already handled both) | `_all_targets` reads bare / annotated / augmented; both derivations share it; `A_DIR: str` in the control |
| `_code_tokens`' `__all__` exclusion matched only the bare assignment | 1 site | same helper; `reexports.py` control with the annotated and augmented spellings |

Also: `_unresolved_records` now stops at the first failing part for every part (its docstring already claimed so; the reader checks could co-fire); population keyed by repo-relative path, not basename; `_unresolved_holders` shares `_defines_function`; `_module_ast` cached and cleared with the other caches; the control named `SIX` asserted seven, now asserts nine once each; the `via` prose no longer claims a first hop's liveness the import graph holds. Merged `origin/main` (3 commits, none touching the gate).

## Correction pass 2 (build-refine-minor, on review-pr pass 1's runway)

| Item | Disposition |
|---|---|
| F1 — the gate's `events.jsonl` baseline comment and `reason` say the reader "is a design decision" | **FIXED** — both, plus the module docstring's own bullet (third site of the same fact, found by code-reviewer), now cite PMP Phase 4's replay and say not to file a candidate; "frozen, not fixed" and the #192 sentence kept |
| F2 — `docs/file_structure.txt` omits the runtime-record class and says "Eight unread producers" | **FIXED** — the annotation names the class as ruled by declaration and states the measured figures (6 producers, 1 runtime record) |
| code-reviewer W — the F1 comment grew a `measure/` placement steer beyond the runway | **FIXED** — steer trimmed; the ratchet-scope fact kept |
| code-reviewer High — a count in the map is held by nothing; "Eight" was already stale on `main` | **FIXED as a class check** — `test_the_REPO_MAP_states_the_UNREAD_figures_this_gate_holds` reads the gate's own map entry (`file_structure_map.LEAF`) and asserts the sentence's digits against `SURFACES` + `UNREAD_RUNTIME_RECORDS`; a vanished sentence fails too. Control `test_the_MAP_FIGURE_check_reads_the_sentence_and_only_that_sentence`; five negative controls, each at its predicted count (see the Decision Log comment) |
| code-reviewer Low — annotation length vs the map's prose density | **REJECTED** — matches the map's local convention; the checker's prose note is informational and pre-existing (net +3 continuation lines after tightening) |

Class sweep for F1 (exhaustive: `grep -i 'design decision|candidates|events.jsonl|event stream|Phase 4'` over the gate and the diff): three in-file sites and this body's finding section — all four now say the same thing. The runway was sized as two prose edits; closing the High's class added one guard (+2 tests, +1 import), stated here rather than absorbed.

## Deviations from plan

- Two extra readers beyond the four the task named (`reconcile_harvest.py` for the harvest index; `emit.py` for the edge-id file) and the run log's three members — forced by the derived population: once the journal's constants are the population, every one needs a ruling, and the phase doc (line 141) says the step is the read obligation "for the class as a whole".
- One CI-only red found and fixed before the draft's push: `test_prose_NAMES_a_symbol_that_RESOLVES.py` read `` `*_FILE` `` in prose as a private symbol. Reworded to cite `_RECORD_CONSTANT`.

## Planning artifacts

**No checkbox flipped.** The Phase 6 doc's 17th step is the operator's reconciliation; the sibling repo was read, not written.

## Test results (correction-pass head `614b5c4`)

- Gate alone: **89 passed, 49 skipped** (+2 over the refine head: the map-figure check and its control).
- `ruff check --select F821,F401,E9`: clean.
- `./testing/run-all.sh unit` (sibling present): `RESULT: PASSED` — 5887 passed / 61 skipped.
- **Sibling-less clone (the CI condition)**, the six affected suites: 113 passed / 64 skipped — the new import (`file_structure_map`) does no planning-repo I/O at import time.
- `scripts/helpers/file_structure_check.py`: exit 0.
- `gh pr checks 193` on the previous head: `suite` pass, CodeQL ×3 pass.

## Test results (refine head, after the merge — superseded above, kept for the record)

- Gate alone: **87 passed, 49 skipped** (baseline at `7709b87`: 81 / 49; +6 new tests, skips unchanged — the per-surface "excludes nothing by name" parametrisations, read one by one, none environment-derived).
- `ruff check --select F821,F401,E9`: clean.
- Full `./testing/run-all.sh`, **sibling-less clone (the CI condition)**: `RESULT: PASSED` — unit 4975 passed / 132 skipped, integration 331 passed.
- Mutation controls: draft 6 + refine 11 (5 re-run against the draft, 6 against the refined predicates) — every predicted count hit. See the Decision Log comments.


Pumapumapumas:
## Decision Log

**Rigour tier: safety control (a gate) — 6 mutations**, each with a predicted failure count stated before the run and restored from a copy afterwards (never `git checkout --`):

| # | Mutation | Predicted | Observed |
|---|---|---|---|
| M1 | rename a registered entry function (`validate_bag` → `validate_bags`) | 1 (RESOLVES) | 1 |
| M2 | point a `via` hop at a tool that does not import the reader | 1 (RESOLVES) | 1 — reported for all four `validate_bag` rows in one failure |
| M3 | delete the `EVENTS_FILE` unread line | 2 (POPULATION unruled + UNREAD empty) | 2 |
| M4 | register `convergence` under `replay_parent_route.py` | 1 (never names in code) | 1 |
| M5 | swap the exit record's symbol to one `exit_record.py` only documents (`jsonschema`) | 1 (declares no) | 1 |
| M6 | drop a scratch fleet module that CALLS `decode_event` | 1 (unread ratchet) | 1 |

Two misses during authoring, both **model-of-fixture wrong** (no guard finding): (a) the tmp_path control predicted 7 defects and observed 8 — an undeclared symbol also reported at the reader; fixed so an entry stops at the first failing part; (b) the live unread ratchet named `journal/__init__.py` as the event stream's reader on the strength of a string in `__all__`. That one changed the predicate — `__all__` entries are now excluded alongside import aliases — and the control fixture gained an `__all__` re-export so the exclusion is asserted, not assumed. Re-run after each: predicted counts hit.

**What the guards do not look at** (also stated in the gate's docstring): the ruling never opens a bag, so a reader naming the right constant and reading the wrong file passes; a record declared under a name `_RECORD_CONSTANT` does not match is outside the derived population (the exit record is registered by hand for that reason); the unread ratchet keys on the read accessor, so a reader that re-types the filename and parses lines itself is invisible to it.

- **[Medium]** Registered the **library function that opens the record** (`validate.py::validate_bag`, `citations.py::read_citations`) as the reader, with the operator tool the task named as a `via` hop, rather than registering the tool itself. Alternatives: register `validate_bag.py` directly and accept an import as evidence. Why: the entrypoints are 40-line shims that name no record constant; "reads that record type" is only assertable where the constant is used, and the `via` chain is what keeps the operator-facing name in the ruling.
- **[Medium]** Derived the population from the journal's `*_FILE`/`*_DIR` constants and `MEMBER_EVENT_TYPES` rather than a hand-kept list of the four readers the task named. Alternatives: the task's four entries only. Why: the file's own principle (a table checked against itself cannot see the member never added) and the phase doc's "read obligation for the class as a whole". Cost: two extra readers and the `events.jsonl` finding, which is the value.
- **[Medium]** Evidence of reading = the reader names the declared symbol as a code token (Name / attribute / string-as-value), module-wide, not entry-function-scoped. Alternatives: scope to the entry function's body; require a call-shape match. Why: `content_store.load_object` reaches `CONTENT_STORE_DIR` two private calls down in the same module; function-scoped evidence would report it falsely. String-as-value is what lets `rl.events(log_dir, "parent_route")` count for the run log.
- **[Low]** Baselined `events.jsonl` as unread rather than writing a reader. Alternatives: add a `replay_events.py`. Why: the brief's constraint (gate extension only; the emit rule just landed) and the gate's own rule — inventing a consumer to make a check green is how a gate gets routed around.
- **[Low]** Folded two local wip commits into one attested commit via `git reset --soft` before the first push. Alternatives: push both wips. Why: the brief requires the trailer on the commit and `git.md` forbids amending; a soft reset of never-pushed commits is neither.

## Deferred Work

- **A reader for the bag's event stream (`events.jsonl`)** — Why deferred: a PROPOSAL, not a defect — capability that does not exist, and a design decision outside a gate-only PR. Tracked at: **surfaced in the PR body under its own heading for `review-pr` to file in `tracked/candidates/`** — I hold no write grant on that store. Verified by: `grep -rln 'decode_event\|applied_intents\|dedupe_on_identity' scripts | grep -v /tests/` → only `modules/journal/events.py` and `modules/journal/__init__.py` (re-export) on this branch; the ratchet line `UNREAD_RUNTIME_RECORDS["a bag's event stream, events.jsonl"]` lands at merge and fails the day a reader appears.
- **Phase 6 doc reconciliation (step 17 → `[x]`, status line)** — Why deferred: the brief reserves it for the operator; flipping it is a sibling-repo write. Tracked at: `skyynet-master-planning/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md` line 141. Verified by: `grep -n 'Rule the runtime-record class' <that file>` on the live sibling → line 141, `- [ ]`, names the four readers this PR rules.

## Post-Run Reflection

- **Friction:** the gate's prose-symbol guard (`test_prose_NAMES_a_symbol_that_RESOLVES.py`) reads `` `*_FILE` `` as a private symbol `_FILE`; only visible in the full suite, not the gate alone. The brief's "run-all.sh under the CI condition" instruction caught it before push — worth keeping in every gate-touching brief.
- **Project-level suggestions (this repo):** the brief named four readers; the tree has six operator-facing readers of journal records (`reconcile_harvest.py` and the emitter's edge-id read were unnamed). A brief that says "rule the class" is better served by naming the population's *derivation* than its members — the derivation found the unread record, the list would not have.
- **Tooling-level suggestions (claude-dot-files):** none beyond the above.

Pumapumapumas:
## Decision Log (build-refine)

**Rigour tier: safety control (a gate) — 11 mutations**, 5 re-run against the draft's predicates and 6 against the refined ones, each with a predicted count stated before the run; files restored from a copy (never `git checkout --`). Plus two live probes that predicted **0** and got 0 — the holes this pass closed:

| # | Mutation | Predicted | Observed |
|---|---|---|---|
| M1 | rename a registered entry function (`validate_bag` → `validate_bags`) | 1 | 1 |
| M3 | delete the `EVENTS_FILE` unread line | 2 | 2 |
| M4 | register `convergence` under `replay_parent_route.py` | 1 | 1 |
| M8 | SCOPE: narrow `_RECORD_CONSTANT` to `_FILE$` | 1 (population control, `A_DIR`) | 1 |
| M9 | SCOPE: stop excluding `__all__` in `_code_tokens` | 2 (live ratchet via `journal/__init__.py` + control) | 2 |
| **M10** | append `SCRATCH_FILE: str = …` to a journal module (draft gate) | **0 — a hole** | 0 |
| **P1** | `RuntimeRecord("decl.py","MEMBERS","decl.py","open_it")` — self-referential, `Store` only (draft gate) | **`[]` — a hole** | `[]` |
| M12 | drop the `Load`-only filter | 1 (`self-store-only` control) | 1 |
| M13 | `__all__` exclusion back to bare `Assign` | 1 (`reexports.py` control) | 1 |
| M14 | remove the `continue` after the entry-fn check | 1 (`entry-and-symbol-gone` reports twice) | 1 |
| M15 | `_ruled_records` back to basename keys | 2 (live population + control) | 2 |
| M16 | `_defines_function` → `return True` | 2 (HOLDER control + RUNTIME control) | 2 |
| M17 | `_all_targets` ignores `AnnAssign` | 1 (population control) | 1 |
| M10′ | M10 re-run on the refined gate | 1 | 1 |
| P1′ | P1 re-run on the refined gate | 1 report | 1 |

**Class sweep.** P1's class is *every entry whose reader is its declaring module*: **5 of 13** (`citations`, `content_store`, `harvest`, `edge_id`, `exit_record`) — exhaustive, read off `RUNTIME_RECORDS`; quality-control reported it as one site. M10's class is *every AST helper reading one assignment spelling*: `_journal_record_symbols` and `_code_tokens`' `__all__` loop — exhaustive over the four new AST helpers (`_declared_tokens` already read both; `_imports_stem` handles both import forms). Blast radius of my own edit: the new backticked prose was run through `test_prose_NAMES_a_symbol_that_RESOLVES.py` (25 passed) — the gate that caught the draft — and the full suite under the CI condition.

- **[Medium]** Closed QC's self-referential hole by counting only `Load`-context names, rather than by excluding the declaring statement's targets. Alternatives: skip top-level `Assign` targets in `_code_tokens`; scope evidence to the entry function's body. Why: "a use is a Load" is the semantic the AST already carries and it also makes the unread ratchet stricter for free (a fleet module that *assigns* `decode_event = …` is not reading); the draft's own Decision Log explains why function-scoped evidence reports `content_store.load_object` falsely.
- **[Medium]** Re-keyed the population by repo-relative path (code-reviewer Medium) instead of adding a basename-uniqueness assertion. Alternatives: the assertion; leave it. Why: the key now equals `RuntimeRecord.declared_in`, so the population test compares like with like and the fixture threads `repo=` the way `_unresolved_records` already does — the assertion would have guarded a representation the path key makes unnecessary.
- **[Low]** Rejected code-reviewer's remedy of requiring `via[0]` under `_TOOLS`, kept the concern: the exit record's first hop is the parent branch the brief itself named, and `emit.py` opens the edge-id file on every later bag open — neither is a tool. Fixed the prose instead: the comment now says the first hop's own liveness is the import graph's, which the module docstring already rules for every module.
- **[Low]** Cached `_module_ast` (code-reviewer Low, "defer") rather than deferring. Why: the unread ratchet tokenises every fleet module per baselined record, so it is the file's `_imported_stems` case exactly; one decorator plus one entry in `_clear_corpus_caches`, and a deferral had no home.
- **[Low]** Rejected retrofitting `HELD_BY` to a dataclass (code-reviewer Low): the brief said mirror `HELD_BY`, not rewrite it; `HELD_BY` is untouched by this PR.

### Disposition of every finding

| Finding | Source | Disposition |
|---|---|---|
| Annotated journal constant escapes the population | Stage 1 (F1), mutation-confirmed | **FIXED** — `_all_targets`; control |
| Control named `SIX` asserts seven | Stage 1 (F2) | **FIXED** — renamed, asserts nine once each |
| Draft's "Verified by" grep does not reproduce as written (5 files, not 2) | Stage 1 (F3) | **FIXED** in the PR body — substance held (no code caller); attestation re-stated with the command that does reproduce |
| Reader checks co-fire vs "stops at first part" docstring | code-reviewer W | **FIXED** — `continue`; `entry-and-symbol-gone` control |
| `__all__` exclusion misses annotated/augmented | code-reviewer W | **FIXED** — `_all_targets`; `reexports.py` control |
| `via[0]` liveness not enforced | code-reviewer W | **REJECTED remedy / FIXED prose** (above) |
| `_unresolved_holders` duplicates `_defines_function` | code-reviewer M | **FIXED** — shared; M16 fires both controls |
| Basename-keyed population | code-reviewer M | **FIXED** — path-keyed |
| `_module_ast` uncached | code-reviewer L | **FIXED** |
| `HELD_BY` idiom mismatch | code-reviewer L | **REJECTED** — out of the brief's scope, file untouched there |
| Self-referential entry cannot fail "names it in code" | quality-control W | **FIXED** — `Load` only; 5 sites |
| "exit record is the one member of neither family" overstates | quality-control I | **FIXED** — comment and docstring name the config-digest tag too |
| `_run_log_members` exec's a module by path | quality-control (security net) | **REJECTED** — trusted repo code, same loader the replay tools use; QC reached the same conclusion |

## Deferred Work

- **A reader for the bag's event stream (`events.jsonl`)** — Why deferred: a PROPOSAL (capability that does not exist) and a design decision outside a gate-only PR. Tracked at: **surfaced in the PR body under its own heading for `review-pr` to file in `tracked/candidates/`** — no write grant here. Verified by: `for f in $(grep -oE "^def [a-z_]+" events.py); grep -rlE "\b$f\(" scripts | grep -v /tests/` → only `emit.py`, and only for `event_identity`/`encode_event`/`gap_event`/`redaction_placeholder_event` — no code caller of `decode_event`, `dedupe_on_identity` or `applied_intents`; `gh issue list --search "journal"` → nothing filed for a reader; #179 corroborates. The ratchet line lands at merge and fails the day a reader appears.
- **Phase 6 doc reconciliation (step 17 → `[x]`, status line)** — Why deferred: the brief reserves it for the operator. Tracked at: `skyynet-master-planning/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md` line 141. Verified by: draft's `grep -n 'Rule the runtime-record class'` → line 141, `- [ ]`; not re-touched by this pass.

## Post-Run Reflection

- **Friction:** the brief's "give every new predicate a self-contained positive control" was satisfied by the draft and still shipped three holes, because a control proves a predicate discriminates *within its fixture* and every fixture entry was cross-file with bare assignments. The instrument that found them was **predicting zero and probing the live tree** (an annotated constant appended to a journal module; a self-referential registry entry) — i.e. asking what each guard does not look at, which the prompt already says. Worth stating in gate-touching briefs as a step rather than a principle: *"for each predicate, name one live shape in the tree the fixture does not contain, and probe it."*
- **Project-level suggestions (this repo):** none beyond the above.
- **Tooling-level suggestions (claude-dot-files):** the Stage 1 instruction "MUTATE it, do not merely run it" and the later "ASK WHAT EACH GUARD DOES NOT LOOK AT" are the two halves of what worked here, but they sit ~40 lines apart and read as separate disciplines. Both holes were found by the second, then confirmed by the first. Consider folding them: *"for each new predicate: mutate it (negative control), then name a shape it never inspects and probe the live tree with that shape (predict 0)."*

Pumapumapumas:
## review-pr · pass 1 · PR #193 — disposition

**Reviewed:** head `3fbc830ad294e54a293282595732fa2218c3bc56` (3 commits: draft `86b9c83`, refine `b8d6728`, merge `3fbc830`). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `closingIssuesReferences: []`, `refs/pull/193/merge` = `072ae0c` (present, fresh); checks on the merge ref: `suite` pass, CodeQL ×3 pass. Worktree repo = PR repo. Body is about THIS PR — every artifact it names is in the one-file diff. **Merge-base `f908122` IS `origin/main`'s tip**, so no rule arrived after the cut and nothing this runway prescribes has already landed. The merge commit's resolution is byte-identical to `git merge-tree` of its parents (empty diff) — nothing discarded. Reflection coverage is per-commit: draft and refine each attested; the merge is mechanical. Prior `pr_review:` blocks on the thread: **0** (fence-anchored count) → this is pass 1; attempts consumed: 2 (two attesting reflections).

**What I re-executed rather than read** (the run's own measurements): journal `_RECORD_CONSTANT` population = **9** bare constants across 6 modules ✅; `MEMBER_EVENT_TYPES` = 3 ✅; "five annotated constants in the journal" = `RULES`, `CASE_D_CHANNELS`, `_CASE_D_REPORTER`, `_CURRENT`, `_DIGEST_FAILURE_REASONS` — 5, none `_FILE`/`_DIR`, which is why M10 predicted 0 ✅; per-function `\bfn\(` grep over `scripts/` minus tests for `decode_event` / `dedupe_on_identity` / `applied_intents` → only their `def`s and one internal call in `events.py` ✅; gate alone **87 passed / 49 skipped**, every skip a per-surface "baselines nothing" ✅; no `*_FILE`/`*_DIR` constant outside the journal in the fleet (`_PHASE_FILE` is a regex) and no journal subpackage, so the non-recursive glob loses nothing today ✅; `run_log.py` LOADS `MEMBER_EVENT_TYPES` at l.315/318, so the self-referential run-log entry resolving is genuine, not the P1 hole ✅. **Mutations re-run at function level** (a different instrument from the run's file edits): M4 → 1 (`never names convergence in code`) ✅; M2 → 1 (`via reconcile_harvest.py does not import validate`) ✅; live ratchet probe → no fleet module of 116 names `decode_event` as a use, and `journal/__init__.py` reads False under the `__all__` exclusion ✅. `ast.walk` is a deque BFS so the parent-before-child property the docstring skip relies on holds.

**Sweeps run:** deleted-artifact — the old docstring bullet's four claims each map to a destination (run-log write side → `HELD_BY` still; readers → `RUNTIME_RECORDS`; `compare_run_config.py` human-findability baseline → was at l.736 and has since CLOSED on main, `unread={}`, `operations.md:255/261` names it; "one open extension" → the ruling). Asymmetric-presence — **found one**: `docs/file_structure.txt`'s annotation of this gate (F2). Inbound-citation — no `test_every_producer_NAMES_ITS_CONSUMER.py:<line>` citation in either repo. Boundary — see notes. Predecessor deferrals — #192's "ratchet red" landed on main as `4255964`; #190's `operations.md` escalation is done on planning main (both tools named). Completion-checkbox — no planning artifact in the diff; the run's "no checkbox flipped" claim verified. Tables — the body's 13-row table equals the registry (4+1+2+1+1+1+3).

**Angle new to this PR:** re-verifying the run's *deferral placement* against the **planning corpus** rather than the issue queue it searched — which is where F1 comes from.

### Findings of this pass

**F1 · `events-reader-baseline-says-undecided`** — *The gate's own baseline says the `events.jsonl` reader is an open design decision, when the plan already schedules it — so the next reader (this PR's body included) files a duplicate candidate.* The `UNREAD_RUNTIME_RECORDS` comment reads *"a reader is a design decision, and the emit rule that just landed (#192) is not the place to grow one"*, and the PR body asks me to file a `tracked/candidates/` item for "a design decision, not a one-liner". Verified against the planning corpus: **PMP Phase 4** (`persistent-memory-protocol/phase4_rebuild_is_a_test.md`, status *not started*, step at l.176: *"Build the replay: read one edge's journal in order, dedupe on event identity, apply each event to a scratch directory"*) IS that reader, and `events.py::applied_intents`'s docstring already names Phase 4 as the consumer it exists for. The run searched `gh issue list` and the candidate store; it did not open the phase doc. Consequence: a comment in a safety gate misstates the planning state, and every later reader re-derives a candidate the plan already holds. **Remedy: `fix-in-place`** — the comment and the `reason` string cite Phase 4's replay step as the scheduled reader (one edit; precheck run and pasted below). `/decide` reframe: *is the reader missing, or is its home missing?* — neither: the home exists and the gate does not point at it. `/best-practices`: a baseline entry names where it will be released from, the way the tools surface's baseline named `_consumer_file`.

**F2 · `repo-map-omits-runtime-record-class`** — *The authoritative repo map describes this gate's populations and counts, and after this PR it names neither the runtime-record class nor the right unread count.* `docs/file_structure.txt:1195–1206` says the gate rules table surfaces + derived surfaces with "Eight unread producers found so far … on a two-way ratchet". Measured on this branch: `sum(len(s.unread))` = **6** (the two tools left on main at `4255964`) **+ 1** `UNREAD_RUNTIME_RECORDS`; and the class this PR added — declaration-derived, not walked — is absent from the annotation. This PR built the class; the annotation is the map's summary of the gate; the PR is the change that made it incomplete (the count going 8→6 is main's, and is corrected in the same line). Consequence: the repo's stated-authoritative map tells the next dispatch the runtime-record class is unruled — the same stale fact the Phase 6 status line carries until the operator reconciles it. **Remedy: `fix-in-place`**, same dispatch as F1. Precheck run and pasted below.

### The producing run's deferrals — re-adjudicated

**`events-jsonl-reader`** — the run deferred "a reader for the bag's event stream" as a PROPOSAL for me to file into `tracked/candidates/`. **DEFERRED, pointer = PMP Phase 4 step l.176, `pointer_verified: true`** (read the file on the planning checkout at `1f87880`). **Not filed** — four-destinations rule #2/#3: it is an existing planning item, and filing a candidate would create a second entry for one concern. The run's placement claim ("no existing issue covers it") was wrong because it searched the wrong corpus, but the deferral itself was honest and disclosed (no grant claimed) → neither laundered nor homeless; the correct pointer exists and I have recorded it. Duplicate check run anyway: `similar-candidates.py --store candidates --repo claude-dot-files --component workflow-decomposition` → nearest are `C-wb1xc1xs` (run-log `convergence` reader — different record, and its reader now exists as `replay_convergence_events.py`, see notes) and intake #179 (event-id collision — a defect in identity, not a reader); `--store issues` → nothing on this mechanism.

**`phase6-step17-reconciliation`** — **DEFERRED, `remedy: operator-action`**, pointer = `workflow-decomposition/phase6_every_producer_names_its_consumer.md` l.141 (`- [ ] **Rule the runtime-record class**`, verified on planning checkout `1f87880`) and its status line l.3 ("16 of 17 … the one still open is the runtime-record class"). The brief reserves the sibling-repo write for the operator. Recommended text after merge: l.141 → `[x]` *(Discharged by Claude-Dot-Files#193: 13 records registered on a declaration-derived population; `events.jsonl` baselined unread — its reader is PMP Phase 4's replay.)*; l.3 → "17 of 17 … Delivered across #170, #181, #190 and #193" and whatever status the operator rules for a phase whose steps are all met.

### Disposition table

|---|---|---|---|
| `events-reader-baseline-says-undecided` (F1) | doc-drift | **HOLD · redispatch** | Baseline comment says undecided; PMP Phase 4 l.176 schedules the replay reader; `applied_intents` docstring names it. One comment edit. |
| `repo-map-omits-runtime-record-class` (F2) | doc-drift | **HOLD · redispatch** | `file_structure.txt:1195–1206` omits the class; "Eight" vs measured 6+1. Same dispatch. |
| `events-jsonl-reader` | deferral | **DEFERRED** | → PMP Phase 4 `phase4_rebuild_is_a_test.md` l.176 (verified). Not filed; the plan already holds it. |
| `phase6-step17-reconciliation` | deferral | **DEFERRED** (operator-action) | → Phase 6 doc l.141 / l.3 (verified `[ ]`). Operator flips after merge; text above. |
| `annotated-constant-escapes-population` | correctness | FIXED | `_all_targets` reads Assign/AnnAssign/AugAssign; control `A_DIR: str`; M17 → 1. Verified in diff and by running the control. |
| `control-name-count-mismatch` | test-gap | FIXED | Renamed `SEVEN`; asserts nine entries covering seven defect kinds (the seven `_unresolved_records` messages). Name = kinds, entries = 9 — consistent. |
| `verified-by-grep-not-reproducible` | doc-drift | FIXED | Body re-states the per-function grep; I re-ran it — only `def`s and one internal call. |
| `reader-checks-co-fire` | correctness | FIXED | `continue` after every part; `entry-and-symbol-gone` reports once at the entry (control asserts). |
| `all-exclusion-single-spelling` | correctness | FIXED | `__all__` via `_all_targets`; `reexports.py` control; M13 → 1. |
| `via-first-hop-liveness` | correctness | FIXED | Concern resolved by prose, remedy (`via[0]` under `_TOOLS`) rejected with sound reasoning: `emit.py` and `review_pr_workflow.py` are first hops and not tools; every fleet module's liveness is the import-graph test's. |
| `unresolved-holders-duplicates-defines-function` | correctness | FIXED | Shared `_defines_function`; M16 fires both controls. |
| `basename-keyed-population` | correctness | FIXED | Path-keyed; control asserts `other/a.py` ≠ `pkg/a.py`; M15 → 2. |
| `module-ast-uncached` | correctness | FIXED | `lru_cache` + entry in `_clear_corpus_caches`. |
| `held-by-dataclass-retrofit` | scope | REJECTED | The brief said mirror `HELD_BY`, not rewrite it; `HELD_BY` is untouched in the diff. A design preference, not a defect. |
| `self-referential-entry-cannot-fail` | correctness | FIXED | `Load`-context only; 5/13 entries in the class; `self-store-only` control; my P1 re-run on `run_log` resolves for genuine Loads at l.315/318. |
| `neither-family-overstates` | doc-drift | FIXED | Docstring and comment name the config-digest tag beside the exit record. |
| `run-log-exec-module` | security | REJECTED | `spec_from_file_location` on a tracked repo file — the identical loader `replay_parent_route.py:44` and `replay_convergence_events.py:55` use; no untrusted input reaches the path. |
| `draft-control-predicted-7-observed-8` | test-gap | FIXED | Root cause was co-firing parts; closed by `reader-checks-co-fire`. |
| `all-string-named-init-as-reader` | correctness | FIXED | `__all__` strings excluded in three spellings; live probe: `journal/__init__.py` → False. |
| `prose-gate-red-on-backticked-glob` | test-gap | FIXED | Reworded to cite `_RECORD_CONSTANT`; `suite` green on the merge ref. |

**Verdict rationale:** every finding the run surfaced is verified fixed or rejected with sound reasoning; both deferrals resolve to existing planning items. Two doc-level defects in the artifact remain — a safety gate's baseline misstating the planning state, and the authoritative repo map not describing the class this PR added — both known one-edit fixes, so the runway is a single `build_minor.sh` pass. Neither would be a HOLD on severity alone at a later pass's convergence check; on pass 1 the bar is "the artifact is right", and it is one dispatch from it.

### WHAT HAPPENS NEXT

1. **redispatch** `build_minor.sh --pr 193` — F1 + F2 (two one-place edits; context and prechecks in the block). Then the next pass is a MERGE.
2. After merge (operator): flip Phase 6 l.141 and reword l.3 as recommended above.

### Notes (not findings — no consequence stated)

- **`C-wb1xc1xs` looks satisfied**: its reader `scripts/helpers/measure/replay_convergence_events.py` exists on main and its docstring says "THIS IS CANDIDATE C-wb1xc1xs … the trigger has fired"; the candidate is still `status: open, count: 1`. A triage close, not this PR's.
- Boundary lens, stated for a later reader: the population glob is non-recursive (no journal subpackage today); `_defines_function` matches `def` not `async def` (all 13 entries are sync); a hand-class record (a second label inside `bag-info.txt`) is outside both derived families — the docstring says so; the unread ratchet trips on a string *value* equal to the accessor name in any fleet module (a loud red with the right message, not a silent pass); the fleet is `scripts/` only, so a Phase 4 replay placed under `testing/` would not trip the ratchet — worth remembering when F1's pointer is followed.
- The run's suggested component for the candidate (`workflow-decomposition`) was the wrong owner — the journal is PMP's — moot now that nothing is filed.

### Post-Run Reflection

- **Friction:** none in the prompt. The one thing the prompt could say more directly: a deferral's "no existing item covers it" claim should be re-verified against the **planning corpus** (phase docs), not only the stores and the intake queue — `similar-candidates.py` searches stores and intakes and cannot see a phase checkbox, which is where this PR's deferral actually lived. The UNIVERSAL "verify against the corpus authoritative for its own claim" rule covers it; the filing section's checklist does not name the phase docs as a place to look.
- **Tooling-level suggestions (claude-dot-files):** `similar-candidates.py` could optionally grep `development/**/phase*.md` checkbox lines for the same rare-term ranking, so a scheduled-but-unbuilt capability surfaces beside the candidates.

```yaml
pr_review:
  pr: 193
  run_id: 667154b273aa4e20b2297096654b34ba
  pass: 1
  attempt: 2
  verdict: HOLD
  converged: false
  findings:
    - id: events-reader-baseline-says-undecided
      title: "The gate's baseline says the events.jsonl reader is an open design decision when PMP Phase 4 already schedules it, so the next reader files a duplicate candidate"
      category: doc-drift
      consequence: "A comment in a safety gate misstates the planning state; this PR's own body already asked for the duplicate filing, and every later reader re-derives it."
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: repo-map-omits-runtime-record-class
      title: "The authoritative repo map does not describe the runtime-record class this PR added and states an unread count the gate no longer holds"
      category: doc-drift
      consequence: "docs/file_structure.txt tells the next dispatch the class is unruled and the count is eight where the gate holds 6 + 1."
      disposition: hold
      remedy: fix-in-place
      hold_kind: redispatch
    - id: events-jsonl-reader
      title: "The bag's event stream is decoded by nothing outside the test suite until PMP Phase 4's replay lands"
      category: deferral
      consequence: "A run's event stream can stop being correct and nothing goes red until the scheduled replay exists; the ratchet line holds the place."
      disposition: deferred
      pointer: "skyynet-master-planning/development/edge-assistant/persistent-memory-protocol/phase4_rebuild_is_a_test.md line 176 (step: Build the replay - read one edge's journal in order, dedupe on event identity, apply each event)"
      pointer_verified: true
      reviewed_sha: 1f87880
    - id: phase6-step17-reconciliation
      title: "The Phase 6 doc says the runtime-record step is open after the PR that closes it merges, until the operator flips it"
      category: deferral
      consequence: "The planning status line under-reports the phase by one step until reconciled."
      disposition: deferred
      remedy: operator-action
      pointer: "skyynet-master-planning/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md line 141 (and status line 3)"
      pointer_verified: true
      reviewed_sha: 1f87880
    - id: annotated-constant-escapes-population
      title: "An annotated journal constant would escape the derived population"
      category: correctness
      consequence: "A record declared `X_FILE: str = ...` would need no reader and nothing would go red."
      disposition: fixed
      remedy: fix-in-place
    - id: control-name-count-mismatch
      title: "A control whose name and assertion count disagree reads as under-tested"
      category: test-gap
      consequence: "A reader trusts the name and misses the extra cases."
      disposition: fixed
      remedy: fix-in-place
    - id: verified-by-grep-not-reproducible
      title: "A \"Verified by\" command that does not reproduce as written weakens the attestation it supports"
      category: doc-drift
      consequence: "A reviewer re-running it gets a different count and doubts the substance."
      disposition: fixed
      remedy: fix-in-place
    - id: reader-checks-co-fire
      title: "Reader checks reporting twice contradict the docstring's stop-at-first-part claim"
      category: correctness
      consequence: "Mutation counts drift from predictions and a control asserts the wrong number."
      disposition: fixed
      remedy: fix-in-place
    - id: all-exclusion-single-spelling
      title: "An annotated or augmented __all__ would count as reading the accessor"
      category: correctness
      consequence: "A re-exporting module trips the unread ratchet falsely."
      disposition: fixed
      remedy: fix-in-place
    - id: via-first-hop-liveness
      title: "The via chain prose claimed a first-hop liveness the check does not assert"
      category: correctness
      consequence: "A reader over-reads the check; the import-graph test holds the property instead."
      disposition: fixed
      remedy: fix-in-place
    - id: unresolved-holders-duplicates-defines-function
      title: "Two copies of the defines-function regex can drift apart"
      category: correctness
      consequence: "One holder check accepts what the other rejects."
      disposition: fixed
      remedy: fix-in-place
    - id: basename-keyed-population
      title: "A basename-keyed population lets a same-basename module elsewhere stand in for the journal's"
      category: correctness
      consequence: "A hand entry pointing at the wrong module satisfies the population test."
      disposition: fixed
      remedy: fix-in-place
    - id: module-ast-uncached
      title: "The unread ratchet re-parses every fleet module per baselined record"
      category: correctness
      consequence: "Gate runtime grows with the baseline; the file already caches the sibling case."
      disposition: fixed
      remedy: fix-in-place
    - id: held-by-dataclass-retrofit
      title: "HELD_BY remains a tuple registry beside a dataclass one"
      category: scope
      consequence: "None in this PR; a design preference outside the brief."
      disposition: rejected
      remedy: reject
    - id: self-referential-entry-cannot-fail
      title: "An entry whose reader is its declaring module could never fail \"names it in code\""
      category: correctness
      consequence: "Five of thirteen entries were unfalsifiable on that part."
      disposition: fixed
      remedy: fix-in-place
    - id: neither-family-overstates
      title: "The docstring claimed one hand-registered record where there are two"
      category: doc-drift
      consequence: "A reader believes the config-digest tag is derived and would not re-register it."
      disposition: fixed
      remedy: fix-in-place
    - id: run-log-exec-module
      title: "The population helper executes run_log.py by path"
      category: security
      consequence: "None — tracked repo file, same loader the replay tools use, no untrusted path."
      disposition: rejected
      remedy: reject
    - id: draft-control-predicted-7-observed-8
      title: "The draft's control observed one more defect than predicted"
      category: test-gap
      consequence: "Model-of-fixture error that hid the co-firing defect until traced."
      disposition: fixed
      remedy: fix-in-place
    - id: all-string-named-init-as-reader
      title: "The first ratchet run named journal/__init__.py as a reader on the strength of __all__"
      category: correctness
      consequence: "The baseline would have been refused for a re-export, not a read."
      disposition: fixed
      remedy: fix-in-place
    - id: prose-gate-red-on-backticked-glob
      title: "The prose-symbol gate read a backticked glob as a private symbol"
      category: test-gap
      consequence: "CI red on the full suite with the gate alone green."
      disposition: fixed
      remedy: fix-in-place
  next_steps:
    - item: events-reader-baseline-says-undecided
      kind: redispatch
      note: "One comment edit in the gate — cite PMP Phase 4's replay step as the scheduled reader."
      dispatch_tool: build_minor.sh
      dispatch_context: |
        On PR #193's branch, in testing/scripts/tests/unit/test_every_producer_NAMES_ITS_CONSUMER.py,
        edit ONLY the comment block above the "a bag's event stream, events.jsonl" entry in
        UNREAD_RUNTIME_RECORDS and that entry's `reason` string. Replace the claim that a reader
        "is a design decision" with the fact: the reader is SCHEDULED — Persistent Memory Protocol
        Phase 4 (`skyynet-master-planning/development/edge-assistant/persistent-memory-protocol/
        phase4_rebuild_is_a_test.md`, step "Build the replay: read one edge's journal in order,
        dedupe on event identity, apply each event"), which `events.py::applied_intents`'s docstring
        already names as its consumer. Keep the "frozen, not fixed" stance and the #192 sentence.
        Then run the gate alone and `test_prose_NAMES_a_symbol_that_RESOLVES.py` (the new prose must
        pass it). Do NOT touch RUNTIME_RECORDS, any predicate, any control, or HELD_BY.
      precheck: |
        Context: the PR branch (build-draft/wd-phase6-runtime-record-class), repo root.
        Command: sed -n '/^UNREAD_RUNTIME_RECORDS/,/^}/p' testing/scripts/tests/unit/test_every_producer_NAMES_ITS_CONSUMER.py > /tmp/pc1.txt; echo "sed=$? lines=$(wc -l < /tmp/pc1.txt)"; grep -c 'phase4_rebuild_is_a_test\|Rebuildability is a test\|Phase 4' /tmp/pc1.txt
        Observed on this pass: sed=0 lines=12, then 0.
        Arms: lines=0 or sed non-zero -> the precheck is broken (block not found): report, do not act.
              count 0 -> proceed (the block does not cite the scheduled reader).
              count >=1 -> ALREADY DONE, STOP.
        Tests for the replacement's PRESENCE, not the old words' absence. PRECEDENCE: the dispatch_context enumeration governs.
    - item: repo-map-omits-runtime-record-class
      kind: redispatch
      note: "Same dispatch — update the gate's annotation in docs/file_structure.txt to describe the runtime-record class and state the measured unread count."
      dispatch_tool: build_minor.sh
      dispatch_context: |
        On PR #193's branch, in docs/file_structure.txt, edit ONLY the annotation lines for
        testing/scripts/tests/unit/test_every_producer_NAMES_ITS_CONSUMER.py (currently l.1195–1206).
        Add one clause: the RUNTIME-RECORD class (journal bag files and tags, the typed exit record,
        the run log's member events) is ruled by DECLARATION — every journal `_RECORD_CONSTANT`
        and run-log member must name a reader that resolves (RUNTIME_RECORDS) or sit in
        UNREAD_RUNTIME_RECORDS on the same two-way ratchet. Replace "Eight unread producers found
        so far" with the count you MEASURE at edit time:
        `python3 -c "import sys; sys.path.insert(0,'testing/scripts/tests/unit'); import test_every_producer_NAMES_ITS_CONSUMER as g; print(sum(len(s.unread) for s in g.SURFACES), len(g.UNREAD_RUNTIME_RECORDS))"`
        (6 and 1 on this pass — state both, e.g. "six unread producers and one unread runtime record").
        Keep the tree structure intact (annotation continuation lines only). Run
        scripts/helpers/file_structure_check.py afterwards; it must exit 0. Do NOT edit any other entry.
      precheck: |
        Context: the PR branch, repo root.
        Command: awk '/test_every_producer_NAMES_ITS_CONSUMER.py/{f=1} f&&/test_no_SURFACE_NAMES_A_RETIRED_AGENT/{f=0} f' docs/file_structure.txt > /tmp/pc2.txt; echo "awk=$? lines=$(wc -l < /tmp/pc2.txt)"; grep -c -i 'runtime-record\|RUNTIME_RECORDS' /tmp/pc2.txt
        Observed on this pass: awk=0 lines=12, then 0.
        Arms: lines=0 or awk non-zero -> the precheck is broken (annotation not found): report, do not act.
              count 0 -> proceed (the annotation does not describe the class).
              count >=1 -> ALREADY DONE for the class clause, STOP — but still re-measure the unread figure and correct it if the annotation's number differs from the measurement.
        Tests the property (the annotation names the class), reconciled against the gate's own registries for the count. PRECEDENCE: the dispatch_context enumeration governs.
  laundered_deferrals:
    caught: 0
    of_total: 2
  homeless_items: 0
  redispatched: false
  filed_intakes: []
```

Pumapumapumas:
## Decision Log (build-refine-minor, correction pass 2)

**Rigour tier: new contract (one added guard) — 5 mutations**, each with a predicted failure count stated before the run; files restored from copies I made (never `git checkout --`), byte-identical afterwards.

| # | Mutation | Predicted | Observed |
|---|---|---|---|
| N1 | map's figure `6` → `8` | 1 (map check) | 1 |
| N2 | map's figures as number-words ("Six … one") | 1 (states none) | 1 |
| N3 | `_map_unread_counts` → always `None` | 2 (map check + control) | 2 |
| N4 | `_map_annotation` stops reading at no leaf | 1 (control: `y.py` reads `z.py`'s sentence) | 1 |
| N5 | delete the map's sentence | 1 (states none) | 1 |

No predicted-zero probe here: the two runway items were prose, and the one guard added was mutated above. The prior passes' 17 mutations on the gate's predicates were not re-run — none of those predicates changed in this pass (`git diff b8d6728..614b5c4` touches the `UNREAD_RUNTIME_RECORDS` comment/`reason`, the docstring bullet, one import, and the new block between the ratchet test and `_runtime_fixture`).

**Mechanism vs the brief:** the runway prescribed two prose edits; both are applied as prescribed (F1: comment + `reason`, F2: the annotation with the measured count). One thing the runway did not prescribe was added and is the first decision below.

- **[Medium]** Added a guard tying the map's stated unread figures to the gate's registries, rather than only correcting the number. Alternatives: (a) correct the number only, per the runway; (b) drop the number from the map (code-reviewer's other remedy). Why: the number went stale once already (`Eight` after two closed on `main`) and nothing held it — that is the class F2 is an instance of, and the review-pr precheck for the next pass commits to re-measuring a number, so (b) would have fought the runway while (a) would have left the class open. Cost: +2 tests, +1 import, in a pass sized for prose — stated in the PR body rather than absorbed. Digits, not number-words, so the sentence has one spelling the regex reads; N2 asserts the words are refused rather than silently parsed.
- **[Low]** Kept "Do NOT file a candidate for the reader; the plan already holds it" in the F1 comment although code-reviewer flagged it as beyond the runway. Why: F1's stated consequence was "the next reader files a duplicate candidate"; the citation alone closes the misstatement, the sentence closes the consequence. The `measure/` placement steer was trimmed as the reviewer asked — it was a design opinion in a comment whose point is not to make one.
- **[Low]** Touched the module docstring's `events.jsonl` bullet (l.155), which the runway's "edit ONLY the comment block" did not name. Why: same fact, third site in the same file, found by code-reviewer; the runway's exclusion list (`RUNTIME_RECORDS`, predicates, controls, `HELD_BY`) does not cover it and leaving it would have left two descriptions of one fact at two staleness levels.
- **[Low]** Did not `git merge origin/main`: `origin/main` is still `f908122`, already the branch's merge-base and merged at `3fbc830`; nothing to merge, suite re-run anyway.

### Disposition of every finding this pass

| Finding | Source | Disposition |
|---|---|---|
| `events-reader-baseline-says-undecided` (F1) | review-pr HOLD | **FIXED** — comment, `reason`, docstring bullet |
| `repo-map-omits-runtime-record-class` (F2) | review-pr HOLD | **FIXED** — class clause + measured figures |
| F1 comment grew a placement steer | code-reviewer W | **FIXED** — trimmed |
| Docstring bullet is a second, staler description | code-reviewer W | **FIXED** |
| Literal count in the map held by nothing | code-reviewer High | **FIXED** — class check + control + 5 negative controls |
| Annotation length | code-reviewer Low | **REJECTED** — matches the map's local convention; `file_structure_check.py` exits 0 and its prose note is informational and pre-existing |
| PR body still asked review-pr to file a candidate | Stage 1 (fourth member of F1's class) | **FIXED** — body's finding section now records the Phase 4 deferral review-pr ruled |

## Deferred Work

- **A reader for the bag's event stream (`events.jsonl`)** — Why deferred: scheduled work, not this PR's. Tracked at: `skyynet-master-planning/development/edge-assistant/persistent-memory-protocol/phase4_rebuild_is_a_test.md` l.176. Verified by: `grep -n 'Build the replay' <that file>` on the live sibling at `1f87880` → line 176, `- [ ] Build the replay: read one edge's journal in order, dedupe on event identity, apply each event to a scratch directory …`; status line reads *not started*. The gate's ratchet line names it and fails the day the replay lands under `scripts/`.
- **Phase 6 doc reconciliation (step 17 → `[x]`, status line)** — Why deferred: the brief reserves the sibling-repo write for the operator. Tracked at: `skyynet-master-planning/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md` l.141. Verified by: review-pr pass 1's read at `1f87880` (`- [ ] **Rule the runtime-record class**`); the sibling is still at `1f87880` this pass (`git rev-parse --short HEAD` there) and I did not write to it.

## Post-Run Reflection

- **Friction:** `gh issue list --search` hit `TLS handshake timeout` twice in a row before succeeding; a run that treated the first failure as "no prior art" would have reported a search it never ran. Worth one retry in the prompt's prior-art step. `grep` on this host is a shell function (`type -a grep` → function), which is why the terminal-output rule's `type -a` line earned its place — it changed nothing here but I only know that because I checked.
- **Project-level suggestions (this repo):** `docs/file_structure.txt` carries other literal counts in annotations ("FOUR partitioned populations", "33 across 12 PRs", …) that nothing holds. The map-figure guard added here keys on one sentence for one gate; if a second annotation's count goes stale the same way, the remedy is a general "counts in the map are asserted or absent" rule in the map checker, not a second per-gate regex — flagging now so the second occurrence is recognised as the class rather than re-fixed as an instance.
- **Tooling-level suggestions (claude-dot-files):** the review-pr runway says *"edit ONLY the comment block … and that entry's `reason` string"* and the workflow prompt says *"fix the class, not the instance"*. On this pass the class had a third member (the docstring bullet) one screen away, and the two instructions pointed opposite ways. The runway's exclusion list (`RUNTIME_RECORDS`, predicates, controls, `HELD_BY`) was the useful part — a runway that names what must NOT move lets the correction pass sweep the class inside the remainder; one that enumerates the only lines that MAY move forbids it. Consider phrasing runway scope as an exclusion list by default.

Pumapumapumas:
## review-pr · pass 2 · PR #193 — disposition

**Reviewed:** head `614b5c40f24177c757534c95556ddfe16a91a97f` (one commit since pass 1's head `3fbc830`: the correction pass `614b5c4`, attested by its own Decision Log comment). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `closingIssuesReferences: []`, `refs/pull/193/merge` = `88266c3` (fresh, differs from pass 1's `072ae0c` — recomputed for the new head). Checks on the merge ref, run `34803960948` with `head_sha` = this head: `suite` pass, CodeQL ×3 pass — same check SET as pass 1, nothing dropped. Worktree repo = PR repo. **`origin/main` is still `f908122`** — the merge-base and the tip — so no rule arrived after the cut and nothing this pass touches has landed elsewhere (both directions run: `git rev-parse origin/main` = merge-base). Prior fence-anchored `pr_review:` blocks on the thread: **1** → this is pass 2 (matches the dispatch label). Attempts: pass 1 carried 2; the correction dispatch landed work → **3**.

**Not re-mined:** the draft's and refine's reflections (pass 1 did that). This pass verifies pass 1's runway against the artifact and reviews the diff since `3fbc830` (2 files, +88/−6).

### Pass 1's prescriptions — verified against the artifact, not the account

- **F1 `events-reader-baseline-says-undecided` → FIXED.** Pass 1's precheck re-run on this head: `sed` block found (20 lines), `grep -c 'phase4_rebuild_is_a_test\|Phase 4'` = **3** (was 0) → ALREADY DONE arm. Read the diff: the `UNREAD_RUNTIME_RECORDS` comment, the entry's `reason`, and the module docstring bullet (l.155 — a third site pass 1 did not name; the correction pass was right to sweep it) all cite PMP Phase 4's replay. Claims inside the new comment re-verified: (a) `_fleet_python()` is `(_REPO / "scripts").rglob("*.py")` minus `tests` — so *"the ratchet walks `scripts/` only"* is true; (b) `applied_intents`'s docstring (`events.py:445–451`) names Phase 4 as the consumer — true; (c) pointer still resolves — planning checkout at `1f87880`, `phase4_rebuild_is_a_test.md:176` is `- [ ] Build the replay: …`, status *not started*. The new prose passes `test_prose_NAMES_a_symbol_that_RESOLVES.py` (ran it: green). No regression from the prescription.
- **F2 `repo-map-omits-runtime-record-class` → FIXED.** Pass 1's precheck re-run: `awk` block found (17 lines), `grep -c -i 'runtime-record\|RUNTIME_RECORDS'` = **3** (was 0) → ALREADY DONE arm; the count arm re-measured — `(sum(len(s.unread) for s in SURFACES), len(UNREAD_RUNTIME_RECORDS))` = **(6, 1)** and the annotation says *"6 unread producers and 1 unread runtime record"*. `scripts/helpers/file_structure_check.py` exits 0 (its two notes — 972 prose lines, and `logs/` naming nothing — are identical on `origin/main`; `logs/` is gitignored and documented as such). Class sweep: `grep -rn 'design decision'` over the gate and the map → 0; `grep -rni 'eight unread\|8 unread'` over `docs/ testing/ scripts/` → 0.

### The diff since pass 1 — the un-prescribed addition, reviewed

**Angle new to this pass, and why:** pass 1 reviewed the gate's predicates; the correction pass added a guard the runway did not ask for (`test_the_REPO_MAP_states_the_UNREAD_figures_this_gate_holds` + control, +1 import), which one code-reviewer lens saw and no review-pr pass has. I ran the **boundary lens on that guard's exits** and the **asymmetric-presence sweep on the helper it now imports** (`file_structure_map.py`, untouched by the PR).

Re-executed: gate alone **89 passed / 49 skipped** (body's figure ✅; pass 1 head was 87); gate + prose gate 114/49; `test_test_tree_hygiene` + the three map/vocabulary gates 60 passed (the new cross-import is helper→test, which hygiene permits — `file_structure_map` is the helper module the hygiene rule itself names as the remedy); ruff `F821,F401,E9` clean. `file_structure_map` imports only `planning_corpus.PLANNING_ROOT`, which the gate already imported — the body's *"no planning-repo I/O at import time"* claim holds, and CI's sibling-less `suite` is green on this exact head.

Exits of the guard, each probed: no map entry → `assert annotation` fails loud; sentence absent → `stated is None` fails loud with the shape in the message; digits differ → fails loud with both tuples. Two probes I ran against the live map text through the guard's own functions: **(i)** the sentence reflowed across two continuation lines (what an editor re-wrapping the annotation does) → `_map_unread_counts` returns `None` → the "restate them" red fires — **loud, not silent**, but a false red on a present figure, because `_map_annotation` joins with `"\n"` and keeps the `#   ` decoration in the text (the exact hazard `test_retired_vocabulary…::_joinable` records for the same file); **(ii)** a tree glyph inside the annotation's prose → the block truncates early (10 of 17 lines) → also `None` → loud. Leaf lookup is by basename, first match; `test_every_producer_NAMES_ITS_CONSUMER.py` occurs once in the map (verified). Nothing is broken on this head; both probes are future-edit shapes. Listed under recommendations below, not held on — see the convergence line.

### Correction-pass reflection — one item is a recurrence of a filed candidate

The correction pass's project-level suggestion — *"`docs/file_structure.txt` carries other literal counts that nothing holds … the remedy is a general 'counts in the map are asserted or absent' rule in the map checker, not a second per-gate regex — flagging now so the second occurrence is recognised as the class"* — is not a new class. It IS **`C-523klr8n`** (*derive declared roster and inventory counts rather than restating them in prose*), `status: open, count: 2`, whose body names `file_structure.txt` three times and whose open question is *"whether that shape generalises to the roster and inventory counts this row was originally about"*. This PR's `Eight`→6 decay is that shape in that file. Duplicate check run: `similar-candidates.py --store candidates --repo-root <planning> --repo claude-dot-files "<subject>"` → `C-523klr8n` second-ranked (first is an unrelated PMP-component ruling); intake queues in both repos → no open intake carries it (the prior recurrence, #146, was harvested and closed). Per Tracked Items §3.1 a recurrence is an increment, not a second item: **filed `RECURRENCE on C-523klr8n` as intake Claude-Dot-Files#195** (`store: candidates`), body carries the SHA, the measurement, and why the per-gate guard is the instance rather than the class. Placement: question 2 (closely related to an existing item → expansion of it); qualification: unrelated to the work in hand (the class rule lives in the map checker, which this PR does not touch), substantial (a checker rule with a convention decision behind it), and covered by an existing item — which is exactly why it is a recurrence and not a filing.

### Deferrals — re-adjudicated (unchanged, pointers re-verified on the live sibling at `1f87880`, working tree clean)

- `events-jsonl-reader` → PMP Phase 4 l.176, still `[ ]`, status *not started*. Re-executed the no-reader measurement per public function of `events.py`: `decode_event`, `dedupe_on_identity`, `applied_intents` → **NONE** outside tests; the other four are called only from `emit.py`. Honest, disclosed, pointer valid → neither laundered nor homeless.
- `phase6-step17-reconciliation` → Phase 6 doc l.141 still `- [ ] **Rule the runtime-record class**`; l.3 still *"16 of 17"*. Operator flips after merge; text as pass 1 recommended. No planning artifact in this PR's diff (`git diff --stat origin/main...HEAD` = the two files) — the *"no checkbox flipped"* claim holds.

### Disposition table

|---|---|---|---|
| `events-reader-baseline-says-undecided` (F1) | doc-drift | **FIXED** | Comment, `reason`, docstring bullet cite PMP Phase 4; precheck → ALREADY DONE (3); prose gate green; the comment's `scripts/`-only claim verified against `_fleet_python`. |
| `repo-map-omits-runtime-record-class` (F2) | doc-drift | **FIXED** | Annotation names the class; states (6, 1) = measured; map checker exit 0; now held by a guard. |
| `map-count-restated-class-recurrence` | doc-drift | **DEFERRED** | → `C-523klr8n` (planning `tracked/candidates/`, open, count 2 → intake #195 increments to 3). Recurrence, not a new item. |
| `events-jsonl-reader` | deferral | **DEFERRED** | → PMP Phase 4 `phase4_rebuild_is_a_test.md` l.176 (re-verified `[ ]` at `1f87880`). |
| `phase6-step17-reconciliation` | deferral | **DEFERRED** (operator-action) | → Phase 6 doc l.141 / l.3 (re-verified `[ ]` at `1f87880`). Operator flips after merge. |
| `annotated-constant-escapes-population` | correctness | FIXED | Carried from pass 1; predicates untouched since (`git diff b8d6728..614b5c4` confirms). |
| `control-name-count-mismatch` | test-gap | FIXED | Carried. |
| `verified-by-grep-not-reproducible` | doc-drift | FIXED | Carried; the per-function grep re-run this pass gives the same answer. |
| `reader-checks-co-fire` | correctness | FIXED | Carried. |
| `all-exclusion-single-spelling` | correctness | FIXED | Carried. |
| `via-first-hop-liveness` | correctness | FIXED | Carried. |
| `unresolved-holders-duplicates-defines-function` | correctness | FIXED | Carried. |
| `basename-keyed-population` | correctness | FIXED | Carried. |
| `module-ast-uncached` | correctness | FIXED | Carried. |
| `held-by-dataclass-retrofit` | scope | REJECTED | Carried — design preference outside the brief; `HELD_BY` untouched. |
| `self-referential-entry-cannot-fail` | correctness | FIXED | Carried. |
| `neither-family-overstates` | doc-drift | FIXED | Carried. |
| `run-log-exec-module` | security | REJECTED | Carried — tracked repo file, same loader as the replay tools, no untrusted path. |
| `draft-control-predicted-7-observed-8` | test-gap | FIXED | Carried. |
| `all-string-named-init-as-reader` | correctness | FIXED | Carried. |
| `prose-gate-red-on-backticked-glob` | test-gap | FIXED | Carried; `suite` green on this head. |

**Verdict rationale:** both runway items landed as prescribed and are verified against the artifact with pass 1's own prechecks (each now returns the ALREADY-DONE arm) plus independent checks; the correction pass's one un-prescribed addition is a guard whose every exit fails loud; the only new material is a recurrence of an already-filed candidate (incremented via intake, not held on) and two preventive shapes. **Converged: this pass's only findings are preventive** — neither would have blocked on pass 1's bar, which was *"the artifact states a wrong fact"*; the artifact now states the right facts and a guard holds one of them. **MERGE.**

### Recommendations (preventive — not findings, no live defect; listed so they are visible without holding the PR)

1. **`_map_annotation` joins continuation lines with `"\n"` and keeps the `#` decoration**, so a re-wrap of the map sentence turns a present figure into a "states none" red. Probe result above: `None`. If it bites, the one-line remedy is to strip the tree-glyph-and-`#` prefix from each continuation and join with a space — the `_joinable` shape in `test_retired_vocabulary_is_gone_from_live_surfaces.py:325`. Loud today, so not held.
2. **`file_structure_map.py`'s docstring says "THE TWO CONSUMERS"** and the map's annotation says *"asked by both map gates"*; there are now three consumers (the producer gate imports `LEAF` and `MAP`). Any change to `LEAF`'s group numbering breaks the gate loudly in the suite, so nothing silently regresses — but the next reader of the helper undercounts who depends on it. One-line docstring touch whenever that file is next open.

### Notes (not findings — no consequence stated)

- Carried from pass 1: `C-wb1xc1xs` looks satisfied on main (its reader `replay_convergence_events.py` exists and says so) and is still `open, count 1` — a triage close, not this PR's.
- Boundary observations carried from pass 1 stand unchanged (non-recursive population glob; `def` not `async def`; hand-class records; ratchet trips on a string value; fleet is `scripts/` only — the F1 comment now states that last one in the gate itself).
- Inbound-citation sweep on the two files whose line counts moved: no `test_every_producer_NAMES_ITS_CONSUMER.py:<n>` citation anywhere; the only `file_structure.txt:<n>` citations are `:61`, `:97`, `:118` in planning prose/tracked items (historical, SHA-pinned, and all above the edit at l.1199+) — nothing lands wrong.
- The map checker's *"NAMES NOTHING — `logs/`"* note is identical on `origin/main` (gitignored directory, annotated as such); not this PR's.

### Post-Run Reflection

- **Friction:** none. The correction pass's suggestion that runways be phrased as an exclusion list ("do NOT touch X") rather than an inclusion list ("edit ONLY Y") is right for the class-sweep case — pass 1's F1 runway said both, and the correction pass correctly read the exclusion list as the boundary and swept the third site. Worth stating in the `dispatch_context` guidance: *fence what must not move; do not enumerate the only lines that may.*
- **Schema:** the gate-0 rule (only `fixed | rejected | hold` for an edited artifact) has no slot for a preventive observation about a guard this PR added — `noted` is unreachable, `rejected` would record a real shape as not-an-issue, and `hold` contradicts the convergence rule. I put both under *Recommendations* outside the findings list, which is what the convergence rule says to do; saying so here so a later pass does not re-derive it.
- **Tooling-level suggestions (claude-dot-files):** none beyond pass 1's (`similar-candidates.py` cannot see phase-doc checkboxes). One confirmation for it: the recurrence path worked exactly as designed this pass — the correction pass proposed a "class rule" as if new, and the duplicate check surfaced the two-count candidate that already owns it.

FILED-INTAKE: https://github.com/helloskyy-io/Claude-Dot-Files/issues/195

```yaml
pr_review:
  pr: 193
  run_id: 64a8722c9c544f6d858e0259b2ef0f03
  pass: 2
  attempt: 3
  verdict: MERGE
  converged: true
  findings:
    - id: events-reader-baseline-says-undecided
      title: "The gate's baseline says the events.jsonl reader is an open design decision when PMP Phase 4 already schedules it, so the next reader files a duplicate candidate"
      category: doc-drift
      consequence: "A comment in a safety gate misstates the planning state and every later reader re-derives a duplicate candidate."
      disposition: fixed
      remedy: fix-in-place
    - id: repo-map-omits-runtime-record-class
      title: "The authoritative repo map does not describe the runtime-record class this PR added and states an unread count the gate no longer holds"
      category: doc-drift
      consequence: "docs/file_structure.txt tells the next dispatch the class is unruled and the count is eight where the gate holds 6 + 1."
      disposition: fixed
      remedy: fix-in-place
    - id: map-count-restated-class-recurrence
      title: "Literal counts in the repo map decay with nothing holding them — this PR's 'Eight' was the third such decay in that file, fixed per-instance"
      category: doc-drift
      consequence: "Each restated count in file_structure.txt goes stale silently until a reader notices; the per-gate guard added here holds one sentence, not the class."
      disposition: deferred
      pointer: "skyynet-master-planning/tracked/candidates/C-523klr8n.md (open, count 2) — recurrence filed as intake https://github.com/helloskyy-io/Claude-Dot-Files/issues/195, store: candidates"
      pointer_verified: true
      reviewed_sha: 1f87880
    - id: events-jsonl-reader
      title: "The bag's event stream is decoded by nothing outside the test suite until PMP Phase 4's replay lands"
      category: deferral
      consequence: "A run's event stream can stop being correct and nothing goes red until the scheduled replay exists; the ratchet line holds the place."
      disposition: deferred
      pointer: "skyynet-master-planning/development/edge-assistant/persistent-memory-protocol/phase4_rebuild_is_a_test.md line 176 (Build the replay)"
      pointer_verified: true
      reviewed_sha: 1f87880
    - id: phase6-step17-reconciliation
      title: "The Phase 6 doc says the runtime-record step is open after the PR that closes it merges, until the operator flips it"
      category: deferral
      consequence: "The planning status line under-reports the phase by one step until reconciled."
      disposition: deferred
      remedy: operator-action
      pointer: "skyynet-master-planning/development/edge-assistant/workflow-decomposition/phase6_every_producer_names_its_consumer.md line 141 (and status line 3)"
      pointer_verified: true
      reviewed_sha: 1f87880
    - id: annotated-constant-escapes-population
      title: "An annotated journal constant would escape the derived population"
      category: correctness
      consequence: "A record declared `X_FILE: str = ...` would need no reader and nothing would go red."
      disposition: fixed
      remedy: fix-in-place
    - id: control-name-count-mismatch
      title: "A control whose name and assertion count disagree reads as under-tested"
      category: test-gap
      consequence: "A reader trusts the name and misses the extra cases."
      disposition: fixed
      remedy: fix-in-place
    - id: verified-by-grep-not-reproducible
      title: "A \"Verified by\" command that does not reproduce as written weakens the attestation it supports"
      category: doc-drift
      consequence: "A reviewer re-running it gets a different count and doubts the substance."
      disposition: fixed
      remedy: fix-in-place
    - id: reader-checks-co-fire
      title: "Reader checks reporting twice contradict the docstring's stop-at-first-part claim"
      category: correctness
      consequence: "Mutation counts drift from predictions and a control asserts the wrong number."
      disposition: fixed
      remedy: fix-in-place
    - id: all-exclusion-single-spelling
      title: "An annotated or augmented __all__ would count as reading the accessor"
      category: correctness
      consequence: "A re-exporting module trips the unread ratchet falsely."
      disposition: fixed
      remedy: fix-in-place
    - id: via-first-hop-liveness
      title: "The via chain prose claimed a first-hop liveness the check does not assert"
      category: correctness
      consequence: "A reader over-reads the check; the import-graph test holds the property instead."
      disposition: fixed
      remedy: fix-in-place
    - id: unresolved-holders-duplicates-defines-function
      title: "Two copies of the defines-function regex can drift apart"
      category: correctness
      consequence: "One holder check accepts what the other rejects."
      disposition: fixed
      remedy: fix-in-place
    - id: basename-keyed-population
      title: "A basename-keyed population lets a same-basename module elsewhere stand in for the journal's"
      category: correctness
      consequence: "A hand entry pointing at the wrong module satisfies the population test."
      disposition: fixed
      remedy: fix-in-place
    - id: module-ast-uncached
      title: "The unread ratchet re-parses every fleet module per baselined record"
      category: correctness
      consequence: "Gate runtime grows with the baseline; the file already caches the sibling case."
      disposition: fixed
      remedy: fix-in-place
    - id: held-by-dataclass-retrofit
      title: "HELD_BY remains a tuple registry beside a dataclass one"
      category: scope
      consequence: "None in this PR; a design preference outside the brief."
      disposition: rejected
      remedy: reject
    - id: self-referential-entry-cannot-fail
      title: "An entry whose reader is its declaring module could never fail \"names it in code\""
      category: correctness
      consequence: "Five of thirteen entries were unfalsifiable on that part."
      disposition: fixed
      remedy: fix-in-place
    - id: neither-family-overstates
      title: "The docstring claimed one hand-registered record where there are two"
      category: doc-drift
      consequence: "A reader believes the config-digest tag is derived and would not re-register it."
      disposition: fixed
      remedy: fix-in-place
    - id: run-log-exec-module
      title: "The population helper executes run_log.py by path"
      category: security
      consequence: "None — tracked repo file, same loader the replay tools use, no untrusted path."
      disposition: rejected
      remedy: reject
    - id: draft-control-predicted-7-observed-8
      title: "The draft's control observed one more defect than predicted"
      category: test-gap
      consequence: "Model-of-fixture error that hid the co-firing defect until traced."
      disposition: fixed
      remedy: fix-in-place
    - id: all-string-named-init-as-reader
      title: "The first ratchet run named journal/__init__.py as a reader on the strength of __all__"
      category: correctness
      consequence: "The baseline would have been refused for a re-export, not a read."
      disposition: fixed
      remedy: fix-in-place
    - id: prose-gate-red-on-backticked-glob
      title: "The prose-symbol gate read a backticked glob as a private symbol"
      category: test-gap
      consequence: "CI red on the full suite with the gate alone green."
      disposition: fixed
      remedy: fix-in-place
  next_steps: []
  laundered_deferrals:
    caught: 0
    of_total: 3
  homeless_items: 0
  redispatched: false
  filed_intakes:
    - https://github.com/helloskyy-io/Claude-Dot-Files/issues/195
```

